PR #5, opened 2022-10-26
WIP - query params merge

Omri-Levy:
As part of this refactor a variable mergedConfig was created (see sdks/web-sdk/src/main.ts line 68), currently it is not in use. This PR passes the variable into updateConfiguration (sdks/web-sdk/src/main.ts line 76) and tests the change.


Omri-Levy:
The expected values are sent within verify's payload and setting the language query param to something unsupported is communicated through the UI. Something to consider is adding client-side validation for things like an email (see the screenshot), this would improve the time-to-feedback to the user and improve server load by avoiding an invalid network call. Can share a single validation schema between the server and client.
image

alonp99:
> The expected values are sent within verify's payload and setting the language query param to something unsupported is communicated through the UI. Something to consider is adding client-side validation for things like an email (see the screenshot), this would improve the time-to-feedback to the user and improve server load by avoiding an invalid network call. Can share a single validation schema between the server and client. image

Yeah, it's a good time to add the validations. can you open an issue?

PR #7, opened 2022-10-26
WIP - Commitlint, Editorconfig, ESLint, Prettier with NX

Omri-Levy:
The configuration and dependencies needed for making the listed tools work within a monorepo.


sdks/web-sdk/package.json
```diff
@@ -78,7 +78,6 @@
     "eslint-plugin-svelte3": "^4.0.0",
     "husky": "^8.0.1",
     "postcss": "^8.4.18",
-    "prettier": "^2.7.1",
```
alonp99, line 81: Im not sure about this one. I saw you moved it to the root pjson, but web-sdk shouldn't be dependent on the monorepo. it should always stay in a stage where we can move it anywhere else and its able to run standalone

packages/config/prettierrc.web-sdk.cjs
```diff
@@ -0,0 +1,9 @@
+module.exports = {
```
alonp99, line 1: this file should probably be inside web-sdk, or called `prettierrc.svelte.cjs`

sdks/web-sdk/package.json
Omri-Levy: I can add it back but the way I see it as long as its part of the monorepo it will depend on it and once its not changes will be made then. To me monorepos do not share the same goal of microservices (100% self contained).

packages/config/prettierrc.web-sdk.cjs
Omri-Levy: Putting it inside web-sdk would be the opposite of our chat regarding moving configs to a single centralized source of truth. I considered .svelte. but I'd like the configs to be based on the package that consumes them, Svelte is the part that may change as well.

alonp99 (approved):

PR #18, opened 2022-10-27
WIP - NX NPM scripts

Omri-Levy:
Adds the missing scripts one would expect from a monorepo, refactors script names and commands to be more consistent and adhere to NX's docs.


Omri-Levy (commented):
Please note the following;

- The use of nx affected might be something undesired for format/lint, subject to change where to use affected
- The --projects flag for commands like dev is something that could include other projects that make sense to be ran in parallel, currently it only includes web-sdk
- Changed web-sdk's preview command to start to ensure parity, non-vite projects such as node projects expect a start script, this allows running start from root with no error or excluding Vite projects
- Formatting changes are due to testing the lint and format commands, should be a non-issue when the formatting/ci PRs are merged
- Right now the test scripts only echo a message, this will be as part of the tests PRs

alonp99 (approved):

PR #51, opened 2022-10-31
WIP - Fix linting errors and warnings

Omri-Levy:
Refactoring of code and addition of TypeScript types required to satisfy ESLint.

### DIFF
--- a/sdks/web-sdk/src/declarations.d.ts
+++ b/sdks/web-sdk/src/declarations.d.ts
@@ -7,7 +7,7 @@ interface IPosthog {
 }
 
 declare global {
-  var __APP_VERSION__: string;
+  let __APP_VERSION__: string;
 
   interface Window {
     isProd: boolean;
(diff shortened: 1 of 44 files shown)


sdks/web-sdk/src/declarations.d.ts
```diff
@@ -0,0 +1,28 @@
+// Should use Posthog's types here instead
+import { AnyRecord } from './types';
+import { DevMocks } from './lib/contexts/app-state';
+
+interface IPosthog {
```
alonp99, line 5: We can remove all references to posthog

alonp99 (approved):

sdks/web-sdk/src/declarations.d.ts
Omri-Levy: Removed src/analytics.ts as its currently the only file using posthost. 
image

PR #172, opened 2023-03-30
feat: added init script to bootstrap the project

victor-ca:
### Description
- added a script that automates bare project initiation
- added convenience profile aliases


deploy/docker-compose.yml
```diff
@@ -1,13 +1,17 @@
...
   backoffice:
+    depends_on: 
+      - db
     build:
       context: ../apps/backoffice-v2/
       args:
         NPM_LOG_LEVEL: notice
     ports:
-      - ${BACKOFFICE_PORT}:3000
+      - ${BACKOFFICE_PORT}:5173
```
Omri-Levy, line 11: Would be nice if the external port would also be an environment variable. :)

services/workflows-service/scripts/seed.ts
```diff
@@ -5,9 +5,8 @@ import { hash } from 'bcrypt';
 import { customSeed } from './custom-seed';
 import { endUserIds, generateEndUser } from './generate-end-user';
 
+dotenv.config();
```
alonp99, line 8: Can you explain this change?

alonp99 (approved):

deploy/docker-compose.yml
victor-ca: change reverted, to the master state.

PR #176, opened 2023-04-03
Chore/ci up

victor-ca:
### Description
fixed lint, dropped obsolete test, slightly updated CI


sdks/workflow-browser-sdk/src/lib/tests/injections.test.ts
```diff
@@ -28,39 +29,4 @@ describe('injections', () => {
       actions: ['USER_NEXT_STEP'],
     });
   });
-
-  it('should inject plugins', () => {
```
alonp99, line 32: It looks like a valid test, im not sure we want to remove it instead of fixing it

sdks/workflow-node-sdk/src/lib/workflow-node-sdk.test.ts
```diff
@@ -28,74 +28,3 @@ test('Simple Server Workflow', t => {
   runner.sendEvent({ type: 'TOGGLE' });
   expect(runner.getSnapshot().value).toBe('inactive');
 });
-
-test('Server Workflow persistance MemoryStore', t => {
```
alonp99, line 32: Same here, doesnt look like a test we want to remove

sdks/workflow-browser-sdk/.eslintrc.cjs
```diff
@@ -8,4 +8,14 @@ module.exports = {
     project: ['./tsconfig.eslint.json'],
   },
   ignorePatterns: ['.eslintrc.cjs'],
+  rules: {
+    '@typescript-eslint/no-unused-vars': [
```
Omri-Levy, line 12: Worried this `@typescript-eslint/no-unused-vars` rule will result in more unused code long term. For places which an unused variable makes sense disabling the rule only for that line is something I'd feel more comfortable with.

.github/actions/lint-action/action.yml
```diff
@@ -34,6 +34,10 @@ runs:
       shell: bash
       run: pnpm install
 
+    - name: Build
```
Omri-Levy, line 37: If the lint step requires the build artifacts check out our nx.json, specifically the `test.dependsOn` line. :)

sdks/workflow-browser-sdk/.eslintrc.cjs
victor-ca: the rule is not disabled, it allows prepending underscore to unused variables;
there are some valid usages of this rule, like dropping something from arrays eg: `const [_drop, _me, useMe] = ...` which looks better than `const [,, useMe]` or skipping some stuff in arguments. 

I'd leave it, it will be visible in in PRs - however i have no strong opinion on it

.github/actions/lint-action/action.yml
victor-ca: please elaborate, i don't get it

sdks/workflow-browser-sdk/.eslintrc.cjs
Omri-Levy: Exactly what I meant, the valid usages are great. Problem is there is no way of limiting the rule to destructuring. I don't mind it in the PR either, some of the instances where we use _ might require revisiting.

.github/actions/lint-action/action.yml
Omri-Levy: If linting fails due to the build step it would make a lot of sense to add a config like test has but for lint.
image

sdks/workflow-browser-sdk/.eslintrc.cjs
alonp99: Maybe a more specific prefix will help

services/workflows-service/.eslintrc.cjs
```diff
@@ -33,6 +33,15 @@ module.exports = {
   plugins: ['@typescript-eslint'],
   rules: {
     'import/no-cycle': 'error',
+
+    '@typescript-eslint/no-unused-vars': [
+      'error',
+      {
+        argsIgnorePattern: '^_',
```
alonp99, line 40: seems like its being duplicated, can we move it to packages/config?

victor-ca: great idea! pulled to root + connected via extend

.github/actions/lint-action/action.yml
victor-ca: indeed it does, moved there

alonp99 (approved):

sdks/workflow-browser-sdk/.eslintrc.cjs
alonp99: "drop_", "unused_"? now when writing down it feel a bit strange - im good with Omri's suggestion

victor-ca: i'll drop the shared config as well then

PR #179, opened 2023-04-04
Adds design to headless-example

Omri-Levy:
### Description
Elaborate on the subject, motivation, and context.

### Related issues

### Breaking changes

### Examples and references

### Checklist
- [x] I have read the [contribution guidelines](CONTRIBUTING.md) of this project
- [x] I have read the [style guidelines](STYLE_GUIDE.md) of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings and errors
- [x] New and existing tests pass locally with my changes


examples/headless-example/src/components/App.svelte
```diff
@@ -153,28 +164,35 @@
 
   let nextWorkflow;
   let shouldResubmit = false;
+  $: isCompleted = $workflowQuery.data?.workflowRuntimeData?.status === 'completed';
```
victor-ca, line 167: we have a lot of null coalescing operators among the project; would it make sense to enforce typings so we would have only the required ones?

examples/headless-example/src/components/App.svelte
```diff
@@ -191,37 +209,37 @@
...
-    You've been requested to re-submit your documents due
-    to {nextWorkflow?.definition?.context?.documentOne?.resubmissionReason?.toLowerCase()?.replace(/_/g, ' ')}
-    . Please click navigate to re-submit your documents.
-  </p>
-  <button on:click={handleResubmit}>Navigate</button>
-{/if}
+<main class="h-full flex flex-col items-center justify-center p-6">
+
+  {#if !endUserId}
+    <SignUp {onSubmit}/>
```
victor-ca, line 215: looks like we're achieving routing by using if's - do we consider using real routing?

victor-ca (approved):

examples/headless-example/src/components/App.svelte
Omri-Levy: In this case it makes sense since data is undefined at times (@tanstack/query behaviour). Some other cases are due to Prisma where we would need to discuss. Otherwise I agree with you.

examples/headless-example/src/components/App.svelte
Omri-Levy: Routing wouldn't make sense since it is unrelated to the URL and this is just a demo. I also spoke with Alon on how we need more reliable properties to create conditions with so we won't need as many ifs or at least as complicated.

PR #262, opened 2023-05-03
blokh/feat/implment-s-3-filemanager

Blokh:
### Description
added S3 client.
added multer S3 file handler
add uri as resource in "File" table

Be aware, uri does not have default value>
therefore current users of the database will have issues with the migraiton since uri does not have default value. 

### Breaking changes
Added S3 bucket manager

add S3 envs.
use the headless example to upload image


services/workflows-service/src/storage/get-file-storage-manager.ts
```diff
@@ -0,0 +1,34 @@
+import { diskStorage } from 'multer';
+import { getFileName } from '@/storage/get-file-name';
+import { z } from 'zod';
+import multerS3 from 'multer-s3';
+import { S3Client, S3ClientConfig } from '@aws-sdk/client-s3';
+
+const isS3BucketConfigured = () => {
+  return process.env.AWS_S3_BUCKET_KEY;
```
alonp99, line 8: consider using our secret manager

apps/backoffice-v2/src/api/storage.ts
```diff
@@ -4,18 +4,29 @@ import { apiClient } from './api-client';
...
       endpoint: endpoints.storage.fileById.endpoint(fileId),
       method: endpoints.storage.fileById.method,
-      schema: z.instanceof(Blob),
-      isBlob: true,
+      schema: FileInfoSchema,
     });
+    const fileInfo = handleZodError(error, fileInfoResponse);
 
-    const data = handleZodError(error, blob);
+    if (fileInfo.bucketKey) {
```
Omri-Levy, line 18: Changing this line to `if (!fileInfo.bucketKey) return fileInfo.uri;` will have a positive impact on readability. :)

services/workflows-service/src/storage/get-file-storage-manager.ts
```diff
@@ -0,0 +1,71 @@
...
+  return {
+    region: z.string().parse(process.env.AWS_REGION),
+    credentials: {
+      accessKeyId: z.string().parse(process.env.AWS_S3_BUCKET_KEY),
+      secretAccessKey: z.string().parse(process.env.AWS_S3_BUCKET_SECRET),
+    },
+  };
+};
+
+export const fetchDefaultBucketName = () => {
```
Omri-Levy, line 26: I think we could access `process.env.AWS_S3_BUCKET_NAME` here without a function, especially if it was already validated. Otherwise changing `fetch` to `get` would be clearer to me.

services/workflows-service/src/storage/get-file-storage-manager.ts
```diff
@@ -0,0 +1,71 @@
...
+  return z.string().parse(process.env.AWS_S3_BUCKET_NAME);
+};
+
+export const manageFileByProvider = () => {
+  if (isS3BucketConfigured()) {
+    return multerS3({
+      s3: new S3Client(generateAwsConfig()),
```
Omri-Levy, line 33: Thoughts about having a single S3 client exported? I see this file has multiple instances of `S3Client`.

services/workflows-service/src/storage/storage.controller.external.ts
```diff
@@ -57,15 +59,37 @@ export class StorageControllerExternal {
...
+      throw new errors.NotFoundException('file not found');
+    }
 
-    if (!fileNameOnDisk) {
+    return res.send(persistedFile);
+  }
+
+  // curl -v http://localhost:3000/api/storage/content/1679322938093
+  @common.Get('/content/:id')
+  async fetchFileContent(@Param('id') id: string, @Res() res: Response) {
```
Omri-Levy, line 75: Please change `fetchFileContent` to `getFileContentById` for consistency. :pray:

services/workflows-service/src/storage/get-file-storage-manager.ts
Blokh: This client is configured and dynamicly passed to the "interface".
as in the future we'll have Fido AWS config and multiple "clients" talking to eachother (moving files from one to another"

services/workflows-service/src/storage/storage.controller.external.ts
Blokh: in this case, it fetches the file's content itself. and not getting the file object out of the databsase.
therefore I am not sure if consistency is the best practice here.
it might. but it does not consistent to other methods.
therefore should the naming be consistent then?

services/workflows-service/src/storage/get-file-storage-manager.ts
```diff
@@ -0,0 +1,71 @@
...
+import { z } from 'zod';
+import multerS3 from 'multer-s3';
+import { S3Client, S3ClientConfig, GetObjectCommand } from '@aws-sdk/client-s3';
+import * as tmp from 'tmp';
+import * as fs from 'fs';
+
+import { Readable } from 'stream';
+export type TLocalFile = string;
+
+const isS3BucketConfigured = () => {
```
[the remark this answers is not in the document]
Blokh: I somewhat disagree, 
yeah onliners are not the best, but, they are more flexible.
If I'll have to add more logic to the bucket configuration. the variable instead of method won't look well.

PR #277, opened 2023-05-08
Feature/remove unnecessary indexes

Blokh:
### Description
Removed unnecessary indexes

### DIFF
--- a/services/workflows-service/prisma/schema.prisma
+++ b/services/workflows-service/prisma/schema.prisma
@@ -18,6 +18,8 @@ model User {
 
   createdAt DateTime @default(now())
   updatedAt DateTime @updatedAt
+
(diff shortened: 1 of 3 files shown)


services/workflows-service/prisma/schema.prisma
```diff
@@ -55,10 +55,9 @@ model EndUser {
   EndUsersOnBusinesses EndUsersOnBusinesses[]
 
   @@index([endUserType])
-  @@index([createdAt])
-  @@index([firstName])
-  @@index([correlationId])
   @@index([approvalState])
+  @@index([firstName])
```
alonp99, line 59: no need, it think we will cover it in free text serach

alonp99 (approved):

PR #284, opened 2023-05-10
Blokh/feat/implment-s-3-filemanager

Blokh:
### Description
generated the logic of upload and downloading file from the backoffice uisng S3 client

### Breaking changes
S3 implemntnation

### Examples and references
configure S3 and use headless example


services/workflows-service/src/storage/get-file-storage-manager.ts
```diff
@@ -0,0 +1,81 @@
...
+    const writableStream = fs.createWriteStream(tmpFile.name);
+
+    return new Promise((resolve, reject) => {
+      readableStream
+        .pipe(writableStream)
+        .on('finish', () => {
+          resolve(tmpFile.name);
+        })
+        .on('error', error => {
+          reject(error);
```
yairper, line 74: critical:
pay attention that u propagate this error to the api, hence to the client, which will return internal s3 errors, which can expose the user to data exposure

services/workflows-service/src/storage/get-file-storage-manager.ts
```diff
@@ -0,0 +1,81 @@
...
+        });
+    });
+  } catch (error) {
+    console.error('Error downloading file from S3:', error);
+    throw error;
```
yairper, line 79: critical:
same here

apps/backoffice-v2/src/lib/zod/schemas/file-info.ts
```diff
@@ -0,0 +1,8 @@
+import { z } from 'zod';
+import { ObjectWithIdSchema } from '../utils/object-with-id';
+export const FileInfoSchema = ObjectWithIdSchema.extend({
+  userId: z.string().nullable().default(''),
```
yairper, line 4: how come it has a `default` if the db schema defines it as required?

services/workflows-service/src/storage/storage.controller.internal.ts
```diff
@@ -44,29 +46,38 @@ export class StorageControllerInternal {
...
+  async fetchFileContent(@Param('id') id: string, @Res() res: Response) {
     // currently ignoring user id due to no user info
-    const fileNameOnDisk = await this.service.getFileNameById({
+    const persistedFile = await this.service.getFileNameById({
       id,
       userId: '',
     });
-
-    if (!fileNameOnDisk) {
+    if (!persistedFile) {
```
yairper, line 69: IMHO, it should be part of the `service`

services/workflows-service/src/storage/storage.controller.internal.ts
```diff
@@ -44,29 +46,38 @@ export class StorageControllerInternal {
...
       throw new errors.NotFoundException('file not found');
     }
 
-    return res.sendFile(fileNameOnDisk, { root: './upload' });
+    if (persistedFile.bucketKey) {
```
yairper, line 73: it repeats itself in `internal` and `external` can it be reused?

yairper (changes requested):

services/workflows-service/src/storage/get-file-storage-manager.ts
```diff
@@ -0,0 +1,81 @@
...
+    },
+  };
+};
+
+const isS3BucketConfigured = (process: NodeJS.ProcessEnv) => {
+  return !!z.string().optional().parse(process.AWS_S3_BUCKET_KEY);
+};
+
+export const fetchDefaultBucketName = (process: NodeJS.ProcessEnv) => {
+  return z.string().parse(process.AWS_S3_BUCKET_NAME);
```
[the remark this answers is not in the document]
Blokh: it will throw exception if it does not set.
therefore it's a good standard to zod wrap it

services/workflows-service/src/storage/get-file-storage-manager.ts
```diff
@@ -0,0 +1,81 @@
...
+};
+
+export const manageFileByProvider = (process: NodeJS.ProcessEnv) => {
```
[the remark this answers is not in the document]
Blokh: I've changed it already lol. for some reason was not pushed.

examples/headless-example/src/components/RemoteImage.svelte
```diff
@@ -20,13 +20,30 @@
...
+      src = fileInfo.uri;
+    } else {
+      const streamedFile = await fetchBlob<Blob>(
+        `http://localhost:3000/api/external/storage/content/${id}`,
+      );
+
+      const base64 = await blobToBase64(streamedFile);
 
-    src = base64?.replace(/application\/octet-stream/gi, fileType);
+      src = base64?.replace(/application\/octet-stream/gi, fileType);
```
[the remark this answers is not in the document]
yairper: @Omri-Levy WDYT?

services/workflows-service/src/storage/storage.controller.internal.ts
yairper: we've decided that's not part of the scope of the task,
as for the future, @alonp99 WDYT?

examples/headless-example/src/components/RemoteImage.svelte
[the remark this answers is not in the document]
Omri-Levy: There can be a function that both uses `blobToBase64` and does `base64?.replace(/application\/octet-stream/gi, fileType)` inside of it. but, I don't think `blobToBase64` should do more than it does now.

services/workflows-service/src/storage/get-file-storage-manager.ts
```diff
@@ -0,0 +1,81 @@
...
+  return {
+    region: AWS_REGION,
+    credentials: {
+      accessKeyId: AWS_S3_BUCKET_KEY,
+      secretAccessKey: AWS_S3_BUCKET_SECRET,
+    },
+  };
+};
+
+const isS3BucketConfigured = (process: NodeJS.ProcessEnv) => {
```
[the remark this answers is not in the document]
Blokh: on that

services/workflows-service/src/storage/storage.controller.internal.ts
alonp99: I think validations (and checks in that sort) should be part of the service

services/workflows-service/src/storage/get-file-storage-manager.ts
```diff
@@ -0,0 +1,82 @@
+import { diskStorage } from 'multer';
+import { getFileName } from '@/storage/get-file-name';
+import { z } from 'zod';
+import multerS3 from 'multer-s3';
+import { S3Client, S3ClientConfig, GetObjectCommand } from '@aws-sdk/client-s3';
+import * as tmp from 'tmp';
+import * as fs from 'fs';
+
+import { Readable } from 'stream';
+export type TLocalFile = string;
```
alonp99, line 10: No need to export AFAICS

services/workflows-service/src/storage/get-file-storage-manager.ts
```diff
@@ -0,0 +1,82 @@
...
+    region: AWS_REGION,
+    credentials: {
+      accessKeyId: AWS_S3_BUCKET_KEY,
+      secretAccessKey: AWS_S3_BUCKET_SECRET,
+    },
+  };
+};
+
+const __isS3BucketConfigured = (processEnv: NodeJS.ProcessEnv) => {
+  return !!z.string().optional().parse(processEnv.AWS_S3_BUCKET_KEY);
```
alonp99, line 33: Can you add a an issue to change those to work with t3-env?

services/workflows-service/src/storage/storage.controller.internal.ts
Blokh: I think the file service has no knowledge of weather the file exists or not.
you provide him the link/configuration/action and it does whatever it needs to.
the file service should be de-coupled of database and controller logic as he has nothing to do with them.

what do you think?

alonp99 (approved):

PR #290, opened 2023-05-11
Add`resolvedAt` to `WorkflowRuntimeData` model

### LINKED ISSUE #275: Add`resolvedAt` to `WorkflowRuntimeData` model
* Add `resolvedAt` column to the `WorkflowRuntimeData` table
* Fill it when the workflow enters a final state

MatanYadaev:
Closes #275

### DIFF
--- a/services/workflows-service/src/workflow/dtos/workflow-definition-update-input.ts
+++ b/services/workflows-service/src/workflow/dtos/workflow-definition-update-input.ts
@@ -20,4 +20,5 @@ export class WorkflowDefinitionUpdateInput {
   context?: InputJsonValue;
 
   status?: string;
+  resolvedAt?: Date | null;
 }
--- a/services/workflows-service/src/workflow/workflow.service.ts
+++ b/services/workflows-service/src/workflow/workflow.service.ts
@@ -126,12 +126,11 @@ export class WorkflowService {
     // @ts-ignore
     const isFinal = workflow.definition?.states?.[currentState]?.type === 'final';
 
-    if (isFinal) {
-      data.resolvedAt = new Date();
-    }
-
     const updateResult = await this.workflowRuntimeDataRepository.updateById(workflowRuntimeId, {
-      data,
+      data: {
+        ...data,
+        resolvedAt: isFinal ? new Date() : undefined,
+      },
     });
 
     if (isFinal) {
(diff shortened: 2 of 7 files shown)


services/workflows-service/src/workflow/workflow.service.ts
```diff
@@ -174,7 +177,7 @@ export class WorkflowService {
         },
       }));
     businessId &&
-      (await this.businessRepository.updateById(businessId as string, {
+      (await this.businessRepository.updateById(businessId, {
```
MatanYadaev, line 180: It was automatically fixed by ESLint because `businessId` is also inferred as `string`, so the assertion is redundant.

services/workflows-service/src/workflow/workflow.service.ts
```diff
@@ -115,18 +115,26 @@ export class WorkflowService {
...
       runtimeData.workflowDefinitionId,
     );
+
     // eslint-disable-next-line @typescript-eslint/ban-ts-comment
     // @ts-ignore
-    if (workflow.definition?.states?.[currentState]?.type === 'final') {
+    const isFinal = workflow.definition?.states?.[currentState]?.type === 'final';
+
+    if (isFinal) {
+      data.resolvedAt = new Date();
```
Omri-Levy, line 130: Let's not mutate `data`. The line where `data.context` is assigned will also be fixed in the upcoming tech-debt reductions.

services/workflows-service/src/workflow/dtos/workflow-definition-update-input.ts
```diff
@@ -20,4 +20,5 @@ export class WorkflowDefinitionUpdateInput {
   context?: InputJsonValue;
 
   status?: string;
+  resolvedAt?: Date | null;
```
Omri-Levy, line 23: Please add the `class-validator` and `@nestjs/swagger` decorators for `resolvedAt`. 🙏

MatanYadaev: It's not relevant for the current use case.

class-validator - The client shouldn't be able to update this property.
Swagger - The property isn't returned in the HTTP response.

Omri-Levy (approved):

PR #299, opened 2023-05-15
Filter cases by assignee

MatanYadaev:
https://github.com/ballerine-io/ballerine/assets/13586343/9bdd4a8a-5dd7-4078-9c5c-be6138cad621


apps/backoffice-v2/src/components/pages/Individuals/Individuals.route.tsx
```diff
@@ -33,10 +35,8 @@ const BusinessesSearchSchema = SearchSchema.extend({
   sortBy: z.enum(['website', 'address']).optional().catch('website'),
   filter: z
     .object({
-      approvalState: z
-        .array(z.enum([States]))
-        .optional()
-        .catch([]),
+      approvalState: z.array(z.enum(States)).optional().catch([]),
```
MatanYadaev, line 38: Fixed an unrelated bug here: changed from `z.enum([States])` to `z.enum(States)` (not wrapping `States` with an array)

apps/backoffice-v2/src/hooks/useFilter/filter.ts
```diff
@@ -24,13 +24,16 @@ export const filter = <TArray extends AnyArray>({
...
     return terms?.every(term => {
       if (!filter?.[term]?.length) return true;
 
       return filter?.[term]?.some(termValue => {
-        return item?.[term]?.includes(termValue);
+        if (typeof item?.[term]?.includes === 'function') {
+          return item?.[term]?.includes(termValue);
+        }
+
+        return item?.[term] === termValue;
```
MatanYadaev, line 36: Refactored this function to support types that don't have `.includes` method, in my specific case I need it for comparing nulls.

apps/backoffice-v2/src/lib/zod/schemas/end-users.ts
```diff
@@ -14,7 +14,7 @@ export const EndUsersListSchema = z
       email: z.string().nullable().default(''),
       phone: z.string().nullable().default(''),
       approvalState: z.enum(States).default(State.PROCESSING),
-      assignedTo: z.string().default(''),
+      assignedTo: z.string().nullable().default(null),
```
MatanYadaev, line 17: The "empty state" of `assignedTo` should be null instead of an empty string. When it's null it's obvious that it's unassigned.

services/workflows-service/prisma/schema.prisma
```diff
@@ -181,15 +185,15 @@ model Policy {
 }
 
 model Filter {
-  id String @id @default(cuid())
+  id        String   @id @default(cuid())
   // Name of the filter i.e. "KYC"
-  name String @unique
```
Omri-Levy, line 186: Why was `@unique` removed?

apps/backoffice-v2/src/hooks/useFilter/filter.ts
Omri-Levy: The structure for a filter is for example  `{ approvalState:  ['NEW', 'APPROVED'] }` which would mean the check is `['NEW', 'APPROVED'].includes('NEW')`, did the check become `null === null` for asignee?

services/workflows-service/prisma/schema.prisma
```diff
@@ -16,8 +16,9 @@ model User {
   password  String
   roles     Json
 
-  createdAt DateTime @default(now())
-  updatedAt DateTime @updatedAt
+  createdAt           DateTime              @default(now())
+  updatedAt           DateTime              @updatedAt
+  WorkflowRuntimeData WorkflowRuntimeData[]
```
Omri-Levy, line 21: Change to `workflowRuntimeData` please. :pray:

apps/backoffice-v2/src/hooks/useFilter/filter.ts
MatanYadaev: If it's unassigned, then yes, it's `null` and should be compared using `===`.

services/workflows-service/prisma/schema.prisma
MatanYadaev: Not sure, maybe I missed it when I resolved conflicts. Thanks for catching it!

services/workflows-service/prisma/schema.prisma
MatanYadaev: Thanks!

PR #305, opened 2023-05-16
API versioning

### LINKED ISSUE #304: Add API versioning

MatanYadaev:
Closes #304

### DIFF
--- a/services/workflows-service/src/app.module.ts
+++ b/services/workflows-service/src/app.module.ts
@@ -1,4 +1,4 @@
-import { MiddlewareConsumer, Module, Scope } from '@nestjs/common';
+import { MiddlewareConsumer, Module, RequestMethod, Scope } from '@nestjs/common';
 import { APP_INTERCEPTOR } from '@nestjs/core';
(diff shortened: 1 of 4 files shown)


services/workflows-service/src/app.module.ts
```diff
@@ -62,6 +62,9 @@ import { SessionAuthMiddleware } from '@/auth/session-auth.middleware';
 })
 export class AppModule {
   configure(consumer: MiddlewareConsumer) {
-    consumer.apply(SessionAuthMiddleware).forRoutes('internal/*');
+    consumer.apply(SessionAuthMiddleware).forRoutes({
+      path: 'v(\\d)+/internal/*',
+      method: RequestMethod.ALL,
```
alonp99, line 67: can you explain this change?

MatanYadaev: It applies the middleware on any API version endpoint. It's a regex, v*/internal/*, for example, v3/internal/example

alonp99 (approved):

Issue #418, opened 2023-05-24
Make every endpoint session-guarded by default


MatanYadaev:
1. Make every endpoint session-guarded by default
2. Allow other guards to override the default guard

PR #440, opened 2023-05-28
Session guard by default

MatanYadaev:
#418


services/workflows-service/src/auth/default-auth.guard.ts
```diff
@@ -1,21 +0,0 @@
-import { ExecutionContext, Injectable } from '@nestjs/common';
-import { Reflector } from '@nestjs/core';
-import { IS_PUBLIC_KEY } from '@/common/decorators/public.decorator';
-import { JwtAuthGuard } from './jwt/jwt-auth.guard';
-
-@Injectable()
-export class DefaultAuthGuard extends JwtAuthGuard {
```
MatanYadaev, line 7: It's no longer used.

alonp99 (approved):

PR #451, opened 2023-05-29
feat(generated assignee-asigned-guard): assign endpoint

Blokh:
generated assignee-asigned-guard for document desition updating

### Description
validate before update on backend controller side - if the document update is decision && is assigned to me

- added gaurd over patch request (AssigneeAsignedGuard)
-  added assign workflow endpoint (assignWorkflowById)
-  updated backoffice to use the endpoint
Tested all logics using curls


package.json
```diff
@@ -28,8 +28,8 @@
...
-    "kyb-manual-review-example": "nx run @ballerine/common:build && nx run @ballerine/workflows-service:setup && cross-env ENV_FILE_NAME=.env.example VITE_POLLING_INTERVAL=3 VITE_EXAMPLE_TYPE=kyb concurrently \"nx run @ballerine/workflows-service:dev\" \"wait-on http://localhost:3000/api/v1/_health/ready && nx run-many --target=dev --projects=@ballerine/headless-example,@ballerine/backoffice-v2\"",
+    "kyc-manual-review-example": "nx run @ballerine/common:build && nx run @ballerine/workflows-service:setup && cross-env ENV_FILE_NAME=.env.example VITE_POLLING_INTERVAL=false VITE_EXAMPLE_TYPE=kyc concurrently \"nx run @ballerine/workflows-service:dev\" \"wait-on http://localhost:3000/api/v1/_health/ready && nx run-many --target=dev --projects=@ballerine/headless-example,@ballerine/backoffice-v2\"",
```
alonp99, line 31: It will break the example flow, change back please

services/workflows-service/scripts/clean.ts
```diff
@@ -18,7 +18,7 @@ async function clean() {
   const tables = await getTables(prisma);
   const types = await getTypes(prisma);
   await dropTables(prisma, tables);
-  await dropTypes(prisma, types);
+  await dropTypes(prisma, types.sort());
```
Omri-Levy, line 21: The sort here is a workaround for an error that happens on the db clean step.

Omri-Levy: I suggest finding a more reliable way of ensuring the order in which the types are getting removed (dependency wise). I still feel like `clean.ts` can be boiled down to some drop db operation or `TRUNCATE`. Prisma might have an API for cleaning the db.

package.json
```diff
@@ -30,7 +30,7 @@
...
     "kyb-manual-review-example": "nx run @ballerine/common:build && nx run @ballerine/workflows-service:setup && cross-env ENV_FILE_NAME=.env.example VITE_POLLING_INTERVAL=3 VITE_EXAMPLE_TYPE=kyb concurrently \"nx run @ballerine/workflows-service:dev\" \"wait-on http://localhost:3000/api/v1/_health/ready && nx run-many --target=dev --projects=@ballerine/headless-example,@ballerine/backoffice-v2\"",
-    "api-manual-review-example": "nx run @ballerine/common:build && nx run @ballerine/workflows-service:setup && cross-env ENV_FILE_NAME=.env.example VITE_POLLING_INTERVAL=false VITE_EXAMPLE_TYPE=kyb concurrently \"nx run @ballerine/workflows-service:dev\" \"wait-on http://localhost:3000/api/v1/_health/ready && nx run-many --target=dev --projects=@ballerine/backoffice-v2\"",
+    "api-manual-review-example": "nx run @ballerine/common:build && nx run @ballerine/workflows-service:setup && cross-env ENV_FILE_NAME=.env.example VITE_POLLING_INTERVAL=3 VITE_EXAMPLE_TYPE=kyb concurrently \"nx run @ballerine/workflows-service:dev\" \"wait-on http://localhost:3000/api/v1/_health/ready && nx run-many --target=dev --projects=@ballerine/backoffice-v2\"",
```
alonp99, line 33: What is the reason for changing this?

services/workflows-service/src/auth/assignee-asigned-guard.service.ts
```diff
@@ -0,0 +1,19 @@
+import { CanActivate, ExecutionContext, Injectable } from '@nestjs/common';
+import { Request } from 'express';
+import { WorkflowService } from '@/workflow/workflow.service';
+import { WorkflowDefinitionUpdateInput } from '@/workflow/dtos/workflow-definition-update-input';
+import { WorkflowRuntimeData } from '@prisma/client';
+import console from 'console';
```
alonp99, line 6: remove

services/workflows-service/src/workflow/workflow.service.ts
```diff
@@ -273,6 +274,15 @@ export class WorkflowService {
     return updateResult;
   }
 
+  async assignWorkflowToAssignee(workflowRuntimeId: string, data: WorkflowWhereAssignInput) {
```
alonp99, line 277: Naming cloud be better here, maybe `assignWorkflowToUser`

services/workflows-service/src/workflow/workflow.service.ts
```diff
@@ -273,6 +274,15 @@ export class WorkflowService {
...
+    const updatedWorkflowRuntime = await this.workflowRuntimeDataRepository.updateById(
+      workflowRuntimeId,
+      { data: data },
```
alonp99, line 280: If the scope of the update is the assignee why we get the whole data and not just the assignee, and explicitly changing only this column?

Less room for errors

services/workflows-service/src/auth/assignee-asigned-guard.service.ts
```diff
@@ -0,0 +1,19 @@
...
+
+@Injectable()
+export class AssigneeAsignedGuard implements CanActivate {
```
alonp99, line 9: This guard relays on workflow data, the name should probably suggest it

`WorkflowAssigneeGuard`

package.json
Blokh: My bad, should've replaced it, and should've reviewed the PR before.

services/workflows-service/src/workflow/workflow.service.ts
Blokh: the class validator of WorkflowWhereAssignInput validates the data being updated.
do you think it's preferable to 
{ data: { assigneeId: data.assigneeId }? even though WorkflowWhereAssignInput formats the updated data?

alonp99: I dont have to pass WorkflowWhereAssignInput, you could just pass `assgineeId` - I think the Input should stay in the controller level - other users of this functions only cares about the assginee id and should not be aware of other structure

alonp99 (approved):

alonp99 (approved):

PR #454, opened 2023-05-29
CD: Run migrations

MatanYadaev:


services/workflows-service/Dockerfile
```diff
@@ -1,5 +1,7 @@
 FROM node:18.12.1 AS base
 
+RUN apt-get update && apt-get install -y --no-install-recommends dumb-init
```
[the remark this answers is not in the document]
alonp99: gj

services/workflows-service/package.json
```diff
@@ -10,6 +10,7 @@
     "lint": "eslint . --fix",
     "start": "nest start",
     "dev": "npm run start:watch",
+    "prod": "npm run db:migrate-up && node dist/main",
```
alonp99, line 13: Do the examples work after this change? (try this one which is relevant for dev https://github.com/ballerine-io/ballerine/blob/main/websites/docs/src/pages/en/learn/back-office.md)

MatanYadaev: It doesn't affect the example at all. It's a new command that is used only when we initiate the container, which happens only on deployment.

PR #480, opened 2023-06-01
Blokh/feat/default case filter assignee

Blokh:
### Description
removed pre-filter creation as it does not support the user fetching/async tasks.
updated it to be set in the onLoad
added fefault filters on assigneeId


alonp99 (approved):

apps/backoffice-v2/src/routes/Entities/pre-search-filters.ts
```diff
@@ -7,24 +8,29 @@ const sharedPreSearchFilters = {
...
-    entity: 'individuals' as const,
-    filter: {
-      approvalState: [],
-      assigneeId: [],
-      caseStatus: [CaseStatus.ACTIVE],
+    individuals: {
+      sortBy: 'caseCreatedAt' as const,
+      entity: 'individuals' as const,
+      filter: {
+        approvalState: [State.PROCESSING],
```
alonp99, line 29:
```suggestion
        approvalState: [],
```

alonp99 (approved):

alonp99 (approved):

PR #501, opened 2023-06-06
Dynamic webhook config

alonp99:


services/workflows-service/src/events/document-changed-webhook-caller.ts
```diff
@@ -85,8 +96,10 @@ export class DocumentChangedWebhookCaller {
           },
         },
       );
-    } catch (error) {
-      this.logger.error('Failed to send webhook', { id, error });
+
+      console.log('Result webhook', res.status, res.statusText, res.data);
```
alonp99, line 100: ill change to logger

services/workflows-service/scripts/run-filter.ts
```diff
@@ -0,0 +1,56 @@
+import { PrismaClient } from '@prisma/client';
+
+export async function fliterQuery() {
```
MatanYadaev, line 3: What is this file? Should it be in the source code?

alonp99: It's a script/tool I used it to try fliters before creating them, should be part of the source code (similar to seed.ts)

services/workflows-service/src/business/business.repository.ts
```diff
@@ -40,6 +40,13 @@ export class BusinessRepository {
     });
   }
 
+  async getCorrelationIdById(id: string): Promise<{ correlationId: string | null } | null> {
+    return await this.prisma.business.findUnique({
+      where: { id },
+      select: { correlationId: true },
+    });
+  }
```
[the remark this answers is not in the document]
alonp99: In this scenario it shouldnt throw since null its a valid result (we check if there is an endUser or Busniess id - one of them will return null)

PR #508, opened 2023-06-06
refactor(db-schema): remove enduser unneccesary fields

alonp99:


services/workflows-service/src/filter/dtos/temp-zod-schemas.ts
```diff
@@ -481,7 +481,7 @@ export const BusinessFilterSchema = FilterSchema.extend({
     .object({
       select: BusinessSelectSchema.strict()
         .refine(
-          value => Object.keys(value).length > 0,
+          (value: {}) => Object.keys(value).length > 0,
```
Omri-Levy, line 484: Note `{}` means any type that is not `null` or `undefined`.

PR #516, opened 2023-06-08
Cases pagination, sorting, filtering, and search

MatanYadaev:


apps/backoffice-v2/src/common/components/organisms/Header/Header.Navbar.tsx
```diff
@@ -53,10 +53,10 @@ export const Navbar: FunctionComponent = () => {
       <ul className={`menu menu-compact w-full space-y-2`}>
         {filters?.map(({ id, name, entity }) => (
           <NavItem
-            href={`/en/case-management/entities?entity=${entity}&filterId=${id}&filterName=${name}`}
+            href={`/en/case-management/entities?entity=${entity}&filterId=${id}`}
             key={id}
             className={ctw(`capitalize`, {
-              'bg-muted font-bold': name === searchParams?.filterName,
+              'bg-muted font-bold': id === searchParams?.filterId,
```
MatanYadaev, line 59: Unrelated fix:
We added the `filterName` to the query params just for comparing the current filter. It was unnecessary because we already have `filterId`.

apps/backoffice-v2/src/common/hooks/useSearchParamsByEntity/validation-schemas.ts
```diff
@@ -3,11 +3,10 @@ import { CaseStatus, CaseStatuses, States } from '../../enums';
 
 export const SearchSchema = z.object({
   sortDir: z.enum(['asc', 'desc']).catch('desc'),
-  pageSize: z.coerce.number().int().positive().catch(10),
+  limit: z.coerce.number().int().positive().catch(50),
```
MatanYadaev, line 6: Increased the page size because the search is in the client only, so at least it searches a big amount of cases.

services/workflows-service/src/prisma/prisma.service.ts
```diff
@@ -3,6 +3,12 @@ import { PrismaClient } from '@prisma/client';
 
 @Injectable()
 export class PrismaService extends PrismaClient implements OnModuleInit {
+  constructor() {
+    super({
+      errorFormat: 'pretty',
```
MatanYadaev, line 8: It will add colors to the Prisma errors.

services/workflows-service/src/workflow/utils/toPrismaOrderBy.ts
```diff
@@ -21,23 +21,23 @@ export const toPrismaOrderBy = <
...
+      } as { [K in TColumn]: TDirection },
     };
   }
 
   return {
-    [column]: direction,
+    business: {
+      [column]: direction,
+    },
   };
```
MatanYadaev, line 42: Same logic, just a bit more readable.

services/workflows-service/src/workflow/utils/toPrismaWhere.ts
```diff
@@ -0,0 +1,36 @@
+import { Prisma, WorkflowRuntimeDataStatus } from '@prisma/client';
+
+type Filters = {
+  assigneeId?: (string | null)[];
+  status?: WorkflowRuntimeDataStatus[];
+};
+
+export const toPrismaWhere = (filters: Filters): Prisma.WorkflowRuntimeDataWhereInput => {
```
MatanYadaev, line 8: This function, and the `toPrismaOrderBy`, are workflow-specific. They won't fit for users, etc.
Should I rename them or place them in another folder to make it clearer?

alonp99 (approved):

PR #539, opened 2023-06-15
Blokh/feat/trans valid interface

Blokh:
### Description
Added the functionality of transformation and validation interfaces for plugins


packages/workflow-core/src/lib/utils/context-transformers/types.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export type TTransformationLogic = string;
+export abstract class BaseContextTransformer {
+  abstract name: string;
+  type = 'context-transformer';
+  context: TContext;
+  transformationLogic: TTransformationLogic;
+  options: {};
+
+  protected constructor(context: TContext, transformationLogic: TTransformationLogic, options: {}) {
```
yairper, line 11: 1. it looks like an interface with a constructor,
2. that's a bad practice to have a constructor in the base class as one of the things which separates  implementations is the constructor itself
3. also it makes the code less readable as you need to go to the base class every time you wanna understand the child implementation

packages/workflow-core/src/lib/utils/context-validator/json-schema-validator.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export class JsonSchemaValidator extends BaseSchemaValidator {
+  name = "json-schema-validator";
+
+  validate(): TSchemaValidatorResponse {
+    const ajv = new Ajv(this.options);
+    const validator = ajv.compile(this.validationLogic);
+
+    if (!validator(this.data)) {
+      return Promise.resolve({ isValid: false, errorMessage: this.errorMessage || 'Bad Request' });
```
yairper, line 12: I guess that should be Promise.reject, in that case `isValid: false` is pretty obvious, just return the message...

packages/workflow-core/src/lib/utils/context-transformers/qj-transformer.ts
```diff
@@ -0,0 +1,10 @@
+import { run } from 'node-jq';
+import { BaseContextTransformer, TContext } from './types';
+
+export class JQTransformer extends BaseContextTransformer {
+  name = 'jq-transformer';
+
+  transform(): any {
```
yairper, line 7: given the interface it returns a promise not any

packages/workflow-core/src/lib/utils/types.ts
```diff
@@ -0,0 +1,8 @@
+import { JQTransformer } from './context-transformers/qj-transformer';
+import { JsonSchemaValidator } from './context-validator/json-schema-validator';
+import { AnyRecord } from '@ballerine/common';
+
+export type TContext = AnyRecord;
+
+export type TTransformers = JQTransformer;
```
yairper, line 7: not in use

packages/workflow-core/src/lib/utils/types.ts
```diff
@@ -0,0 +1,8 @@
...
+export type TValidators = JsonSchemaValidator;
```
yairper, line 8: not in use

yairper (changes requested):

yairper:
we need to talk about such PRs as this code 
1. does not implement any user scenario and we only imagine how allegedly it gonna be used, which could change
2. we are going to merge code which does not add any increment to the product we are at a risk of adding dead/wrong code

packages/workflow-core/src/lib/utils/context-validator/json-schema-validator.ts
Blokh: I somewhat disagree.
validator should return is valid or not. the validation should be on that field.

packages/workflow-core/src/lib/utils/types.ts
Blokh: this is what you should be using for transformation and validation

packages/workflow-core/src/lib/utils/types.ts
Blokh: this is what you should be using for transformation and validation

packages/workflow-core/src/lib/utils/context-transformers/types.ts
Blokh: yeah inheritance has its issues.

but for now, it is built as intendant. maybe later we'll have to change the constructor.
later on, if the abstract constructor will require additional functionalities/requirements by any other "transformer". we'll update the logic. for now it works as intendant

packages/workflow-core/src/lib/utils/context-transformers/types.ts
```diff
@@ -0,0 +1,18 @@
...
+    this.context = context;
+    this.transformationLogic = transformationLogic;
+    this.options = options;
+  }
+
+  abstract transform(): Promise<any>;
```
Blokh, line 17: pass the necessary information for transformation

packages/workflow-core/src/lib/utils/context-validator/json-schema-validator.ts
```diff
@@ -0,0 +1,18 @@
+import { BaseSchemaValidator, TSchemaValidatorResponse } from './types';
+import Ajv from 'ajv';
+
+export class JsonSchemaValidator extends BaseSchemaValidator {
+  name = "json-schema-validator";
+
+  validate(): TSchemaValidatorResponse {
```
Blokh, line 7: async

PR #625, opened 2023-07-02
Omri levy/feat/child workflows

Omri-Levy:
### Description
Elaborate on the subject, motivation, and context.

### Related issues

### Breaking changes

### Examples and references

### Checklist
- [] I have read the [contribution guidelines](CONTRIBUTING.md) of this project
- [] I have read the [style guidelines](STYLE_GUIDE.md) of this project
- [] I have performed a self-review of my own code
- [] I have commented my code, particularly in hard-to-understand areas
- [] I have made corresponding changes to the documentation
- [] My changes generate no new warnings and errors
- [] New and existing tests pass locally with my changes


packages/workflow-core/src/lib/types.ts
```diff
@@ -29,15 +29,83 @@ export interface WorkflowContext {
...
 
+export type JMESPathString = string;
+
+export interface ChildWorkflow {
+  name: string;
+  runtimeId: string;
+  definitionId: string;
+  definitionVersion: 1;
+  stateNames: Array<string>;
+  contextToCopy: JMESPathString;
```
Blokh, line 40: I would suggest keeping the same logic as the API plugin with the request transformer and response transformer. so we would have consistency api-wise

packages/workflow-core/src/lib/types.ts
```diff
@@ -29,15 +29,83 @@ export interface WorkflowContext {
...
+  callbackInfo: {
```
Blokh, line 41: does it have on-fail callback ? 
should it have any kind of validations?

Blokh:
interface-wise what are the differences between child workflow and parent workflow.
From my point of view they are the same except the "governing" entity and how they are handling the onDone.

in parent workflow, he's the authority. therefore he does not service anyone

in child workflow. the parent workflow is the authority and therefore the parent is the governing entity and the child workflow is servicing the parent.

if you do not handle the child workflow like a parent workflow  - you will have to re-write the runner for the child.

packages/workflow-core/src/lib/workflow-runner.ts
```diff
@@ -317,20 +339,57 @@ export class WorkflowRunner {
...
-        workflowId: snapshot.machine?.id,
+        workflowId: postSendSnapshot.machine?.id,
       })(this.#__context, event);
     }
+
+    if (this.#__onInvokeChildWorkflow && childWorkflowsToInvoke?.length) {
+      await Promise.all(
+        childWorkflowsToInvoke?.map(
+          async ({ definitionId, runtimeId, name, definitionVersion, initOptions }) => {
+            await this.#__onInvokeChildWorkflow?.({
```
alonp99, line 367: const { workflowRuntimeId, error } = await ...

packages/workflow-core/src/lib/workflow-runner.ts
```diff
@@ -317,20 +375,109 @@ export class WorkflowRunner {
...
+                data: undefined,
+                error,
+                runtimeId,
+              };
+            }
+          },
+        ),
+      );
+      const childWorkflows = results?.map(result => {
+        if (result?.status !== 'fulfilled') return;
```
alonp99, line 430: I think we should leave a todo and corresponding task to handle non fulfilled promises and errors (log them and send event to the host)

packages/workflow-core/src/lib/workflow-runner.ts
```diff
@@ -317,20 +375,109 @@ export class WorkflowRunner {
...
+          data: result?.value?.data,
+          error: undefined,
+        };
+      });
+
+      this.#__context.childWorkflows = childWorkflows;
+      service.send({
+        type: 'UPDATE_CONTEXT',
+        payload: {
+          childWorkflows,
```
alonp99, line 443: Merge logic?

packages/workflow-core/src/lib/workflow-runner.ts
```diff
@@ -317,20 +375,109 @@ export class WorkflowRunner {
...
+        if (result?.status !== 'fulfilled') return;
+
+        return {
+          runtimeId: result?.value?.runtimeId,
+          data: result?.value?.data,
+          error: undefined,
+        };
+      });
+
+      this.#__context.childWorkflows = childWorkflows;
```
alonp99, line 439: it seems like we are updating explicitly + with event

packages/workflow-core/src/lib/workflow-runner.ts
Omri-Levy: > 

Here `try` and `catch` are in use so there shouldn't be a real `rejected` status from `allSettled`.

PR #802, opened 2023-08-09
fix(webhooks): fix webhookd regestration bug

alonp99:


services/workflows-service/src/events/get-webhook-info.ts
```diff
@@ -8,10 +8,11 @@ export const getWebhookInfo = (
   NODE_ENV: string | undefined,
   WEBHOOK_URL: string | undefined,
   WEBHOOK_SECRET: string | undefined,
+  event: string,
```
Omri-Levy, line 11: Might be an existing type, could be `TEvent` which is exported from the nearby types file.

Omri-Levy (approved):

PR #872, opened 2023-08-17
Verify signature of Unified API callbacks

### LINKED ISSUE #869: Verify signature of Unified API callbacks

MatanYadaev:
Closes #869


apps/backoffice-v2/package.json
```diff
@@ -49,7 +49,7 @@
     "build-storybook": "storybook build"
   },
   "dependencies": {
-    "@ballerine/common": "0.5.19",
+    "@ballerine/common": "0.5.20",
```
Omri-Levy: Upgrade to 0.5.22

services/workflows-service/scripts/workflows/parent-kyb-kyc-session-workflow.ts
```diff
@@ -144,7 +145,8 @@ export const parentKybWithSessionWorkflowDefinition = {
               from: 'no-reply@ballerine.com',
               receivers: [entity.data.additionalInfo.mainRepresentative.email],
               templateId: 'd-7305991b3e5840f9a14feec767ea7301',
-              revisionReason: documents[].decision[].revisionReason | [0]
+              revisionReason: documents[].decision[].revisionReason | [0],
+              adapter: '${env.MAIL_DRIVER}'
```
alonp99: Im not sure adding an env here makes sense, we would probably want to use it more dynamically instead of deciding what driver to use on the seed run or when inserting a definition to the db

what I suggest is to add configuration to the context and then reference something like `config.mailDriver` from here

alonp99 (approved):

PR #876, opened 2023-08-17
Fake email adapter

### LINKED ISSUE #875: Fake email adapter
Add a new fake email adapter (alongside Sendgrid).

MatanYadaev:
Closes #875


packages/workflow-core/src/lib/plugins/external-plugin/email-plugin.ts
```diff
@@ -46,6 +46,16 @@ export class EmailPlugin extends ApiPlugin {
       ...templateId,
     };
 
+    if (payload.adapter === 'log') {
+      console.log('Email payload: ', emailPayload);
+
+      return {
```
Blokh, line 52: extract to method that says what those lines do.
e.g. mockedReturnOfFetchResponse

Omri-Levy: I see more value in having the code visible and not abstracted away. (specifically in this case)

Omri-Levy (approved):

alonp99 (approved):

PR #951, opened 2023-08-22
Fix backoffice images in environments without signed url setup

### LINKED ISSUE #1200: [BAL-298] Fix backoffice images in environments without signed url setup
null

<sub>[BAL-298](https://linear.app/ballerine/issue/BAL-298/fix-backoffice-images-in-environments-without-signed-url-setup)</sub>

alonp99:


tomer-shvadron (approved):

apps/backoffice-v2/src/domains/storage/fetchers.ts
```diff
@@ -39,7 +37,7 @@ export const fetchFileInfoById = async (fileId: string) => {
 };
 export const fetchFileById = async (fileId: string) => {
   const fileInfo = await fetchFileInfoById(fileId);
-  if (fileInfo.fileNameInBucket) {
+  if (fileInfo.fileNameInBucket && import.meta.env.VITE_FETCH_SIGNED_URL === 'TRUE') {
```
Omri-Levy, line 40: Is this intentional? (all caps)

Blokh: by default it's true i think lowercased. it should be downcased

Blokh: yes it is uppercased by default

PR #1021, opened 2023-09-01
Add GH and UG document schemas

### LINKED ISSUE #1210: Fix document schemas

MatanYadaev:


apps/backoffice-v2/src/pages/Entity/components/EditableDetails/EditableDetails.tsx
```diff
@@ -123,68 +123,94 @@ export const EditableDetails: FunctionComponent<IEditableDetails> = ({
             'grid-cols-3': id === 'entity-details',
           })}
         >
-          {formData?.map(({ title, isEditable, type, format, pattern, value, dropdownOptions }) =>
```
MatanYadaev, line 126: Wrapped the return value by curly brackets so there are lots of difference, but I'll mark what specifically I changed in this file.

apps/backoffice-v2/src/pages/Entity/components/EditableDetails/EditableDetails.tsx
```diff
@@ -123,68 +123,94 @@ export const EditableDetails: FunctionComponent<IEditableDetails> = ({
...
+                  render={({ field }) => {
+                    let inputType: string;
+
+                    if (!format) {
+                      inputType = type === 'string' ? 'text' : type;
+                    } else if (format === 'date-time') {
+                      inputType = 'datetime-local';
+                    } else {
+                      inputType = format;
+                    }
```
MatanYadaev, line 152: This is new code

apps/backoffice-v2/src/pages/Entity/components/EditableDetails/EditableDetails.tsx
```diff
@@ -123,68 +123,94 @@ export const EditableDetails: FunctionComponent<IEditableDetails> = ({
...
+                              })}
+                            </SelectContent>
+                          </Select>
+                        ) : (
+                          <FormControl>
+                            <Input
+                              type={inputType}
+                              {...(inputType === 'datetime-local' && { step: '1' })}
+                              {...(minimum && { min: minimum })}
+                              {...(maximum && { max: maximum })}
```
MatanYadaev, line 187: And this is a new code

packages/common/src/schemas/documents/workflow/documents/schemas/UG.ts
```diff
@@ -0,0 +1,140 @@
...
+import { Type } from '@sinclair/typebox';
+
+export const getUgandaDocuments = (): TDocument[] => {
+  const TypeAlphanumericString = Type.String({ pattern: '^[a-zA-Z0-9]*$' });
+  const TypePastDate = Type.String({
+    format: 'date',
+    formatMaximum: new Date().toISOString().split('T')[0], // @TODO: Test this
+  });
+
+  const TypeStringAtLeastOneWord = Type.String({ minLength: 1 });
```
Omri-Levy: Let's extract to some shared schemas file.

MatanYadaev: I don't think these types should be reusable as future cases might want to add additional config properties. I'm okay with setting these types ad-hoc when you need them. In one place "at least one word" can be 2 characters, and in another it could be 1. It's a tough thing managing reuseable code that needs adjustments per usage.
The benefit of making these reusable isn't high enough IMO.
WDYT?

Omri-Levy (approved):

PR #1022, opened 2023-09-03
Blokh/feat/workflow definition mt

### LINKED ISSUE #1247: workflow definition mt

Blokh:
### Description
added workflow definition clone logic and scoping of workflow definition

### DIFF
--- a/services/workflows-service/src/workflow/workflow-definition.repository.ts
+++ b/services/workflows-service/src/workflow/workflow-definition.repository.ts
@@ -53,7 +53,7 @@ export class WorkflowDefinitionRepository {
     projectId?: string,
   ): Promise<WorkflowDefinition> {
     return await this.prisma.workflowDefinition.findFirstOrThrow({
-      where: { id, isTemplate: true, projectId: projectId ?? null },
+      where: { id, isPublic: true, projectId: projectId ?? null },
       ...args,
     });
   }
--- a/services/workflows-service/src/workflow/workflow-definition.repository.ts
+++ b/services/workflows-service/src/workflow/workflow-definition.repository.ts
@@ -45,6 +45,19 @@ export class WorkflowDefinitionRepository {
     );
   }
 
+  async findTemplateByIdUnscoped<
+    T extends Omit<Prisma.WorkflowDefinitionFindFirstOrThrowArgs, 'where'>,
+  >(
+    id: string,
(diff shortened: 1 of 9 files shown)


services/workflows-service/src/workflow/workflow-definition.repository.ts
```diff
@@ -45,6 +45,19 @@ export class WorkflowDefinitionRepository {
...
 
+  async findTemplateByIdUnscoped<
+    T extends Omit<Prisma.WorkflowDefinitionFindFirstOrThrowArgs, 'where'>,
+  >(
+    id: string,
+    args: Prisma.SelectSubset<T, Omit<Prisma.WorkflowDefinitionFindFirstOrThrowArgs, 'where'>>,
+    projectId?: string,
+  ): Promise<WorkflowDefinition> {
+    return await this.prisma.workflowDefinition.findFirstOrThrow({
+      where: { id, isTemplate: true, projectId: projectId ?? null },
```
MatanYadaev: Remove `projectId` as it's irrelevant here.
```suggestion
  ): Promise<WorkflowDefinition> {
    return await this.prisma.workflowDefinition.findFirstOrThrow({
      where: { id, isTemplate: true },
```

services/workflows-service/src/workflow/workflow.service.ts
```diff
@@ -149,9 +150,40 @@ export class WorkflowService {
...
+    };
+
+    const workflowDefinition = await this.workflowDefinitionRepository.findTemplateByIdUnscoped(
+      data.id,
+      { select },
+    );
+
+    return await this.workflowDefinitionRepository.create(
+      // @ts-expect-error - types of workflow definition does not propagate to the prisma creation type
+      { data: { ...workflowDefinition, name: data.name }, select },
```
MatanYadaev: `workflowDefinition` has an `id` property, it won't use it?

MatanYadaev (approved):

services/workflows-service/src/workflow/workflow.service.ts
Blokh: nop, it is not being selected in the select args.

services/workflows-service/src/workflow/workflow-definition.repository.ts
Blokh: I somewhat agree.
but there's some kind of logic of creating "templates" (publics) to specific customers.
e.g. nuvei will have multiple of the same template

PR #1046, opened 2023-09-05
Workflow tokens table

### LINKED ISSUE #1048: Create WorkflowToken table

### LINKED ISSUE #1214: Workflow tokens table

MatanYadaev:
Closes #1048


services/workflows-service/src/workflow/workflow.controller.external.ts
```diff
@@ -293,4 +298,54 @@ export class WorkflowControllerExternal {
...
+      workflowRuntimeDataId: workflow[0].workflowRuntimeData.id,
+      endUserId: endUser.id,
+      expiresAt: nowPlus30Days,
+    });
+
+    // @TODO: Send email with token
+
+    return { token: workflowToken.token };
+  }
 }
```
MatanYadaev: This endpoint is temporary. Daniel suggested that the implementation should be more generic, leveraging our workflow infra. The token generation and sending the email should be part of a workflow. But it requires new plugins and new workflow definition, I wanted to implement it ASAP.

services/workflows-service/prisma/migrations/20230905130203_create_workflow_runtime_data_token_table/migration.sql
```diff
@@ -0,0 +1,26 @@
+-- CreateTable
+CREATE TABLE "WorkflowRuntimeDataToken" (
+    "id" VARCHAR(500) NOT NULL,
+    "token" VARCHAR(500) NOT NULL,
```
Blokh: add index over token - as it is the one that will be queried.

p.s. why is it varchar and not text as everything else?

MatanYadaev: 1. There is a unique index on it
2. Updated to text instead of varchar just for consistency. Anyway, I believe the best practice is the use varchar unless you have to use text.

Blokh (approved):

PR #1072, opened 2023-09-11
tomer shvadron/feat/active cases counter

### LINKED ISSUE #1339: backoffice active case counter

tomer-shvadron:
### Description
* Added active cases counter
* Fixed scroll bar on entire backoffice

### Examples and references
image

### DIFF
--- a/apps/backoffice-v2/src/pages/Entities/components/Cases/interfaces.ts
+++ b/apps/backoffice-v2/src/pages/Entities/components/Cases/interfaces.ts
@@ -17,5 +17,5 @@ export interface ICasesProps extends DivComponent {
   onSortBy: ChangeEventHandler<HTMLSelectElement>;
   onSortDirToggle: () => void;
   search: string;
-  caseCounter: number;
+  count: number;
 }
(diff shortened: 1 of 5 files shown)


apps/backoffice-v2/src/pages/Entities/hooks/useEntities/useEntities.tsx
```diff
@@ -94,6 +94,7 @@ export const useEntities = () => {
     onSortDirToggle,
     search,
     cases: searched,
+    caseCounter: data?.meta?.totalItems,
```
MatanYadaev: `count` makes more sense than `counter`. It's a more common convention
```suggestion
    caseCount: data?.meta?.totalItems,
```

apps/backoffice-v2/src/pages/Entities/components/Cases/interfaces.ts
```diff
@@ -17,4 +17,5 @@ export interface ICasesProps extends DivComponent {
   onSortBy: ChangeEventHandler<HTMLSelectElement>;
   onSortDirToggle: () => void;
   search: string;
+  caseCounter: number;
```
MatanYadaev: This property is in the context of cases, so the word `case` is redundant in this property name. The same as you don't call `user.user_email`, you call it `user.email`
```suggestion
  count: number;
```

MatanYadaev (approved):
Good job 🔥

PR #1073, opened 2023-09-11
tomer shvadron/feat/images skeleton

### LINKED ISSUE #1341: backoffice case level image skeleton

tomer-shvadron:
### Description
Added skeleton to block-level image loading

### Examples and references
image


apps/backoffice-v2/src/common/components/organisms/ImageViewer/ImageViewer.SelectedImage.tsx
MatanYadaev, line 1: I think you can remove line 54, it's a condition that is always false.

apps/backoffice-v2/src/domains/storage/fetchers.ts
```diff
@@ -39,10 +39,12 @@ export const fetchFileInfoById = async (fileId: string) => {
 };
 export const fetchFileById = async (fileId: string) => {
   const fileInfo = await fetchFileInfoById(fileId);
+
```
MatanYadaev, line 42: Thanks for the line breaks! 🔥

MatanYadaev, line 42: I would appreciate if you could comment and say in the PR that this is a boyscouting thing. It will make the code review go smoother. 🙏

apps/backoffice-v2/src/common/components/organisms/ImageViewer/ImageViewer.SelectedImage.tsx
```diff
@@ -42,7 +43,9 @@ export const SelectedImage = forwardRef<HTMLImageElement | HTMLIFrameElement, TS
       );
     }
 
-    return (
+    return isPlaceholder ? (
```
Omri-Levy: `isPlaceholder` is meant to be true if an invalid (or no URL) URL is passed, or we are unable to display an image. Skeletons should be visible if either we're fetching the URL for the image or while the asset itself is being downloaded. What's the behaviour for cases with no image?

Omri-Levy: Please refactor to `if (isPlaceholder) { return ... }` 🙂

apps/backoffice-v2/src/pages/Entity/components/Case/Case.Documents.tsx
```diff
@@ -120,15 +120,15 @@ export const Documents: FunctionComponent<IDocumentsProps> = ({ documents, isLoa
...
+                    disabled={isRotatedOrTransformed}
                   >
                     {isCropping && !isLoadingOCR && <CheckSvg className={`p-0.5`} />}
                     {!isCropping && !isLoadingOCR && <span className={`p-0.5`}>OCR</span>}
                   </button>
                 </div>
               </>
             )}
-            <ImageViewer.ZoomButton disabled={isLoading} />
+            {!isLoading && <ImageViewer.ZoomButton />}
```
MatanYadaev, line 131: Suggestion: You can make another fragment that hides all actions at once.
```jsx
{!isLoading && <>
  // Non-PDF code
  // Zoom Button
</>}
```

MatanYadaev (approved):

PR #1079, opened 2023-09-12
Sync Salesforce with KYB data 

MatanYadaev:
On completion:
image


services/workflows-service/src/events/workflow-completed-webhook-caller.ts
```diff
@@ -120,4 +126,36 @@ export class WorkflowCompletedWebhookCaller {
...
+    }
+
+    const statusMap = {
+      [StateTag.APPROVED]: 'Approved',
+      [StateTag.REJECTED]: 'Rejected',
+    } as const;
+
+    let status: (typeof statusMap)[keyof typeof statusMap] | undefined;
+
+    (workflowRuntimeData.tags ?? []).forEach((tag: string) => {
```
Omri-Levy:
```suggestion
    workflowRuntimeData?.tags?.forEach((tag: string) => {
```

services/workflows-service/src/events/workflow-completed-webhook-caller.ts
```diff
@@ -120,4 +126,36 @@ export class WorkflowCompletedWebhookCaller {
...
   }
+
+  private async updateSalesforceRecord(workflowRuntimeData: WorkflowRuntimeData) {
+    const { salesforceRecordId, salesforceObjectName } = workflowRuntimeData;
+
+    if (!salesforceRecordId || !salesforceObjectName) {
+      return;
+    }
+
+    const statusMap = {
```
Omri-Levy, line 137: Thoughts about using an array here instead? Could use `some` or `includes` and such. Most of our enums also export a type and an array.

Omri-Levy: Picking `some` or `includes` here over `forEach` on `tags` itself might also fit more.

services/workflows-service/src/workflow/workflow.controller.external.ts
```diff
@@ -318,6 +318,8 @@ export class WorkflowControllerExternal {
       currentProjectId,
     );
 
+    const hasSalesforceRecord = 'salesforceObjectName' in body && 'salesforceRecordId' in body;
```
Omri-Levy: What if their value is `undefined` or an empty string?

services/workflows-service/src/workflow/workflow.service.ts
```diff
@@ -1884,4 +1907,23 @@ export class WorkflowService {
...
   }
+
+  async updateSalesforceRecord({
+    workflowRuntimeData,
+    data,
+  }: {
+    workflowRuntimeData: WorkflowRuntimeData;
+    data: {
+      KYB_Started_At__c?: Date;
+      KYB_Status__c?: 'Not Started' | 'In Progress' | 'Approved' | 'Rejected';
```
Omri-Levy, line 1924: Should we extract to an enum?

services/workflows-service/src/workflow/workflow.controller.external.ts
MatanYadaev: Empty string is ok, but `undefined` is not. It's a client input, how can it be set to `undefined`?

services/workflows-service/src/workflow/workflow.service.ts
MatanYadaev: I don't think so...

services/workflows-service/src/workflow/workflow.controller.external.ts
Omri-Levy: Let me re-phrase. Is validation not needed there beyond the existence of property?

MatanYadaev: We didn't enable `exactOptionalPropertyTypes` so I changed the condition a bit. Thanks.

tomer-shvadron (approved):

PR #1086, opened 2023-09-13
Fix task decision validation error

### LINKED ISSUE #1208: Fix task decision validation error
### Description
Closes #1086. Validation fails once `document.decision.status` is set to `null` as part of the cancel re-submit flow.

Omri-Levy:
### Description
Closes #1208. Prior to this change trying to approve a task or ask for re-upload would fail due to invalid decision status.

### DIFF
--- a/services/workflows-service/src/workflow/workflow.service.ts
+++ b/services/workflows-service/src/workflow/workflow.service.ts
@@ -722,6 +722,10 @@ export class WorkflowService {
       documents: workflow?.context?.documents?.map(
         (document: DefaultContextSchema['documents'][number]) => ({
           ...document,
+          decision: {
+            ...document?.decision,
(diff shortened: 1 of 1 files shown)


alonp99 (approved):

services/workflows-service/src/workflow/workflow.service.ts
```diff
@@ -722,6 +722,10 @@ export class WorkflowService {
       documents: workflow?.context?.documents?.map(
         (document: DefaultContextSchema['documents'][number]) => ({
           ...document,
+          decision: {
+            ...document?.decision,
+            status: document?.decision?.status === null ? undefined : document?.decision?.status,
```
Omri-Levy, line 727: This is a copy-paste mistake. Should be:
```suggestion
status: document?.decision?.status === null ? undefined : decision?.status,
```

PR #1095, opened 2023-09-14
Fix that fields were editable when they shouldn't be

### LINKED ISSUE #1212: Bug: Fields are editable when user is not assigned

MatanYadaev:


apps/backoffice-v2/src/pages/Entity/components/EditableDetails/EditableDetails.tsx
```diff
@@ -284,6 +284,7 @@ export const EditableDetails: FunctionComponent<IEditableDetails> = ({
                         {isInput && !isSelect && (
                           <FormControl>
                             <Input
+                              {...field}
```
MatanYadaev, line 287: `field` had a `disabled` property that was set to `undefined`. This was reproduced only when the backoffice was installed as a standalone project (not part of a monorepo)

services/workflows-service/.env.example
```diff
@@ -16,7 +16,13 @@ API_KEY=secret
 NODE_ENV=local
 SENTRY_DSN=
 WEBHOOK_SECRET=webhook_secret
-IS_DEMO=
+WEBHOOK_URL=
```
tomer-shvadron, line 19: Is this necessary to be in the code-base?

tomer-shvadron (approved):

apps/backoffice-v2/src/pages/Entity/components/EditableDetails/EditableDetails.tsx
Omri-Levy: Great work! 👑

Omri-Levy (approved):

services/workflows-service/.env.example
MatanYadaev: Currently, yes

PR #1103, opened 2023-09-18
blokh/feat/ (fix-badges & intial states in DB)

### LINKED ISSUE #1222: (fix-badges & intial states in DB)

### LINKED ISSUE #1252: fix-badges & intial states of defintion when starting instance

Blokh:
- [x] added tags on missing workflows
- [x] updates workflow service to set state and initial state in DB
- [x] add tags logic on created workflow


tomer-shvadron (approved):

apps/backoffice-v2/src/pages/Entity/components/Case/hooks/useCaseActionsLogic/useCaseActionsLogic.tsx
```diff
@@ -70,7 +70,7 @@ export const useCaseActionsLogic = ({ workflowId, fullName }: IUseActions) => {
   );
 
   const tag = useMemo(() => {
-    if (!workflow?.context?.entity?.data?.__isFinished) {
+    if (workflow?.context?.entity?.data?.__isFinished === false) {
```
Blokh: this is being removed

services/workflows-service/scripts/generate-end-user.ts
```diff
@@ -132,6 +132,7 @@ export const generateBusiness = ({
         workflowDefinitionId,
         createdAt: faker.date.recent(2),
         projectId: projectId,
+        tags: ['manual_review'],
```
Omri-Levy: Should get the tag configured to the initial state, if not we should still use the tags enum here.

services/workflows-service/scripts/workflows/kyc-email-process-example.ts
```diff
@@ -13,6 +13,7 @@ export const kycEmailSessionDefinition = {
     initial: 'idle',
     states: {
       idle: {
+        tags: [StateTag.PENDING_PROCESS],
```
Omri-Levy, line 16: `PENDING_PROCESS` signifies an in-progress process with a vendor. Is that the case for `idle`?

services/workflows-service/scripts/workflows/parent-kyb-kyc-session-workflow.ts
```diff
@@ -135,19 +136,22 @@ export const parentKybWithSessionWorkflowDefinition = {
...
       run_kyb_enrichment: {
+        tags: [StateTag.COLLECTION_FLOW],
         on: {
           KYB_DONE: [{ target: 'pending_kyc_response_to_finish' }],
           // TODO: add 404 handling
           FAILED: [{ target: 'auto_reject' }],
         },
       },
       pending_kyc_response_to_finish: {
+        tags: [StateTag.COLLECTION_FLOW],
```
Omri-Levy: Doesn't this step waits for a vedor process to be done?

services/workflows-service/src/workflow/workflow.service.ts
```diff
@@ -1386,6 +1385,8 @@ export class WorkflowService {
             } as InputJsonValue,
             config: merge(workflowDefinition.config, validatedConfig || {}) as InputJsonValue,
             status: 'active',
+            state: workflowDefinition.definition.initial,
```
Omri-Levy, line 1388: If makes sense, running the machine and getting the current state will be more reliable. Would give access to the current tag, too.

Omri-Levy: Let's make sure we're not missing other instances of `workflowRuntimeDataRepository.create`.

services/workflows-service/scripts/workflows/kyc-email-process-example.ts
Blokh: in this context, the StateTag.PENDING_PROCESS of end user over KYC session ( this state is pending for getting kyc session )

services/workflows-service/src/workflow/workflow.service.ts
Blokh: true, but the change is quite scary - I wanted to fix it here without changing the logic.
so, if we want to run the machine when creating it. we need to re-validate that all the schemas are valid from the get-go

services/workflows-service/scripts/workflows/kyc-email-process-example.ts
Omri-Levy: But is it pending a response from a 3rd party?

services/workflows-service/src/workflow/workflow.service.ts
Omri-Levy: Alright, let's just make sure it's documented somewhere. It could be a good candidate for Prisma middleware too. 🙂

services/workflows-service/scripts/generate-end-user.ts
```diff
@@ -188,6 +190,7 @@ export const generateEndUser = ({
           createdAt: faker.date.recent(2),
           projectId: projectId,
           parentRuntimeDataId: workflow.parentRuntimeId,
+          tags: StateTag.MANUAL_REVIEW,
```
Omri-Levy: Missing square brackets.

alonp99 (approved):

PR #1104, opened 2023-09-18
fix: fixed issue where document properties could not be erased

### LINKED ISSUE #1106: Backoffice - Demo to Dev issues
Issues

- [x] - Fields cannot be updated to empty (Partially done, date fields still could not be saved as empty, [PR)](https://github.com/ballerine-io/ballerine/pull/1104)
- [x] - Avatar icon is not centered right [(PR)](https://github.com/ballerine-io/ballerine/pull/1105)
- [x] - Some avatars are cut [(PR)](https://github.com/ballerine-io/ballerine/pull/1107)
- [x] - happens only in individuals test view Assignee uses the same avatar as the case [(PR)](https://github.com/ballerine-io/ballerine/pull/1108)
- [x] - happens only in individuals test view Assignment dropdown all items have the same avatar [(PR)](https://github.com/ballerine-io/ballerine/pull/1108)
- [x] - happens only in individuals test view Avatar initials are not in use [(PR)](https://github.com/ballerine-io/ballerine/pull/1108)
- [ ] - Old cases are affected by the document image being downloaded [(case)](https://backoffice-sb-test.eu.ballerine.app/en/case-management/entities/cljzy13xs0044pil8hscpypyd?sortDir=desc&pageSize=50&page=2&search=&filterId=cljzy13ne0007pil80nxzfdlx&entity=&sortBy=createdAt&filter[assigneeId][0]=cljzy13lu0003pil87r52fwtu&filter[assigneeId][1]=__null&filter[status][0]=active) (I could not reproduce it locally,it seems that this is related to host configuration where app is deployed)

[Notion](https://www.notion.so/ballerine/Demo-to-Dev-0d2cfaea91ca48189309f690c9990866)

chesterkmr:
### Description
- Implemented endpoint for document properties update.(Not in use currently)
- updated method `updatedWorkflowRuntimeData` to override document properties so they could be removed by operator from `Backoffice`
- merge with `demo`

### Related issues

### Breaking changes

### Examples and references

### Checklist
- [] I have read the [contribution guidelines](CONTRIBUTING.md) of this project
- [] I have read the [style guidelines](STYLE_GUIDE.md) of this project
- [] I have performed a self-review of my own code
- [] I have commented my code, particularly in hard-to-understand areas
- [] I have made corresponding changes to the documentation
- [] My changes generate no new warnings and errors
- [] New and existing tests pass locally with my changes


apps/backoffice-v2/src/pages/Entity/components/EditableDetails/helpers.ts
```diff
@@ -0,0 +1,23 @@
+import { AnyObject } from '@ballerine/ui';
+
+export const isNullishDateValue = <T = unknown>(value: T, jsonSchema: AnyObject): boolean => {
+  if (!value && jsonSchema?.format === 'date') return true;
```
Omri-Levy: We have a utility called `isNullish`. Nullish should only cover `undefined` and `null`, otherwise rename the function to something like `isFalsy`. 🙂

apps/backoffice-v2/src/pages/Entity/components/EditableDetails/EditableDetails.tsx
```diff
@@ -90,8 +91,10 @@ export const EditableDetails: FunctionComponent<IEditableDetails> = ({
 
         const properties = Object.keys(document?.propertiesSchema?.properties ?? {}).reduce(
           (acc, curr) => {
-            if (!formData?.[curr]) return acc;
-            acc[curr] = formData?.[curr];
+            const schema = (document?.propertiesSchema?.properties[curr] || {}) as AnyObject;
+
+            if (!isValueEligibleToBePersisted(formData?.[curr], schema)) return acc;
```
Omri-Levy: Can just use the `isNullish` util here and return if true.

apps/backoffice-v2/src/pages/Entity/components/EditableDetails/EditableDetails.tsx
```diff
@@ -90,8 +91,11 @@ export const EditableDetails: FunctionComponent<IEditableDetails> = ({
 
         const properties = Object.keys(document?.propertiesSchema?.properties ?? {}).reduce(
```
liorzam, line 92: Alternative for the same result is about to run over entries instead of manipulating the object's keys 
```suggestion
const properties = Object.entries(document?.propertiesSchema?.properties ?? {}).reduce(
  (acc, [curr, schema]) => {
    const formDataValue = formData?.[curr];

    // Check if form data value is nullish; if so, skip this property
    if (isNullish(formDataValue)) return acc;

    // Serialize the form data value using the schema
    acc[curr] = serializeValue(formDataValue, schema);

    return acc;
  },
  {}
);
```

liorzam (approved):

services/workflows-service/src/workflow/workflow.service.ts
```diff
@@ -1924,6 +1940,26 @@ export class WorkflowService {
...
 
+  async updateDocumentProperties(
+    workflowId: string,
+    targetDocumentId: string,
+    properties: object,
+    projectIds: TProjectIds,
+  ) {
+    const workflow = await this.workflowRuntimeDataRepository.findById(workflowId, {}, projectIds);
+
+    workflow.context.documents?.map((document: any) => {
```
alonp99, line 1951: 1. You don't use the map value. and pass the existing context from the database to be merged - so it wont change anything
2. If you're updating a single document you don't have to keep the other document, just the document you want to update
3. Error handling: Youre depending on array update by id, make sure there is an id
4. be explicit when calling the update method and pass 'by-id' param

services/workflows-service/src/workflow/workflow.service.ts
```diff
@@ -1924,6 +1940,26 @@ export class WorkflowService {
     return documentsWithPersistedImages;
   }
 
+  async updateDocumentProperties(
+    workflowId: string,
+    targetDocumentId: string,
+    properties: object,
+    projectIds: TProjectIds,
+  ) {
```
alonp99, line 1948: Add tests: This code won't work as is. Tests would've caught the issues.

services/workflows-service/src/workflow/workflow.service.ts
```diff
@@ -897,6 +897,22 @@ export class WorkflowService {
       contextHasChanged = !isEqual(data.context, runtimeData.context);
       mergedContext = merge({}, runtimeData.context, data.context);
 
+      const recievedDocumentsMap = keyBy(data.context?.documents ?? [], 'id');
+
+      // This is hack to re-assign properties to documents after merge as temp solution
```
alonp99, line 902: We want to tackle tech debt, and we can't add more right now. let's go with the proper solution.

MatanYadaev: @alonp99 I agree with your statement, but I don't think it's correct in this particular case, as it doesn't add more tech debt, it just keeps the existing tech debt. The amount of tech debt remains the same before and after this PR.

services/workflows-service/src/workflow/workflow.controller.internal.ts
```diff
@@ -197,6 +199,21 @@ export class WorkflowControllerInternal {
     }
   }
 
+  @common.Put(':id/properties/:documentId')
```
Omri-Levy: An endpoint to update a document by id should suffice. We also update a document's type and category for example.

Omri-Levy:
```suggestion
  @common.Put(':id/documents/:documentId')
```

services/workflows-service/src/workflow/workflow.service.ts
```diff
@@ -897,6 +897,22 @@ export class WorkflowService {
...
+      // TO DO: Remove it and update backoffice to work with new endpoint Refactor and use :id/properties/:documentId endpoint instead
+      mergedContext.documents =
+        mergedContext.documents.map((document: any) => {
+          const updatedDocumentProperties = recievedDocumentsMap[document.id];
+
+          return {
+            ...document,
+            properties:
+              (updatedDocumentProperties && updatedDocumentProperties.properties) ??
+              document.properties,
```
Omri-Levy, line 912:
```suggestion
            properties: updatedDocumentProperties?.properties ?? document.properties,
```

services/workflows-service/src/workflow/workflow.service.ts
```diff
@@ -897,6 +897,22 @@ export class WorkflowService {
       contextHasChanged = !isEqual(data.context, runtimeData.context);
       mergedContext = merge({}, runtimeData.context, data.context);
 
+      const recievedDocumentsMap = keyBy(data.context?.documents ?? [], 'id');
+
+      // This is hack to re-assign properties to documents after merge as temp solution
+      // TO DO: Remove it and update backoffice to work with new endpoint Refactor and use :id/properties/:documentId endpoint instead
+      mergedContext.documents =
+        mergedContext.documents.map((document: any) => {
```
Omri-Levy, line 905:
```suggestion
        mergedContext.documents.map((document: DefaultContextSchema['documents'][number]) => {
```

services/workflows-service/src/workflow/workflow.service.ts
Omri-Levy, line 1951:
```suggestion
    workflow.context.documents?.map((document: DefaultContextSchema['documents'][number]) => {
```

PR #1174, opened 2023-09-27
Merge dev into demo

MatanYadaev:


apps/backoffice-v2/src/pages/Entity/components/EditableDetails/EditableDetails.tsx
```diff
@@ -91,6 +91,13 @@ export const EditableDetails: FunctionComponent<IEditableDetails> = ({
         const properties = Object.keys(document?.propertiesSchema?.properties ?? {}).reduce(
           (acc, curr) => {
             if (!formData?.[curr]) return acc;
+
+            if (document?.propertiesSchema?.properties?.[curr]?.format === 'date-time') {
```
alonp99: Can you leave a comment explaining this fix?

apps/backoffice-v2/src/pages/Entity/components/EditableDetails/EditableDetails.tsx
```diff
@@ -91,6 +91,13 @@ export const EditableDetails: FunctionComponent<IEditableDetails> = ({
...
+              if (formData?.[curr]?.length === 16) {
+                formData[curr] = `${formData[curr]}:00`;
+              }
+            }
+
```
Omri-Levy:
```suggestion
            if (document?.propertiesSchema?.properties?.[curr]?.format === 'date-time' && formData?.[curr]?.length === 16) {
                formData[curr] = `${formData[curr]}:00`;
            }
```

Omri-Levy: `formData` here is React state. Mutating it is a bad practice and may behave in unexpected ways.

apps/backoffice-v2/src/pages/Entity/components/EditableDetails/EditableDetails.tsx
MatanYadaev: Yes, sorry: HTML5 `datetime-local` input holds a date time value. It should return `YYYY-MM-DDTHH:MM:SS`, but there is an edge-case when the seconds are `00`, it returns `YYYY-MM-DDTHH:MM` instead.
So here I fixed it, ensuring that when there are no seconds, I add `:00` to the string: `YYYY-MM-DDTHH:MM:00`

Omri-Levy (approved):

PR #1343, opened 2023-10-03
Dynamic UI Feature branch PR

Blokh:
dynamic ui schema


services/workflows-service/prisma/migrations/20230921074435_enable_is_legacy_reject_flag_for_existing_workflow_definitions/migration.sql
```diff
@@ -0,0 +1 @@
+UPDATE "WorkflowDefinition" SET "config" = jsonb_set(COALESCE("config", '{}'::jsonb), '{isLegacyReject}', 'true'::jsonb);
```
alonp99, line 1: I dont think it should be in the migrations, we can't be sure in the future it will always run on data expect it to run on - it should be a one time thing done on a specific environment with a standalone script like the migration script

services/workflows-service/src/ui-definition/utils/collection-flow/8-website-basic-requirement.ts
```diff
@@ -0,0 +1,316 @@
+const validationSchema = {
```
alonp99, line 1: Verify that we have a task in the relevant project to move this into the db

services/workflows-service/src/auth/session-serializer.ts
```diff
@@ -17,7 +17,6 @@ export class SessionSerializer extends PassportSerializer {
         firstName: user.firstName,
         lastName: user.lastName,
       },
-      projectIds: user.userToProjects?.map(userToProject => userToProject.projectId) || null,
```
alonp99, line 20: this is the fix

services/workflows-service/src/auth/session-serializer.ts
```diff
@@ -43,7 +42,7 @@ export class SessionSerializer extends PassportSerializer {
         user: userData,
         projectIds: userToProjects?.map(userToProject => userToProject.projectId) || null,
```
liorzam, line 43: should be delete as well

liorzam (changes requested):

PR #1378, opened 2023-10-08
Move all capabilities of /with-token endpoint into /run endpoint

MatanYadaev:


packages/common/src/schemas/documents/default-context-schema.ts
```diff
@@ -1,30 +1,22 @@
...
-        },
-        data: {
-          type: 'object',
-          properties: {
-            additionalInfo: {
-              type: 'object',
-            },
+import { Type } from '@sinclair/typebox';
+
+export const defaultContextSchema = Type.Object({
```
MatanYadaev, line 3: I refactored the JSON Schema into a TypeBox object so I could use `Type.Composite` in a different file. The output schema should be almost identical, but please review it.
Why it's not exactly identical? Because in some places the JSON Schema wasn't valid (`integer` instead of `number`, etc)

services/workflows-service/scripts/workflows/kyb-kyc-workflow-definition.ts
```diff
@@ -0,0 +1,280 @@
+import { PrismaClient } from '@prisma/client';
+import { kycEmailSessionDefinition } from './kyc-email-process-example';
+import { env } from '../../src/env';
+
+import { defaultContextSchema, StateTag } from '@ballerine/common';
+import { Type } from '@sinclair/typebox';
+
+export const kybKycWorkflowDefinition = {
```
MatanYadaev, line 8: Please compare this with `parentKybWithSessionWorkflowDefinition` in order to see the differences.

services/workflows-service/src/collection-flow/controllers/collection-flow.controller.ts
```diff
@@ -27,8 +27,8 @@ export class ColectionFlowController {
   }
 
   @common.Get('/user')
-  async getUser(@TokenScope() tokenScope: ITokenScope) {
-    return this.service.getUser(tokenScope.endUserId, tokenScope.projectId);
+  async getUser() {
+    return {}; // @TODO: Make sure the client is not using this endpoint and remove it.
   }
```
MatanYadaev: Need to check in which places the `kyb-app` is using this endpoint and make sure it's no longer in usage. Illia says it's not necessary anymore in the `kyb-app`: https://teamballerine.slack.com/archives/C0510CD1R8T/p1696762753574889

services/workflows-service/src/storage/storage.service.ts
```diff
@@ -11,16 +12,19 @@ export class StorageService {
   async createFileLink({
     uri,
     fileNameOnDisk,
-    userId,
+    userId = '', // @TODO: Check the circumstances
```
MatanYadaev: There's no longer `userId` when using this function because token doesn't have an end-user anymore. Need to check if it's ok to not have a `userId` in this case.

services/workflows-service/src/workflow/workflow.controller.external.ts
```diff
@@ -172,12 +171,19 @@ export class WorkflowControllerExternal {
...
     const actionResult = await this.service.createOrUpdateWorkflowRuntime({
       workflowDefinitionId: workflowId,
       context,
       config,
       projectIds,
       currentProjectId,
+      ...(hasSalesforceRecord && {
+        salesforceObjectName: body.salesforceObjectName,
+        salesforceRecordId: body.salesforceRecordId,
+      }),
```
MatanYadaev, line 185: Need to QA that these properties are stored in the DB.

services/workflows-service/src/workflow/workflow.service.ts
```diff
@@ -1440,8 +1443,6 @@ export class WorkflowService {
             } as InputJsonValue,
             config: merge(workflowDefinition.config, validatedConfig || {}) as InputJsonValue,
             status: 'active',
-            state: workflowDefinition.definition.initial,
-            tags: workflowDefinition.definition.states[workflowDefinition.definition.initial]?.tags,
```
MatanYadaev, line 1441: These properties get filled in the `this.event(...)` in line 1481

services/workflows-service/scripts/workflows/kyb-kyc-workflow-definition.ts
```diff
@@ -0,0 +1,280 @@
...
+        type: 'final' as const,
+      },
+    },
+  },
+  extensions: {
+    apiPlugins: [
+      {
+        name: 'invitation_email',
+        pluginKind: 'email',
+        url: `{secret.EMAIL_API_URL}`,
```
liorzam, line 117: I couldnt see this env var in hashicorp

services/workflows-service/scripts/workflows/kyb-kyc-workflow-definition.ts
```diff
@@ -0,0 +1,280 @@
...
+            {
+              transformer: 'jmespath',
+              mapping: `{
+              templateId: 'd-00a0d5d14cb14fbb9034b53c6ef7e5fa',
+              adapter: '${env.MAIL_ADAPTER}'
+              from: 'no-reply@ballerine.com',
+              receivers: [mainRepresentative.email],
+              name: mainRepresentative.fullName,
+              provider: customerName,
+              url: join('',['${env.KYB_EXAMPLE_CORS_ORIGIN}?token=',token])
```
liorzam, line 137: this origin is a secret 
image

services/workflows-service/scripts/workflows/kyb-kyc-workflow-definition.ts
```diff
@@ -0,0 +1,280 @@
...
+        },
+        request: {
+          transform: [
+            {
+              transformer: 'jmespath',
+              mapping: `{
+              kybCompanyName: entity.data.companyName,
+              customerCompanyName: entity.data.additionalInfo.ubos[0].entity.data.additionalInfo.customerCompany,
+              firstName: entity.data.additionalInfo.mainRepresentative.firstName,
+              resubmissionLink: join('',['https://',entity.data.additionalInfo.ubos[0].entity.data.additionalInfo.normalizedCustomerCompany,'.demo.ballerine.app','/workflowRuntimeId=',workflowRuntimeId,'?resubmitEvent=RESUBMITTED']),
```
liorzam, line 195: is it hard coded only for demo env? `'.demo.ballerine.app'`

services/workflows-service/scripts/workflows/kyb-kyc-workflow-definition.ts
MatanYadaev: I think it's not a new variable that I added. Can you check in the other definition? I think I just copy-pasted this ine

services/workflows-service/scripts/workflows/kyb-kyc-workflow-definition.ts
MatanYadaev: Ok, so?

services/workflows-service/scripts/workflows/kyb-kyc-workflow-definition.ts
MatanYadaev: It shouldn't be hard-coded

PR #1394, opened 2023-10-11
Changed blocks based on signoff comments

### LINKED ISSUE #1372: Display new collection flow data in backoffice

Omri-Levy:
Mostly improves spacing of UI in the backoffice.


apps/backoffice-v2/src/pages/Entity/hooks/useEntity/utils.ts
```diff
@@ -77,10 +77,18 @@ export const extractCountryCodeFromWorkflow = (workflow: TWorkflowById) => {
   })?.issuer?.country;
 };
 
-export const omitPropsFromObject = (obj, ...props) => {
+export const omitPropsFromObject = <
```
Omri-Levy, line 80: `omitPropsFromObject` now autocompletes to keys available on the passed object, and the type of the output object is now the object without the omitted props.

alonp99 (approved):

PR #1407, opened 2023-10-12
Replaced non-dropdown fields with a dropdown

Omri-Levy:
Related to BAL-402 .


services/workflows-service/src/ui-definition/utils/collection-flow/languages.ts
```diff
@@ -0,0 +1,189 @@
+export const languages = [
```
alonp99: Why its in the workflow service?

Blokh (approved):

PR #1425, opened 2023-10-16
Remove from entity block props represented in blocks of their own

Omri-Levy:
- refactor(backoffice-v2): removed props from the entity block which are already blocks themselves
- refactor(backoffice-v2): now removing the ubos property from entity block if rendering kyc blocks

### DIFF
--- a/apps/backoffice-v2/src/pages/Entity/hooks/useTasks/useTasks.tsx
+++ b/apps/backoffice-v2/src/pages/Entity/hooks/useTasks/useTasks.tsx
@@ -382,11 +382,7 @@ export const useTasks = ({
                     ...Object.entries(
                       omitPropsFromObject(entity?.data, 'additionalInfo', 'address') ?? {},
                     ),
-                    ...Object.entries(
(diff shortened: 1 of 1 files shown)


apps/backoffice-v2/src/pages/Entity/hooks/useTasks/useTasks.tsx
```diff
@@ -381,7 +382,11 @@ export const useTasks = ({
                     ...Object.entries(
                       omitPropsFromObject(entity?.data, 'additionalInfo', 'address') ?? {},
                     ),
-                    ...Object.entries(entity?.data?.additionalInfo ?? {}),
+                    ...Object.entries(
+                      workflow?.childWorkflows?.length
+                        ? omitPropsFromObject(entityDataAdditionalInfo ?? {}, 'ubos')
```
Omri-Levy: Using `entityDataAdditionalInfo` instead of `entity.data.additionalInfo` since it does not include `mainContact` and so on.

tomer-shvadron (approved):

PR #1433, opened 2023-10-17
S3 downloads pdfs instead of showing fix

Omri-Levy:
### Description
No longer using `application/octet-stream` `Content-Type` header for S3 files.


services/workflows-service/src/providers/file/file-provider/aws-s3-file.service.ts
```diff
@@ -152,7 +161,8 @@ export class AwsS3FileService implements IStreamableFileProvider {
       ...getObjectCommandInput,
       Body: readableStream,
       Key: fileName,
-    };
+      ContentType: mimeType,
```
Omri-Levy, line 164: The main change.

PR #1454, opened 2023-10-19
Fixed KYC block having editable state when unassigned

Omri-Levy:
### Description
Temporarily sets `isEditable` to `false` on KYC blocks until editing is fully implemented.


apps/backoffice-v2/src/pages/Entity/components/KycBlock/hooks/useKycBlock/useKycBlock.tsx
```diff
@@ -122,7 +122,7 @@ export const useKycBlock = ({
       title,
       value,
       pattern: '',
-      isEditable: true,
```
Blokh, line 125: wouldn't it fuck it up for fido?

Blokh (approved):

PR #1471, opened 2023-10-23
fix(workflows): fix kyb workflow, duplicate business creation on run

alonp99:


packages/workflow-core/src/lib/workflow-runner.ts
```diff
@@ -420,14 +420,21 @@ export class WorkflowRunner {
...
 
           if (state.done) {
             console.log('Reached final state');
           }
 
+          if (state.tags.has('failure')) {
+            console.log('Reached failure state', {
+              correlationId: context?.entity?.id,
+              ballerineEntityId: context?.entity?.ballerineEntityId,
+            });
```
liorzam, line 435: use logger instead console.log

packages/workflow-core/src/lib/plugins/external-plugin/api-plugin.ts
```diff
@@ -23,6 +23,7 @@ export class ApiPlugin {
     this.method = pluginParams.method;
     this.headers = {
       'Content-Type': 'application/json',
+      accept: 'application/json',
```
liorzam, line 26: can remove

packages/workflow-core/src/lib/workflow-runner.ts
```diff
@@ -420,14 +420,21 @@ export class WorkflowRunner {
...
         if (state.changed) {
           console.log('Transitioned into', state.value);
 
           if (state.done) {
             console.log('Reached final state');
           }
 
+          if (state.tags.has('failure')) {
+            console.log('Reached failure state', {
+              correlationId: context?.entity?.id,
```
liorzam, line 433: entity id has to be present why do you access with question mark?

liorzam (approved):

packages/workflow-core/src/lib/workflow-runner.ts
alonp99: no logger yet in the wf-core

packages/workflow-core/src/lib/plugins/external-plugin/api-plugin.ts
alonp99: It might be beneficial for other API, no downside here

PR #1482, opened 2023-10-24
added logic of child to parent event persistence

Blokh:
added the logic of persistence of parent to child workflow

### DIFF
--- a/services/workflows-service/scripts/workflows/dynamic-ui-workflow.ts
+++ b/services/workflows-service/scripts/workflows/dynamic-ui-workflow.ts
@@ -72,6 +72,7 @@ export const dynamicUiWorkflowDefinition = {
           approve: 'approved',
           reject: 'rejected',
           revision: 'pending_resubmission',
+          KYC_REVISION: 'revision',
         },
       },
       pending_resubmission: {
@@ -103,6 +104,17 @@ export const dynamicUiWorkflowDefinition = {
               },
             },
             { target: 'pending_kyc_response_to_finish' },
+          ],
+          KYC_RESPONDED: [
+            {
+              target: 'manual_review',
+              cond: {
+                type: 'jmespath',
+                options: {
+                  rule: 'length(childWorkflows.kyc_email_session_example.*.[result.vendorResult.decision][]) == length(childWorkflows.kyc_email_session_example.*[])',
+                },
+              },
+            },
           ],
         },
       },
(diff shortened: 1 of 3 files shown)


services/workflows-service/scripts/workflows/dynamic-ui-workflow.ts
```diff
@@ -263,6 +275,11 @@ export const dynamicUiWorkflowDefinition = {
         persistenceStates: ['kyc_manual_review'],
         deliverEvent: 'KYC_RESPONDED',
       },
+      {
+        definitionId: kycEmailSessionDefinition.name,
+        persistenceStates: ['revision_email_sent'],
```
alonp99, line 280: It probably should be renamed, since not necessarily something is being persisted - maybe "callbackStates"

services/workflows-service/scripts/workflows/dynamic-ui-workflow.ts
```diff
@@ -72,6 +72,7 @@ export const dynamicUiWorkflowDefinition = {
           approve: 'approved',
           reject: 'rejected',
           revision: 'pending_resubmission',
+          KYC_REVISION: 'revision',
```
alonp99, line 75: it should go to 'pending_kyc_response_to_finish'

services/workflows-service/src/workflow/workflow.service.ts
```diff
@@ -1017,7 +1017,9 @@ export class WorkflowService {
         {
           data: {
             status: 'active',
-            state: parentMachine?.workflowDefinition?.definition?.initial as string,
+            state:
```
alonp99, line 1020: Have questions about this part

services/workflows-service/scripts/workflows/dynamic-ui-workflow.ts
```diff
@@ -104,6 +105,17 @@ export const dynamicUiWorkflowDefinition = {
             },
             { target: 'pending_kyc_response_to_finish' },
           ],
+          KYC_RESPONDED: [
```
alonp99, line 108: This one should be reverted i think, since per Nitzan, while there a document in revision we should stay in revision regardless of what happens with the KYC

services/workflows-service/scripts/workflows/dynamic-ui-workflow.ts
```diff
@@ -104,6 +105,17 @@ export const dynamicUiWorkflowDefinition = {
             },
             { target: 'pending_kyc_response_to_finish' },
```
alonp99, line 106: what happens here where its the second kyc is submitted, will the rule catch it?

PR #1511, opened 2023-10-29
blokh/feat/finsh-mt-migration 

Blokh:
feat: removed scoping of create or update as the prisma types validates it already.
p.s. the eslint rules for create or update are not longer needed as Prisma times are validating the inputs


services/workflows-service/plugins/verify-repository-project-scoped/verify-repository-project-scoped.rule.js
```diff
@@ -17,7 +17,10 @@ module.exports = {
     return {
       MethodDefinition: function (node) {
         if (!isRepository || node.key.name === 'constructor') return;
-        const isUnscoped = node.key.name.toLowerCase().includes('unscoped');
+        const isUnscoped =
```
Omri-Levy: Can this be an array of strings instead?

Omri-Levy (approved):

alonp99 (approved):

services/workflows-service/plugins/verify-repository-project-scoped/verify-repository-project-scoped.rule.js
alonp99: @Blokh

PR #1519, opened 2023-10-30
feat: specify envrionment name for sentry

liorzam:
### Description
This Pull Request introduces a new feature: the addition of the Sentry environment name.
The Sentry environment name is now dynamically set in the application, improving error tracking and debugging.
This feature aims to enhance the development and monitoring experience, making it easier to distinguish errors in different environments (e.g., development, staging, production).


services/workflows-service/src/env.ts
```diff
@@ -10,7 +10,8 @@ export const env = createEnv({
    */
   clientPrefix: 'PUBLIC_',
   server: {
-    NODE_ENV: z.enum(['development', 'production', 'test', 'local']),
+    NODE_ENV: z.enum(['development', 'production', 'test', 'local']), // TODO: remove 'test', 'local'
```
Omri-Levy, line 13: This validation should include valid `NODE_ENV` values, here 'test' is valid.

Omri-Levy (approved):

PR #1523, opened 2023-10-30
Blokh/feat/child workflow revision

Blokh:
generate logic of revision from children to parent and persistence to parent + updating rules


services/workflows-service/src/workflow/workflow.service.ts
```diff
@@ -1028,7 +1029,9 @@ export class WorkflowService {
         {
           data: {
             status: 'active',
-            state: parentMachine?.workflowDefinition?.definition?.initial as string,
+            state:
```
alonp99: Remove state change entierly

services/workflows-service/src/workflow/workflow.service.ts
```diff
@@ -1965,11 +1969,7 @@ export class WorkflowService {
     workflowDefinition: WorkflowDefinition,
     isPersistableToParent: boolean,
   ) {
-    if (!isPersistableToParent)
-      return this.composeContextWithChildResponse(
-        parentWorkflowRuntime.context,
-        workflowDefinition.id,
-      );
+    if (!isPersistableToParent) return parentWorkflowRuntime.context;
```
alonp99: Seems like this check should be before calling this function and not inside it

alonp99 (approved):

PR #1532, opened 2023-10-31
bal 683 collection flow upload file set new files in a customer

liorzam:


services/workflows-service/src/collection-flow/collection-flow.service.ts
```diff
@@ -324,4 +327,39 @@ export class CollectionFlowService {
...
+    );
+
+    const entityId = runtimeDataId.businessId || runtimeDataId.endUserId;
+
+    if (!entityId) {
+      throw new NotFoundException("Workflow does't exists");
+    }
+
+    // Remove file extension (get everything before the last dot)
+    const nameWithoutExtension = (file.originalname || generateRandomId(16)).replace(
```
Blokh: use uuid instead of random. simpler without replacing anything

services/workflows-service/src/collection-flow/collection-flow.service.ts
```diff
@@ -324,4 +327,39 @@ export class CollectionFlowService {
...
+      /\.[^.]+$/,
+      '',
+    );
+    // Remove non characters
+    const alphabeticOnlyName = nameWithoutExtension.replace(/\W/g, '');
+
+    const persistedFile = await this.fileService.copyToDestinationAndCreate(
+      { id: alphabeticOnlyName, uri: file.path, provider: 'file' },
```
Blokh: provider is the file provider, in this case of the file is local file system. it should be 'file-system'

services/workflows-service/src/common/interceptors/remove-temp-file.interceptor.ts
```diff
@@ -0,0 +1,53 @@
+import { AppLoggerService } from '@/common/app-logger/app-logger.service';
+import { Injectable, NestInterceptor, ExecutionContext, CallHandler } from '@nestjs/common';
+import { Observable, catchError, firstValueFrom, of, tap } from 'rxjs';
+import * as fs from 'fs';
+
+@Injectable()
+export class RemoveTempFileInterceptor implements NestInterceptor {
+  constructor(private readonly logger: AppLoggerService) {}
+  private async deleteTempFile(req: any) {
+    if (req?.file?.path) {
```
Blokh, line 10: the file in the request is no the file on your local system.
also, this is an overcomplicated n and should be just in the process of if you downloaded to your local env afield via the import-export process.  ensure the downloaded file is deleted.
and it should be a part of the process and not a part of the server's functionality/middleware.

services/workflows-service/src/common/utils/random/index.ts
```diff
@@ -0,0 +1,35 @@
...
+  includeAlphabetics?: boolean;
+};
+
+const defaultOptions: Options = {
+  includeSymbols: false,
+  includeNumbers: true,
+  includeAlphabetics: true,
+};
+
+export const generateRandomId = (length: number, options?: Options) => {
```
Blokh: why do you need this?
you have uuid

services/workflows-service/src/common/utils/random/index.ts
```diff
@@ -0,0 +1,35 @@
+const SYMBOLS = '!@#$%^&*()_+-=[]{}|;:,.<>?';
```
Blokh: what is the usage of this? what does it have to do with the PR?
overall why?

services/workflows-service/src/collection-flow/collection-flow.service.ts
```diff
@@ -324,4 +327,39 @@ export class CollectionFlowService {
...
+      throw new NotFoundException("Workflow does't exists");
+    }
+
+    // Remove file extension (get everything before the last dot)
+    const nameWithoutExtension = (file.originalname || generateRandomId(16)).replace(
+      /\.[^.]+$/,
+      '',
+    );
+    // Remove non characters
+    const alphabeticOnlyName = nameWithoutExtension.replace(/\W/g, '');
```
Blokh, line 350: same, this code is unnecessary

services/workflows-service/src/collection-flow/collection-flow.service.ts
liorzam: i prefer this method in order to control what symbols the generated string will include and also the length.
i can replace it with uuid if there concrete problem

services/workflows-service/src/collection-flow/collection-flow.service.ts
liorzam: will change, btw there is no such an implementation for file, this one goes to the default anyway which is local file system

services/workflows-service/src/collection-flow/collection-flow.service.ts
liorzam: let me explain what it does, i will keep the original file name concatenated with random string and also keeping the extension file name

liorzam: Remove all non characters like symbol spaces from the original name:
image

image

services/workflows-service/src/common/utils/random/index.ts
liorzam: its not used here, just added a utils function that has the option to generate a random string with symbols.

services/workflows-service/src/common/utils/random/index.ts
liorzam: just shorter id and supported cross platform compared to crypto library

services/workflows-service/src/common/interceptors/remove-temp-file.interceptor.ts
liorzam: but i dont download any file, once ive got the file in the controller the file already uploaded to the filesystem

liorzam: I agree with you when in case our multer was `memoryStorage();`
since we are using file storage, i have to remove the file because multer already keep it on file system, that's why you can see that im skip the downloading part .

these days you dont *actually remove* them when running locally with file provider

PR #1536, opened 2023-11-01
Adding AsiaVerify integrations to workflow

tomer-shvadron:
### Description
Adding AsiaVerify integrations to new workflow


services/workflows-service/scripts/workflows/dynamic-ui-workflow.ts
```diff
@@ -32,13 +32,57 @@ export const dynamicUiWorkflowDefinition = {
...
+              },
+            },
+            {
+              target: 'failed',
+            },
+          ],
+        },
+      },
+      run_cn_kyb: {
+        tags: [StateTag.COLLECTION_FLOW],
```
alonp99: it should be something else, not COLLECTION_FLOW - maybe ask Nitzan what we should show on that stages

alonp99 (approved):

PR #1541, opened 2023-11-01
fix(wf-service): webhooks tests fix

alonp99:


services/workflows-service/src/workflow/workflow.service.unit.test.ts
```diff
@@ -89,12 +89,17 @@ function buildWorkflowDeifintion(sequenceNum) {
   };
 }
 
-function buildDocument(category, status) {
+function buildDocument(category, status, fileType = 'jpg') {
```
liorzam, line 92: why dont you assume the type is jpg?

PR #1548, opened 2023-11-02
Added a layout for when server is down

Omri-Levy:
### Description
This change also ensures no fetch infinite loops are happening in the background.
image


apps/backoffice-v2/src/pages/Root/ServerDown.layout.tsx
```diff
@@ -0,0 +1,16 @@
...
+import { ErrorAlert } from '../../common/components/atoms/ErrorAlert/ErrorAlert';
+
+export const ServerDownLayout: FunctionComponent = () => {
+  const { isSuccess } = useHealthQuery();
+
+  if (isSuccess) return <Outlet />;
+
+  return (
+    <main className={`flex h-full flex-col items-center`}>
+      <ErrorAlert>Our services are temporarily down.</ErrorAlert>
```
alonp99, line 13: Sign it off with Nitzan please

alonp99 (approved):

PR #1557, opened 2023-11-04
Add Config Param to Document Change Event in Workflows Service

alonp99:
This PR introduces the following changes to the workflows-service:

- `document-changed-webhook-caller.ts` now accepts an `EventConfig` parameter to dictate event handling behavior.
- `workflow-event-emitter.service.ts` has been updated to handle the new `EventConfig` parameter.
- `workflow.controller.internal.ts` includes a new endpoint for system event emission with admin guard.
- `workflow.service.ts` supports emitting system workflow events, leveraging the new config parameter.

These changes enhance the flexibility of the event handling system, especially for cases where event emission should be forced.

### DIFF
--- a/services/workflows-service/src/events/document-changed-webhook-caller.ts
+++ b/services/workflows-service/src/events/document-changed-webhook-caller.ts
@@ -1,7 +1,10 @@
 /* eslint-disable @typescript-eslint/no-unsafe-assignment */
 /* eslint-disable @typescript-eslint/no-unsafe-return */
 
-import { WorkflowEventEmitterService } from '@/workflow/workflow-event-emitter.service';
+import {
+  EventConfig,
+  WorkflowEventEmitterService,
+} from '@/workflow/workflow-event-emitter.service';
 import { Injectable } from '@nestjs/common';
 import { HttpService } from '@nestjs/axios';
 import { AxiosInstance } from 'axios';
@@ -37,9 +40,9 @@ export class DocumentChangedWebhookCaller {
   ) {
     this.#__axios = this.httpService.axiosRef;
 
-    workflowEventEmitter.on('workflow.context.changed', async data => {
+    workflowEventEmitter.on('workflow.context.changed', async (data, config) => {
       try {
-        await this.handleWorkflowEvent(data);
+        await this.handleWorkflowEvent(data, config);
       } catch (error) {
         console.error(error);
         alertWebhookFailure(error);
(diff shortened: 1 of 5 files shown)


services/workflows-service/src/workflow/workflow-event-emitter.service.ts
```diff
@@ -2,21 +2,29 @@ import { EventEmitter2 } from '@nestjs/event-emitter';
...
 export class WorkflowEventEmitterService {
   constructor(private eventEmitter: EventEmitter2) {}
 
-  emit<TEvent extends TEventName>(eventName: TEvent, eventData: ExtractWorkflowEventData<TEvent>) {
+  emit<TEvent extends TEventName>(
+    eventName: TEvent,
+    eventData: ExtractWorkflowEventData<TEvent>,
+    config = { forceEmit: false },
+  ) {
     if (!eventName) {
```
liorzam, line 20: in case someone will use this config with a partial flags then forceEmit wont be present
extract { forceEmit: false } to a default contant and use merge config with default
```
const DEFAULT_CONFIG = { forceEmit: false };
  emit<TEvent extends TEventName>(
    eventName: TEvent,
    eventData: ExtractWorkflowEventData<TEvent>,
    config = { forceEmit: false },
  )
    config = {...DEFAULT_CONFIG, ...(config || {}) }
  ```

**Review by liorzam — APPROVED** [review]

File: services/workflows-service/src/workflow/workflow.controller.internal.ts [hunk 2]
```diff
@@ -218,6 +220,25 @@ export class WorkflowControllerInternal {
...
+  @UseGuards(AdminAuthGuard)
+  async emitWorkflowEvent(
+    @common.Param() params: EmitSystemParamInput,
+    @common.Body() data: EmitSystemBodyInput,
+  ) {
+    if (data.systemEventName !== 'workflow.context.changed') {
+      throw new common.BadRequestException(`Invalid system event name: ${data.systemEventName}`);
+    }
+    return await this.service.emitSystemWorkflowEvent({
+      workflowRuntimeId: params?.id,
```
liorzam: workflowRuntimeId cant be undefined

liorzam (changes requested):

services/workflows-service/src/events/document-changed-webhook-caller.ts
```diff
@@ -37,17 +40,20 @@ export class DocumentChangedWebhookCaller {
   ) {
     this.#__axios = this.httpService.axiosRef;
 
-    workflowEventEmitter.on('workflow.context.changed', async data => {
+    workflowEventEmitter.on('workflow.context.changed', async (data, config) => {
       try {
-        await this.handleWorkflowEvent(data);
+        await this.handleWorkflowEvent(data, config);
       } catch (error) {
         console.error(error);
```
liorzam, line 47: why do we have console.error ?! 
i'll change it in my PR

services/workflows-service/src/workflow/workflow-event-emitter.service.ts
alonp99: Done

PR #1588, opened 2023-11-08
Fixed a flicker of the server down layout on load

Omri-Levy:
### Description
Now checking for the loading state of the health check fetch request before displaying a layout.

### DIFF
--- a/apps/backoffice-v2/src/pages/Root/ServerDown.layout.tsx
+++ b/apps/backoffice-v2/src/pages/Root/ServerDown.layout.tsx
@@ -4,9 +4,9 @@ import { Outlet } from 'react-router-dom';
 import { ErrorAlert } from '../../common/components/atoms/ErrorAlert/ErrorAlert';
 
 export const ServerDownLayout: FunctionComponent = () => {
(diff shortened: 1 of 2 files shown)


apps/backoffice-v2/src/pages/Root/ServerDown.layout.tsx
```diff
@@ -4,9 +4,9 @@ import { Outlet } from 'react-router-dom';
 import { ErrorAlert } from '../../common/components/atoms/ErrorAlert/ErrorAlert';
 
 export const ServerDownLayout: FunctionComponent = () => {
-  const { isSuccess } = useHealthQuery();
+  const { isSuccess, isLoading } = useHealthQuery();
 
-  if (isSuccess) return <Outlet />;
+  if (isSuccess || isLoading) return <Outlet />;
```
tomer-shvadron: Are you sure we want to show `<Outlet />` when still loading?
Maybe a loader or something like that instead?
What happens on periodic check? Will the user experience a flicker?

tomer-shvadron (approved):

alonp99 (approved):

PR #1651, opened 2023-11-21
Fixed 401 session fetch loop

Omri-Levy:
### Description
Previously receiving a 401 error would repeatedly try to fetch session again.

### DIFF
--- a/apps/backoffice-v2/src/pages/Root/Root.error.tsx
+++ b/apps/backoffice-v2/src/pages/Root/Root.error.tsx
@@ -1,11 +1,28 @@
 import { Button } from '../../common/components/atoms/Button/Button';
-import { Link } from 'react-router-dom';
+import { Link, Navigate, useRouteError } from 'react-router-dom';
 import { FunctionComponent } from 'react';
 import { Providers } from '../../common/components/templates/Providers/Providers';
 import { Toaster } from 'react-hot-toast';
 import { ErrorAlert } from '../../common/components/atoms/ErrorAlert/ErrorAlert';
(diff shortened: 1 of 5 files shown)


apps/backoffice-v2/src/pages/Root/Root.error.tsx
```diff
@@ -1,11 +1,28 @@
...
 import { Toaster } from 'react-hot-toast';
 import { ErrorAlert } from '../../common/components/atoms/ErrorAlert/ErrorAlert';
+import { useAuthenticatedLayoutLogic } from '../../domains/auth/components/AuthenticatedLayout/hooks/useAuthenticatedLayoutLogic/useAuthenticatedLayoutLogic';
+import { isErrorWithMessage } from '@ballerine/common';
 
 export const RootError: FunctionComponent = () => {
+  const error = useRouteError();
+  const { redirectUnauthenticatedTo, location } = useAuthenticatedLayoutLogic();
+
+  if (isErrorWithMessage(error) && error.message === 'Unauthorized (401)') {
```
liorzblrn: it would be better to check status code instead of message

liorzblrn (approved):

apps/backoffice-v2/src/pages/Root/Root.error.tsx
Omri-Levy: Good chance the error does not have the code. Will ensure first thing tomorrow. 👍🏼

liorzblrn:
can be closed: 
https://github.com/ballerine-io/ballerine/pull/1659

PR #1653, opened 2023-11-21
Blokh/feat/data migration

tomer-shvadron:
Previously: 
we were unable to generate data migrations for each environment

Feature:
data migration scripts enabling updates of data in multiple environment support.

db:data-migration:generate - generates a new data migration version
db:data-migration:migrate - run the migrations currently in our data migration folder ( by environment )
db:data-migration:import - imports latest version of data migrations


services/workflows-service/prisma/data-migrations/migration-template.hbs
```diff
@@ -0,0 +1,8 @@
+import { PrismaClient } from '@prisma/client';
+import { INestApplicationContext } from "@nestjs/common";
+
+export const migrate = async(client: PrismaClient, app: INestApplicationContext) => {
+    await client.$transaction(async () =>{
+    // put your migration code here
+    });
+};
```
liorzam: in order to be more resilient each migration should be has its own logic for failed migration
i suggest to create a class that extends from our base class

than by override methods each specific migration can manipulate some hooks like 
* post data migration
* failure
* etc

services/workflows-service/prisma/schema.prisma
```diff
@@ -374,3 +374,14 @@ model WorkflowRuntimeDataToken {
   expiresAt             DateTime
   deletedAt             DateTime?
 }
+
+model DataMigrationVersion {
+  id            Int      @id @default(autoincrement())
```
liorzam: dont use auto increment for id!
id should be string according to the name of the migration 
in that way you'll be notice that you've run the migration already and the database will prevent any additional records for the same migration

services/workflows-service/prisma/schema.prisma
```diff
@@ -374,3 +374,14 @@ model WorkflowRuntimeDataToken {
...
+  version       String   @unique
+  migratedAt    DateTime @default(now())
+  success       Boolean  @default(true)
```
liorzam: instead of success column, it would be better to have a string column which represent a status
'PROCESSING'
'FAILED'
'COMPLETED'

services/workflows-service/prisma/schema.prisma
```diff
@@ -374,3 +374,14 @@ model WorkflowRuntimeDataToken {
   expiresAt             DateTime
   deletedAt             DateTime?
 }
+
+model DataMigrationVersion {
+  id            Int      @id @default(autoincrement())
+  version       String   @unique
```
liorzam: version should be PK

services/workflows-service/src/data-migration/scripts/migrate.ts
```diff
@@ -0,0 +1,140 @@
...
+
+const fetchEnvMigrationFiles = async () => {
+  let migrationFiles: Array<string> = [];
+  const commonMigrations = path.join(__dirname, DATA_MIGRATION_FOLDER_RELATIVE_PATH, 'common');
+  migrationFiles = migrationFiles.concat(await fetchMigrationFiles(commonMigrations));
+
+  const environmentName = env.ENVIRONMENT_NAME || '';
+  if (['local', 'dev', 'development', 'demo'].includes(environmentName)) {
+    migrationFiles = migrationFiles.concat(
+      await fetchMigrationFiles(path.join(__dirname, DATA_MIGRATION_FOLDER_RELATIVE_PATH, 'dev')),
```
liorzam, line 46: What about the option to configure migration folder from config|env var?
lets create a project on linear and add some leftover tasks like that one

services/workflows-service/src/data-migration/scripts/migrate.ts
```diff
@@ -0,0 +1,140 @@
...
+  return fullFilePath;
+};
+
+const fetchEnvMigrationFiles = async () => {
+  let migrationFiles: Array<string> = [];
+  const commonMigrations = path.join(__dirname, DATA_MIGRATION_FOLDER_RELATIVE_PATH, 'common');
+  migrationFiles = migrationFiles.concat(await fetchMigrationFiles(commonMigrations));
+
+  const environmentName = env.ENVIRONMENT_NAME || '';
+  if (['local', 'dev', 'development', 'demo'].includes(environmentName)) {
```
liorzam: optional values for env name:
`['prod', 'dev', 'staging', 'qa', 'demo', 'sb', 'test']`

services/workflows-service/src/data-migration/scripts/migrate.ts
```diff
@@ -0,0 +1,140 @@
...
+    );
+    try {
+      await migrationProcess.migrate(client, appContext);
+
+      await dataMigrationRepository.create({
+        data: {
+          version: migrationVersion,
+          success: true,
+        },
+      });
```
liorzam: what would happen in case of a failed migration that will run again?

services/workflows-service/src/data-migration/scripts/migrate.ts
```diff
@@ -0,0 +1,140 @@
...
+import { DATA_MIGRATION_FOLDER_RELATIVE_PATH } from './consts';
+
+interface IMigrationProcess {
+  version: string; // also the timestamp
+  migrate: (client: PrismaClient, app: INestApplicationContext) => Promise<void>;
+  fileName: string;
+}
+
+const fetchMigrationFiles = async (migrationFolderPath: string) => {
+  if (!fs.existsSync(migrationFolderPath)) {
```
liorzam, line 19: use application logger, and in case migration folder doesnt exists print a warning

services/workflows-service/src/data-migration/scripts/migrate.ts
```diff
@@ -0,0 +1,140 @@
...
+        ? Number(latestSuccessfulMigration?.version)
+        : 0;
+      return Number(migrationProcess.version) > latestMigrationVersion;
+    })
+    .sort((a, b) => Number(a.version) - Number(b.version));
+};
+
+migrate()
+  .then(() => {
+    console.log('Migration finished');
```
liorzam, line 157: replace with our logger

`const logger = app.get(AppLoggerService);`

services/workflows-service/src/providers/file/file-provider/aws-s3-file.service.ts
```diff
@@ -188,4 +188,84 @@ export class AwsS3FileService implements IStreamableFileProvider {
   }): string {
     return [customerName, directory, fileName].filter(pathPart => !!pathPart).join('/');
   }
+
+  async downloadFolderContent(
```
liorzam, line 192: you can remove related code to s3, alon told me you made it with submodule

liorzam (changes requested):

services/workflows-service/prisma/data-migrations/migration-template.hbs
liorzam: in that way, you can also add some additional class properties like `this.logger.`

liorzam:
Leftover: Readme.md, tests

services/workflows-service/src/data-migration/scripts/migrate.ts
```diff
@@ -0,0 +1,140 @@
...
+          success: true,
+        },
+      });
+    } catch (error) {
+      await dataMigrationRepository.create({
+        data: {
+          version: migrationVersion,
+          success: false,
+          failureReason: `Error in migration file: ${migrationProcess.fileName}: ${
+            isErrorWithMessage(error) && error.message
```
Blokh: stacktrace

services/workflows-service/prisma/data-migrations/migration-template.hbs
Blokh: resolved after discussing.
WE can import the logger from app

services/workflows-service/prisma/schema.prisma
Blokh: it is quite redundant. but not hurting anyone

services/workflows-service/.gitignore
```diff
@@ -9,3 +9,7 @@ dist
 *.log
 scripts/seed_*.ts
 vite.config.ts.timestamp-*.mjs
+prisma/data-migrations/dev/*
```
alonp99: Its not necessary since this folder is a submodule

services/workflows-service/src/data-migration/scripts/migrate.ts
```diff
@@ -0,0 +1,152 @@
...
+
+  for (const migrationProcess of migrationProcessesToRun) {
+    const migrationVersion = migrationProcess.version;
+    console.log(
+      `Running Data Migration: ${migrationProcess.fileName.split('/').pop()!.replace('.js', '')}`,
+    );
+    try {
+      await migrationProcess.migrate(client, appContext);
+
+      await dataMigrationRepository.create({
```
alonp99: Its safer to create this record before the migration runs, as "processing" state and then update it to completed or failed - if for any reason the migration hangs or crash the pod we can see it stuck on the db as "processing"

PR #1655, opened 2023-11-21
Open document in another tab

tomer-shvadron:
### Description
Added a new FE route that shows a specific document a specific workflow in a specific filter
Added a button that opens the selected document in a new tab

### DIFF
--- a/apps/backoffice-v2/src/pages/Document/Document.page.tsx
+++ b/apps/backoffice-v2/src/pages/Document/Document.page.tsx
@@ -1,0 +1,29 @@
+import { useParams, useSearchParams } from 'react-router-dom';
+import { useWorkflowQuery } from '../../domains/workflows/hooks/queries/useWorkflowQuery/useWorkflowQuery';
+import { Case } from '../Entity/components/Case/Case';
+
+export const DocumentLayout = () => {
+  const { entityId, documentId } = useParams();
+  const [searchParams] = useSearchParams();
+
+  const filterId = searchParams.get('filterId');
+  const { data: workflow, isLoading } = useWorkflowQuery({ workflowId: entityId, filterId });
+
+  if (isLoading) {
+    return;
+  }
+
+  const document = workflow.context.documents
+    .map(({ pages }) => pages)
+    .flat()
+    ?.find(({ ballerineFileId }) => ballerineFileId === documentId);
+
+  return (
+    <Case.Documents
+      hideOpenExternalButton
+      documents={[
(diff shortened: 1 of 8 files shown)


apps/backoffice-v2/src/pages/Document/Document.page.tsx
```diff
@@ -0,0 +1,29 @@
+import { useParams, useSearchParams } from 'react-router-dom';
+import { useWorkflowQuery } from '../../domains/workflows/hooks/queries/useWorkflowQuery/useWorkflowQuery';
+import { Case } from '../Entity/components/Case/Case';
+
+export const DocumentLayout = () => {
+  const { entityId, documentId } = useParams();
+  const [searchParams] = useSearchParams();
+
+  const filterId = searchParams.get('filterId');
```
Omri-Levy: In general know we have `useFilterId`. (no need for a change)

apps/backoffice-v2/src/pages/Document/Document.page.tsx
```diff
@@ -0,0 +1,29 @@
...
+  const { data: workflow, isLoading } = useWorkflowQuery({ workflowId: entityId, filterId });
+
+  if (isLoading) {
+    return;
+  }
+
+  const document = workflow.context.documents
+    .map(({ pages }) => pages)
+    .flat()
+    ?.find(({ ballerineFileId }) => ballerineFileId === documentId);
```
Omri-Levy: Likely to break without optional chaining.
```suggestion
  const document = workflow?.context?.documents
    ?.map(({ pages }) => pages)
    ?.flat()
    ?.find(({ ballerineFileId }) => ballerineFileId === documentId);
```

apps/backoffice-v2/src/pages/Entity/components/Case/hooks/useDocuments/useDocuments.tsx
```diff
@@ -87,6 +87,15 @@ export const useDocuments = (documents: IDocumentsProps['documents']) => {
     [],
   );
 
+  const onOpenDocumentInNewTab = useCallback(document => {
+    const url = `${location.href.substring(
+      0,
+      location.toString().indexOf(location.search),
+    )}/document/${document.id}?filterId=${new URLSearchParams(location.search).get('filterId')}`;
```
Omri-Levy: Is there a better way of doing this? Can we maybe at least break it to a variable or two and add a comment? A bit hard to read and understand.

Omri-Levy (approved):

PR #1688, opened 2023-11-25
feat: exclusive logging for important stuff

liorzam:
### Description
* We would like to stop toasing not relevant notifications like Unauthorized error

* We started to log our Axios fetch request

    -   Will intercept http service

    -   Api plugin

    -   Saleforce plugin

*  **Workflow Service**
    -   Every-time our backend couldn't fetch an http route url remotely, the server thrown 500 status code instead of 404
    -   We would like reduce redundant:

        **  Logs for incoming request
        
        **  Min log level for application logger

        **   Stop log none important responses for (401/404 status codes)

### Related issues
https://linear.app/ballerine/issue/BAL-888/remove-redundant-toast-in-backoffice-and-better-logging


apps/backoffice-v2/src/common/utils/fetcher/fetcher.ts
```diff
@@ -49,8 +49,6 @@ export const fetcher: IFetcher = async ({
         : message;
     }
 
-    console.error(message);
-
```
Omri-Levy: We might want to at least make the log optional or provide some other indicator for errors.

apps/backoffice-v2/src/domains/workflows/hooks/mutations/useUpdateWorkflowByIdMutation/useUpdateWorkflowByIdMutation.tsx
```diff
@@ -48,8 +48,9 @@ export const useUpdateWorkflowByIdMutation = ({ workflowId }: { workflowId: stri
     onSuccess: (data, { action }) => {
       toast.success(t(`toast:${action}.success`));
     },
-    onError: (error, { action }, context) => {
-      toast.error(t(`toast:${action}.error`, { errorMessage: error.message }));
+    onError: (error: Error, { action }, context) => {
```
Omri-Levy: Aslong as a check like `instanceof Error` or a function like `isErrorWithMessage` is not in use error is of type `unknown`.

apps/backoffice-v2/src/domains/workflows/hooks/mutations/useUpdateWorkflowByIdMutation/useUpdateWorkflowByIdMutation.tsx
```diff
@@ -48,8 +48,9 @@ export const useUpdateWorkflowByIdMutation = ({ workflowId }: { workflowId: stri
...
+      const translatedError = t(`toast:${action}.error`, { errorMessage: error.message });
+      toast.error(translatedError);
```
Omri-Levy: A space between line 52 and 53 for readability would be appreciated. 🙏🏼

apps/backoffice-v2/src/lib/react-query/query-client.ts
```diff
@@ -15,6 +15,25 @@ export const isErrorWithCode = (error: unknown): error is IErrorWithCode => {
...
+
+async function clearAuthenticatedUser(queryClient: QueryClient) {
+  try {
+    await queryClient.cancelQueries();
+
+    queryClient.clear();
+
+    const authenticatedUser = authQueryKeys.authenticatedUser();
+
+    queryClient.setQueryData(authenticatedUser.queryKey, {
```
Omri-Levy, line 17: Does this matter after `queryClient.cancelQueries`?

apps/backoffice-v2/src/lib/react-query/query-client.ts
```diff
@@ -15,6 +15,25 @@ export const isErrorWithCode = (error: unknown): error is IErrorWithCode => {
   return isObject(error) && 'code' in error && typeof error.code === 'number';
 };
 
+export const isErrorWithMessage = (error: unknown): error is IErrorWithMessage => {
+  return isObject(error) && 'message' in error && typeof error.message === 'string';
+};
```
Omri-Levy: Why not use the version from `@ballerine/common`?

apps/backoffice-v2/src/lib/react-query/query-client.ts
```diff
@@ -27,28 +46,32 @@ export const queryClient = new QueryClient({
...
-
         return;
       }
 
-      if (!isErrorWithMessage(error) || error.message === 'undefined' || error.message === 'null')
+      if (!isErrorWithMessage(error) || error.message === 'undefined' || error.message === 'null') {
         return;
+      }
+
+      if (isErrorWithCode(error)) {
```
Omri-Levy: Is there a likelihood that an error would have a code but no message or vice versa and we'd still like to render a toast?

services/workflows-service/src/common/filters/AllExceptions.filter.ts
```diff
@@ -10,19 +14,86 @@ export class AllExceptionsFilter extends BaseExceptionFilter {
...
+      status,
+      error: inspect(errorRes),
+      message: error.message,
+      responseTime: Date.now() - request.startTime,
     });
   }
+
+  private _isObject(obj: unknown): obj is object {
+    return obj !== null && typeof obj === 'object';
+  }
```
Omri-Levy: We already have many versions of this function. Please use an existing option from either `@ballerine/common` or `lodash`.

services/workflows-service/src/common/filters/session-exception.filter.ts
```diff
@@ -20,8 +20,8 @@ export class SessionExpiredExceptionFilter extends BaseExceptionFilter {
       const asyncLogout = util.promisify(req.logout.bind(req));
       await asyncLogout();
 
-      response.clearCookie('session');
-      response.clearCookie('session.sig');
+      // response.clearCookie('session');
+      // response.clearCookie('session.sig');
```
Omri-Levy: Maybe we should add a comment as for why this is commented out?

apps/backoffice-v2/src/lib/react-query/query-client.ts
```diff
@@ -27,28 +46,32 @@ export const queryClient = new QueryClient({
...
+          void clearAuthenticatedUser(queryClient);
+        }
 
-      toast.error(error.message, {
-        id: error.message,
-      });
+        // Dont toast for no important errors
+        if (
+          statusCode === undefined ||
+          (statusCode >= 500 && statusCode !== 401 && statusCode !== 404)
```
Omri-Levy: `statusCode` cannot be both equal or greater 500 and in the 400s. Additionally, with this change we're no longer displaying validation errors nor do we have an alternative.

services/workflows-service/src/storage/storage.controller.internal.ts
```diff
@@ -137,15 +154,15 @@ export class StorageControllerInternal {
...
+    const downloadedFilePath = await new HttpFileService({
+      client: this.httpService,
+      logger: this.logger,
+    }).download(persistedFile.uri, localeFilePath);
 
     return downloadedFilePath;
   }
 
-  __isImageUrl(persistedFile: File) {
+  _isUriScheme(persistedFile: File) {
```
Omri-Levy: Also, since its `z.string().url()` why was this change made?
```suggestion
  _isUri(persistedFile: File) {
```

services/workflows-service/src/health/consts.ts
```diff
@@ -0,0 +1,6 @@
+export const CONTROLLER_NAME = '_health';
+
+export const ROUTES = {
+  LIVE: 'live',
+  READY: 'ready',
+};
```
Omri-Levy: Note this is the syntax for enums as objects.
```suggestion
} as const;
```

apps/backoffice-v2/src/common/utils/handle-promise/handle-promise.ts
```diff
@@ -1,6 +1,6 @@
 export const handlePromise = async <TData>(
   promise: Promise<TData>,
-): Promise<[TData, undefined] | [undefined, Error]> => {
+): Promise<[Response<TData>, undefined] | [undefined, Error]> => {
```
Omri-Levy: `handlePromise` handles any promise and not only fetch/http related promises. If a need came up then its still possible to pass a type as a type argument.

services/workflows-service/src/common/interceptors/log-request.interceptor.ts
```diff
@@ -1,23 +1,45 @@
...
+  HttpStatus,
+  Injectable,
+  NestInterceptor,
+} from '@nestjs/common';
+import { isAxiosError } from 'axios';
+import { catchError, tap, throwError } from 'rxjs';
 
 @Injectable()
-export class LogRequestInterceptor implements NestInterceptor {
+export class AxiosRequestErrorInterceptor implements NestInterceptor {
```
liorzam: rename the file name

services/workflows-service/src/common/http-service/http-config.service.ts
```diff
@@ -0,0 +1,55 @@
+import { HttpModule, HttpModuleOptions, HttpModuleOptionsFactory } from '@nestjs/axios';
+import { HttpStatus, Injectable } from '@nestjs/common';
+import { ConfigModule, ConfigService } from '@nestjs/config';
+import { AppLoggerService } from '../app-logger/app-logger.service';
+
+export const RETRY_DELAY_IN_MS = 1500; // extract to config, delay retry
+
+const _httpStatusFromAxiosMap = {
```
liorzam: use this mapping in logger interceptor

services/workflows-service/src/common/middlewares/request-id.middleware.ts
```diff
@@ -1,38 +1,76 @@
...
+    };
+
+    const abortFn = () => {
+      cleanup();
+      this.logger.warn('Request aborted by the client');
+    };
+
+    const errorFn = (error: Error) => {
+      cleanup();
+      this.logger.log(`Request pipeline error`, { error });
```
liorzam, line 57: logger.error

alonp99 (approved):

apps/backoffice-v2/src/lib/react-query/query-client.ts
liorzam: Not sure, but i supposed it should be here because some component bind to this query client state ( so I wonder when use cancel if related pages will get rendered)

will check it

apps/backoffice-v2/src/lib/react-query/query-client.ts
liorzam: you're right it shouldnt happened, but we are not charge of all the errors that's come up from the query client,
so in case some other issue which is not related to HTTP operation will raise we still would like to show the error message

services/workflows-service/src/common/filters/AllExceptions.filter.ts
liorzam: yeah you can see no one is using this method, removed. 
thanks

liorzam: i used import `{ isObject } from '@ballerine/common';`

apps/backoffice-v2/src/lib/react-query/query-client.ts
liorzam: yeah you are right it wont cover any 4XX status
Do you think we should show 400 error code, i wonder if its not the developer fault instead of`Bad Request` toast which tell nothing to the customer

apps/backoffice-v2/src/lib/react-query/query-client.ts
liorzam: what you would expect us to do?
in that case we have an issue that we cant provide a readable message for our customers

liorzam: the logic is:
if we dont have a message we wont show any toast,
in case we have a message we still will check the status code to make sure its a code we care to tell the user about.

if both not permit its a problem because something wrong happened and we didnt mentioned it, a refresh can might help in this scenario ?

apps/backoffice-v2/src/common/utils/fetcher/fetcher.ts
```diff
@@ -15,13 +15,21 @@ export const fetcher: IFetcher = async ({
...
   isBlob = false,
+}: {
+  url: string;
+  method: string;
+  body: BodyInit | null;
+  headers: HeadersInit;
+  timeout: number;
+  options: Partial<RequestInit>;
+  isBlob: boolean;
 }) => {
```
Omri-Levy: `fetcher` receives its types from `IFetcher`. If the types are broken then we need to fix `IFetcher` instead.

apps/backoffice-v2/src/lib/react-query/query-client.ts
liorzam: its seems its works like that all over the application

apps/backoffice-v2/src/common/utils/fetcher/fetcher.ts
liorzam: reverted

apps/backoffice-v2/src/lib/react-query/query-client.ts
```diff
@@ -27,28 +33,34 @@ export const queryClient = new QueryClient({
...
-      if (isErrorWithCode(error) && error.code === 401) {
-        const authenticatedUser = authQueryKeys.authenticatedUser();
-
-        void queryClient.cancelQueries();
-        queryClient.setQueryData(authenticatedUser.queryKey, {
-          user: undefined,
-        });
-      }
-
+    onError: (error: unknown) => {
```
Omri-Levy: `error` is inferred. In general, it's better to prefer inference (for return type as well) unless it's a specific library code where there are consumers for the type or code that requires a specific signature, those being the exception, not the rule.

services/workflows-service/tsconfig.json
```diff
@@ -3,6 +3,7 @@
   "compilerOptions": {
     "emitDecoratorMetadata": true,
     "experimentalDecorators": true,
+    "useUnknownInCatchVariables": false,
```
Omri-Levy: Revert this change, please. 🙏🏼

apps/backoffice-v2/src/lib/react-query/query-client.ts
```diff
@@ -27,28 +33,34 @@ export const queryClient = new QueryClient({
...
+          error.message !== 'null'
+        ) {
+          // Dont toast for no important errors
+          if (![401, 403, 404].includes(status)) {
+            toast.error(error.message, {
+              id: error.message,
+            });
+          }
+        }
+      }
```
Omri-Levy: This block is getting hard to read. My suggestion is the extreme to the other direction so feel free to say if you see a middle-ground.
```suggestion
if (!isErrorWithCode(error)) return;

const status = error.code;

if (status === 401) {
  void clearAuthenticatedUser(queryClient);
}

if (!isErrorWithMessage(error) || error.message === 'undefined' || error.message === 'null') {
  return;
}

// Dont toast for no important errors
if ([401, 403, 404].includes(status)) return;

toast.error(error.message, {
  id: error.message,
});
```

apps/backoffice-v2/src/lib/react-query/query-client.ts
```diff
@@ -27,24 +33,27 @@ export const queryClient = new QueryClient({
...
+
+      if (!isErrorWithCode(error)) throw error;
+
+      const status = error.code;
+
+      // Dont toast for no important errors
+      if ([401, 403, 404].includes(status)) {
+        if (status === 401) {
+          void clearAuthenticatedUser(queryClient);
+        }
```
Omri-Levy: There's no reason for this if condition to be nested.

apps/backoffice-v2/src/lib/react-query/query-client.ts
```diff
@@ -27,24 +33,27 @@ export const queryClient = new QueryClient({
...
-      if (isErrorWithCode(error) && error.code === 401) {
-        const authenticatedUser = authQueryKeys.authenticatedUser();
-
-        void queryClient.cancelQueries();
-        queryClient.setQueryData(authenticatedUser.queryKey, {
-          user: undefined,
-        });
-      }
-
+    onError: (error: unknown) => {
```
Omri-Levy: My previous comment was resolved but the change have not been made. 🙂 
```suggestion
    onError: (error) => {
```

services/workflows-service/jest.config.cjs
```diff
@@ -5,7 +5,6 @@ module.exports = {
   testRegex: '(/__tests__/.*|(\\.|/)(unit|e2e|intg)\\.test)\\.ts$',
   moduleNameMapper: {
     '^@/(.*)$': '<rootDir>/src/$1',
-    axios: 'axios/dist/node/axios.cjs',
```
Omri-Levy, line 8: Without this line tests failed due to Axios. Is this no longer the case?

services/workflows-service/package.json
```diff
@@ -15,20 +15,19 @@
...
-    "test": "jest --verbose",
+    "test": "jest",
     "test:unit": "cross-env SKIP_DB_SETUP_TEARDOWN=true jest --testRegex '.*\\.unit\\.test\\.ts$'",
     "test:integration": "jest --testRegex '.*\\.intg\\.test\\.ts$'",
     "test:e2e": "jest --testRegex '.*\\.e2e\\.test\\.ts$'",
     "test:watch": "jest --verbose --watch",
     "seed": "ts-node scripts/seed.ts",
     "seed:ex": "ts-node scripts/seed-ex.ts",
-    "db:migrate-save": "prisma migrate dev",
+    "db:migrate-dev": "prisma migrate dev",
```
Omri-Levy, line 25: Good change! Worth notifying the team it has been made. 🙏🏼

liorzam:
In order to solve the test issues you can follow this link which specify what actually have done in order to fix those tests:

https://stackoverflow.com/questions/76491001/unit-testing-services-that-use-httpmodule/77676262#77676262

services/workflows-service/package.json
liorzam: no one actually use it

liorzam: it usually used when running db:init
```
"db:init": "npm run db:migrate-dev -- --name 'initial version' && npm run db:migrate-up seed",
```

services/workflows-service/src/common/http-service/utils.ts
```diff
@@ -0,0 +1,48 @@
+import { HttpException, HttpStatus } from '@nestjs/common';
+import { AxiosError, isAxiosError } from 'axios';
+
+const _httpStatusFromAxiosMap = {
```
Omri-Levy: Note the `_` prefix suggests this map is not in use.
```suggestion
const httpStatusFromAxiosMap = {
```

services/workflows-service/src/common/http-service/utils.ts
```diff
@@ -0,0 +1,48 @@
...
+  ENOTFOUND: HttpStatus.NOT_FOUND,
+  ECONNABORTED: HttpStatus.INTERNAL_SERVER_ERROR,
+  ERR_CANCELED: HttpStatus.INTERNAL_SERVER_ERROR, // connection canceled
+  ERR_NETWORK: HttpStatus.INTERNAL_SERVER_ERROR, // connection problems
+};
```
Omri-Levy: Remember to add `as const` to maps. 🙂 
```suggestion
} as const;
```

services/workflows-service/src/providers/file/file-provider/http-file.service.ts
```diff
@@ -1,45 +1,77 @@
...
   }
 
+  // private _shouldRetry(error: AxiosError) {
+  //   const { response } = error;
+  //   const isTooManyRequests =
+  //     response?.status === HttpStatusCode.TooManyRequests;
+  //   const isNetworkIssue = error.code === 'ENETUNREACH';
+
+  //   return isTooManyRequests || isNetworkIssue;
+  // }
```
Omri-Levy, line 53: Should these comments be removed?

services/workflows-service/src/test/helpers/nest-app-helper.ts
```diff
@@ -46,7 +50,9 @@ const aclFilterResponseInterceptor = {
 export const fetchServiceFromModule = async <T>(
   service: Type<T>,
   dependencies: Provider[] = [],
-  modules: Type<any>[] = [],
+  modules:
+    | Array<Type<any> | DynamicModule | Promise<DynamicModule> | ForwardReference>
+    | Type<unknown>[] = [],
```
Omri-Levy, line 55: Appreciated! 🙌🏼

services/workflows-service/package.json
Omri-Levy: It is used when making a change to the Prisma schema.

services/workflows-service/src/common/http-service/http-config.service.ts
```diff
@@ -0,0 +1,51 @@
...
+export class HttpConfigService implements HttpModuleOptionsFactory {
+  constructor(
+    private readonly configService: ConfigService,
+    private readonly logger: AppLoggerService,
+  ) {}
+
+  createHttpOptions(): HttpModuleOptions {
+    return {
+      timeout: this.configService.get('HTTP_TIMEOUT_IN_MS', 5000),
+      maxRedirects: this.configService.get('HTTP_MAX_REDIRECTS', 10),
```
Omri-Levy, line 18: Does it not make sense to use `env` from @t3-env here?

services/workflows-service/src/providers/file/file-provider/http-file.service.ts
liorzam: because this PR getting to big so i decided to add it as technical debt, this comment is still important

services/workflows-service/src/common/http-service/http-config.service.ts
liorzam: oh no this is the wrong way,
we should alway alway use ConfigService instead t3-env file...

As you can see here i already migrated the [env to our configService](https://github.com/ballerine-io/ballerine/blob/dev/services/workflows-service/src/app.module.ts#L64)
image

services/workflows-service/src/common/http-service/utils.ts
liorzam: is it in used here

Omri-Levy (approved):
LGTM

PR #1705, opened 2023-11-28
blokh/feat/wf-definition-versioning

Blokh:
- [x] split definition repository and service from the main workflow service 
- [x] generate logic of each new workflow runtime will run on the newest version ( in /run endpoint )
- [x] add functionality of generating and upgraded version of definition
- [x] Test the whole logic.
- [x] release for common


services/workflows-service/src/workflow-defintion/workflow-definition.service.intg.test.ts
```diff
@@ -0,0 +1,208 @@
...
+      '',
+      '',
+      'webhookSharedSecret',
+    );
+    project = await createProject(prismaService, customer, '3');
+    differentProject = await createProject(prismaService, customer, '4');
+  });
+
+  describe('#upgradeDefinitionVersion', () => {
+    it('should generate a new definition with the new version and updates the filter', async () => {
```
Blokh: and update only the relevant once

packages/common/src/utils/replace-null-with-undefined/replace-null-with-undefined.ts
```diff
@@ -0,0 +1,17 @@
...
+  }
+
+  const newObj: any = {};
+  for (const [key, value] of Object.entries(obj)) {
+    newObj[key] =
+      value === null ? undefined : replaceNullsWithUndefined(value as Record<string, object>);
+  }
+
+  return newObj;
+};
```
Omri-Levy: 1. Current naming doesn't seem to suggest specificity to objects
2. Working with the `Record` type over the `object` type is usually the intention and is easier to work with in this context
3. Reduce handles for you creating the new object, iteration, and typing
4. Using `unknown` makes sense due to the runtime checks, recursion, and handling `null`
5. Line 12~13 is redundant due to the `obj === null` check at the start of the function

```suggestion
export const replaceNullWithUndefinedInObjects = (obj: unknown) => {
  if (obj === null) {
    return undefined;
  }

  if (typeof obj !== 'object' || Array.isArray(obj)) {
    return obj;
  }

  const objWithoutNulls = Object.entries(obj).reduce((acc, [key, value]) => {
    acc[key] = replaceNullWithUndefinedInObjects(value);

    return acc;
  }, {} as Record<PropertyKey, unknown>);

  return objWithoutNulls;
};
```

packages/workflow-core/src/lib/utils/definition-validator/extensions-validator.ts
```diff
@@ -21,67 +23,75 @@ export const extensionsValidator = (extensions: WorkflowExtensions, states: Stat
       pluginKind === 'email' ||
       pluginKind === 'kyc'
     ) {
-      validateApiPlugin(plugin as unknown as ISerializableHttpPluginParams, states)
+      validateApiPlugin(plugin as unknown as ISerializableHttpPluginParams, states);
```
Omri-Levy: No change required - here is where a type guard of `isApiPlugin` or `isPlugin` would be used so type assertions won't be needed.

packages/workflow-core/src/lib/utils/definition-validator/extensions-validator.ts
```diff
@@ -21,67 +23,75 @@ export const extensionsValidator = (extensions: WorkflowExtensions, states: Stat
...
+  transformers: SerializableValidatableTransformer['transform'],
+) => {
   transformers.forEach(transform => {
     try {
-      if (transform.transformer === 'jmespath') validateMapping('jmespath', transform.mapping as string);
-      if (transform.transformer === 'json-logic') validateMapping('json-logic', transform.mapping as unknown as Record<string, unknown>)
+      if (transform.transformer === 'jmespath')
+        validateMapping('jmespath', transform.mapping as string);
+      if (transform.transformer === 'json-logic')
+        validateMapping('json-logic', transform.mapping as unknown as Record<string, unknown>);
```
Omri-Levy: Add missing curly braces to `if` statements please. Line 82 is a bit hard to read. 🙏🏼

packages/workflow-core/src/lib/utils/definition-validator/extensions-validator.ts
```diff
@@ -21,67 +23,75 @@ export const extensionsValidator = (extensions: WorkflowExtensions, states: Stat
...
     } catch (ex) {
-      throw new Error(`Invalid transform for ${pluginName} - type: ${transform.transformer} transformer: ${isErrorWithMessage(ex) ? ex.message : ''}`);
+      throw new Error(
+        `Invalid transform for ${pluginName} - type: ${transform.transformer} transformer: ${
+          isErrorWithMessage(ex) ? ex.message : ''
+        }`,
+      );
```
Omri-Levy: Consider using `JSON.stringify` on invalid values so the logs would show what was passed and not for example `[object Object]`

services/workflows-service/src/workflow-defintion/workflow-definition.module.ts
Omri-Levy, line 1: Great change! 💪🏼

services/workflows-service/src/workflow-defintion/workflow-definition.repository.ts
```diff
@@ -113,4 +113,14 @@ export class WorkflowDefinitionRepository {
       ),
     );
   }
+
+  async findLatestVersion(name: string, projectIds: TProjectIds) {
```
Omri-Levy:
```suggestion
  async findByLatestVersion(name: string, projectIds: TProjectIds) {
```

services/workflows-service/src/workflow/workflow.controller.internal.unit.test.ts
```diff
@@ -63,6 +63,8 @@ function buildWorkflowDeifintion(sequenceNum) {
       type: 'json-schema',
       schema: {},
     },
+    projectId: projectId,
+    isPublic: !projectId,
```
Omri-Levy: This seems like it could lead to confusion or flaky tests. Myabe `isPublic` should be an argument?

Omri-Levy (approved):

PR #1706, opened 2023-11-28
File upload fixes (end-to-end)

Omri-Levy:
- refactor(*): improved file upload handling by providing fallback to mime type and extension
- fix(workflow-core): fixed error on trying to transform optional response and added a clearer error
- fix(*): re-introduced file type into the system - missing type broke the backoffice


Blokh (approved):

services/workflows-service/prisma/schema.prisma
```diff
@@ -222,6 +222,7 @@ model File {
   mimeType         String?
   userId           String
   fileNameOnDisk   String
+  fileName         String?
```
liorzam, line 226: @Omri-Levy Create a data script that set file name for all the other rows in files table which use the migration process @Blokh built.

```
const { PrismaClient } = require('@prisma/client');
const path = require('path');
const prisma = new PrismaClient();

async function processFiles() {
  const pageSize = 1000; // Adjust as needed
  let skip = 0;
  let totalCount = 0;

  do {
    const files = await prisma.file.findMany({
      skip,
      take: pageSize,
    });

    if (files.length === 0) {
      break; // No more records to process
    }

    for (const file of files) {
      const fileName = extractFileNameFromUri(file.uri);
      console.log(`Processing file: ${fileName}`);
      
      // Update the file record with the extracted file name
      await prisma.file.update({
        where: { id: file.id },
        data: { fileName },
      });
    }

    totalCount += files.length;
    skip += pageSize;

    console.log(`Processed ${totalCount} files so far...`);
  } while (true);
}

function extractFileNameFromUri(uri) {
  return path.basename(uri);
}

processFiles();
```

liorzam: Later change fileName column to not null

services/workflows-service/src/providers/file/file.service.ts
```diff
@@ -266,22 +272,26 @@ export class FileService {
...
       const tmpFile = tmp.fileSync();
+
       tmpLocalFilePath = await sourceServiceProvider.download(sourceRemoteFileConfig, tmpFile.name);
     }
 
-    const file = await fromFile(tmpLocalFilePath ?? fileDetails.uri);
-
+    const fileType = await getFileMetadata({
+      file: tmpLocalFilePath || fileDetails.uri,
+      fileName: fileDetails.fileName || fileDetails.uri || '',
```
liorzam, line 290: You can have uri in file name column, try to extract the file name:

```
function extractFileNameFromUri(uri) {
  // Use the path module to extract the file name
  return path.basename(uri);
}
```

`extractFileNameFromUri(fileDetails.uri)`

services/workflows-service/src/providers/file/file.service.ts
```diff
@@ -266,22 +272,26 @@ export class FileService {
...
     if (shouldDownloadFromSource) {
       const tmpFile = tmp.fileSync();
+
       tmpLocalFilePath = await sourceServiceProvider.download(sourceRemoteFileConfig, tmpFile.name);
     }
 
-    const file = await fromFile(tmpLocalFilePath ?? fileDetails.uri);
-
+    const fileType = await getFileMetadata({
+      file: tmpLocalFilePath || fileDetails.uri,
```
liorzam, line 289: why did you switch to `||` instead of `??` operator ?

liorzam (changes requested):

services/workflows-service/src/common/get-file-extension-with-fallback/get-file-extension-with-fallback.ts
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * Returns file's extension and mime type with fallback for each.
+ * @param file - file's path or buffer
+ * @param fileName - file's name with extension
+ * @param preferFileName - if false, will prefer extracting the file's extension from the file's mime type over file's name
+ */
+export const getFileExtensionWithFallback = async ({
+  file,
+  fileName,
+  preferFileName = true,
```
liorzam: in case someone call you function like that:
`getFileExtensionWithFallback({ file: 'aa', fileName: 'bb'})` your preferFileName will be present with default value?

liorzam (changes requested):

services/workflows-service/src/workflow/workflow.service.ts
```diff
@@ -1590,15 +1591,33 @@ export class WorkflowService {
...
+          },
           entityId,
           projectId,
           customerName,
         );
 
         const ballerineFileId = documentPage.ballerineFileId || persistedFile?.id;
+        const mimeType =
+          persistedFile?.mimeType ||
+          mime.getType(persistedFile.fileName || persistedFile.uri || '');
```
liorzam, line 1616: Does mime package can handle uri path?

services/workflows-service/src/workflow/workflow.service.ts
```diff
@@ -1590,15 +1591,33 @@ export class WorkflowService {
         const documentId = document.id! || getDocumentId(document, false);
 
         const persistedFile = await this.fileService.copyToDestinationAndCreate(
-          { id: documentId, uri: documentPage.uri, provider: documentPage.provider },
+          {
+            id: documentId,
+            uri: documentPage.uri,
+            provider: documentPage.provider,
+            // Will be solved once DefaultContextSchema is typed by Typebox
```
liorzam, line 1601: add a prefix TODO: in your comment

services/workflows-service/src/providers/file/file.service.ts
Omri-Levy: To cover empty strings. As for `0` or `false` its similar reasoning to empty string. Currently we need stricter runtime validation and types to truly use `??` over `||` safely.

liorzam (changes requested):

services/workflows-service/src/common/get-file-extension-with-fallback/get-file-extension-with-fallback.ts
liorzam: { ...DEFAULT_OPTIONS, ...options }

services/workflows-service/src/common/get-file-extension-with-fallback/get-file-extension-with-fallback.ts
```diff
@@ -0,0 +1,38 @@
+import { getFileExtension } from '@/common/get-file-extension/get-file-extension';
+import mime from 'mime';
+import { getMimeType } from '@/common/get-mime-type/get-mime-type';
+import { getFileType } from '@/common/get-file-type/get-file-type';
+
+/**
+ * Returns file's extension and mime type with fallback for each.
```
liorzam: change

services/workflows-service/src/common/get-file-metadata/get-file-metadata.ts
```diff
@@ -0,0 +1,30 @@
+import { getFileExtensionWithFallback } from '@/common/get-file-extension-with-fallback/get-file-extension-with-fallback';
```
liorzam: move this function

services/workflows-service/src/common/get-file-extension-with-fallback/get-file-extension-with-fallback.ts
```diff
@@ -0,0 +1,38 @@
...
+  file,
+  fileName,
+  preferFileName = true,
+}: {
+  file: string | Buffer;
+  fileName?: string;
+  preferFileName?: boolean;
+}) => {
+  const fileMetadata = await getFileType(file);
+  const mimeType = await getMimeType({ file, fileName });
```
liorzam: you extract mime type twice

services/workflows-service/src/common/get-file-metadata/get-file-metadata.ts
```diff
@@ -0,0 +1,30 @@
...
+ * @param preferFileName - if false, will prefer extracting the file's extension from the file's mime type over file's name
+ */
+export const getFileMetadata = async ({
+  file,
+  fileName,
+  preferFileName = true,
+}: {
+  file: string | Buffer;
+  fileName?: string;
+  preferFileName?: boolean;
```
liorzam: we dont have to specify 'preferFileName' , you can just remove it and use fileName existance

liorzam (changes requested):

services/workflows-service/src/storage/storage.controller.external.ts
```diff
@@ -62,10 +63,11 @@ export class StorageControllerExternal {
       uri: file.location || String(file.path),
       fileNameOnDisk: String(file.path),
       fileNameInBucket: file.key,
+      fileName: file.originalname,
       // Probably wrong. Would require adding a relationship (Prisma) and using connect.
       userId: '',
       projectId: currentProjectId,
-      mimeType: file.mimetype,
+      mimeType: file.mimetype || mime.getType(file.originalname || ''),
```
liorzam: use get mime type with fallback

services/workflows-service/prisma/schema.prisma
Omri-Levy: I believe we should do this once a need arises, then allocate a time and priority.

services/workflows-service/src/common/get-file-extension-with-fallback/get-file-extension-with-fallback.ts
Omri-Levy: Default values work as intended.

services/workflows-service/src/common/get-file-extension-with-fallback/get-file-extension-with-fallback.ts
```diff
@@ -0,0 +1,32 @@
...
+  file: string | Buffer;
+  fileName?: string;
+}) => {
+  const fileMetadata = await getFileType(file);
+
+  if (fileMetadata?.ext) {
+    return fileMetadata.ext;
+  }
+
+  if (!fileMetadata?.ext && fileName) {
```
liorzam: at line 19 you check this ext is not null,
so this check is redundant you can remove it `!fileMetadata?.ext`

liorzam (changes requested):

services/workflows-service/src/common/get-file-extension-with-fallback/get-file-extension-with-fallback.ts
```diff
@@ -16,7 +15,6 @@ export const getFileExtensionWithFallback = async ({
   fileName?: string;
 }) => {
   const fileMetadata = await getFileType(file);
-  const mimeType = await getMimeType({ file, fileName });
```
liorzam: im not sure about this deletion 
you can extract mimetype using this function in case fileMetadata?.mime is not defined 

```js
let mimeType = fileMetadata?.mime
if(!mimeType) {
   mimeType = await getMimeType({ file, fileName });
}
```

liorzam (changes requested):

services/workflows-service/src/workflow/workflow.service.ts
Omri-Levy: Yes.

services/workflows-service/src/common/get-file-extension-with-fallback/get-file-extension-with-fallback.ts
Omri-Levy: This was a readability decision. Will remove it though.

services/workflows-service/src/common/get-file-extension-with-fallback/get-file-extension-with-fallback.ts
Omri-Levy: In the entire scope of `getFileExtensionWithFallback` the same steps happen as in `getMimeType`. If there's a specific one missing this change still contradicts other PR comments. (not running `file-type` twice)

liorzam (approved):

packages/common/src/schemas/documents/default-context-schema.ts
```diff
@@ -99,7 +99,26 @@ export const defaultContextSchema = Type.Object({
         version: Type.Optional(Type.Number()),
         pages: Type.Array(
           Type.Union([
-            Type.Object({ ballerineFileId: Type.String() }, { additionalProperties: false }),
+            Type.Object(
+              {
+                ballerineFileId: Type.String(),
```
alonp99, line 104: Please check it doesn't break existing workflows

services/workflows-service/src/common/get-file-extension-with-fallback/get-file-extension-with-fallback.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  if (fileName) {
+    return getFileExtension(fileName);
+  }
+
+  if (fileMetadata?.mime) {
+    return mime.getExtension(fileMetadata.mime) ?? undefined;
+  }
+
+  return;
```
alonp99, line 35: Maybe a log could help here to understand if none of the methods emits the file name, or maybe one of them did emit void

services/workflows-service/src/common/get-file-metadata/get-file-metadata.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  if (fileMetadata?.mime) {
+    mimeType = fileMetadata.mime;
+  }
+
+  if (!fileMetadata?.mime && fileName) {
+    mimeType = mime.getType(fileName) ?? undefined;
+  }
+
+  if (fileMetadata?.ext) {
```
alonp99, line 31: seems like duplicate code with `services/workflows-service/src/common/get-file-extension-with-fallback/get-file-extension-with-fallback.ts`

alonp99 (approved):

services/workflows-service/src/common/get-file-metadata/get-file-metadata.ts
Omri-Levy: It is. Previous PR comments suggested that helpers are both not clear enough and repeat expensive operations.

PR #1708, opened 2023-11-29
fix: multi-arch docker images for ballerine

pratapalakshmi:
If the image is pushed it should have the following :

image

Working on a branch in ballerine

image

### DIFF
--- a/.github/workflows/publish-backoffice.yml
+++ b/.github/workflows/publish-backoffice.yml
@@ -20,9 +20,6 @@ env:
 
 jobs:
   build-and-push-image:
-    strategy:
-      matrix:
-        platform: ["linux/arm64", "linux/amd64"]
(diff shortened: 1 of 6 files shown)


liorzam:
Do we want this to happen for each build?
Why we need it actually ,in case someone need a different OS/Arch he can build it for himself

alonp99 (approved):

pratapalakshmi:
> Do we want this to happen for each build? Why we need it actually ,in case someone need a different OS/Arch he can build it for himself


```
    paths:
      # Run this pipeline only if there are changes in specified path
      - "apps/backoffice-v2/**"
    branches:
      - dev
      - test
      - prod
      - staging
      - sb
      - demo
```

alonp99 (approved):

alonp99 (approved):

PR #1716, opened 2023-11-29
Blokh/feat/generate document schema configurations

Blokh:
generated document schema configurations for each document saved in DB
updated backoffice logic to use the configurations in the workflow defintion instead of common.

added uneditable type & category


packages/common/src/schemas/documents/workflow/documents/schemas/index.ts
```diff
@@ -17,6 +17,17 @@ export const getDocumentsByCountry = (countryCode: (typeof countryCodes)[number]
   return documentIdsByCountry[countryCode]?.() || [];
 };
 
+export const getDocumentSchemaByDefinition = (
```
alonp99: *Get document by country

services/workflows-service/prisma/schema.prisma
```diff
@@ -149,6 +149,7 @@ model WorkflowDefinition {
   definitionType     String
   definition         Json
   contextSchema      Json?
+  documentsSchema    Json?
```
alonp99, line 152: something like documentsConfig

alonp99 (approved):

PR #1717, opened 2023-11-30
CSV, XLSX, XLS end-to-end handling

Omri-Levy:
### Description
Added support for CSV, XLSX, and XLS in `@ballerine/common`, the collection flow, workflows service, and backoffice.


apps/backoffice-v2/src/domains/storage/fetchers.ts
```diff
@@ -37,14 +36,20 @@ export const fetchFileInfoById = async (fileId: string) => {
...
+  fileId,
+  withSignedUrl = true,
+}: {
+  fileId: string;
+  withSignedUrl?: boolean;
+}) => {
   const fileInfo = await fetchFileInfoById(fileId);
 
-  if (fileInfo.fileNameInBucket && env.VITE_FETCH_SIGNED_URL) {
+  if (withSignedUrl && fileInfo?.fileNameInBucket) {
```
Omri-Levy, line 48: Moved env condition higher as part of the `withSignedUrl` argument so the conditions for signed URL are clearer and easily overridable.

tomer-shvadron (approved):

PR #1719, opened 2023-11-30
feat: removed reduntant workflow fetch & updated component props & blocks upd

chesterkmr:
### Description
- removed reduntant workflow fetch from Details component
- added document selector prop
- updated blocks schemas to pass workflow in details component


apps/backoffice-v2/src/pages/Entity/components/Details/Details.tsx
```diff
@@ -1,26 +1,25 @@
...
+  selectDocuments = baseDetailsDocumentsSelector,
   onSubmit,
 }) => {
-  const { entityId } = useParams();
-  const filterId = useFilterId();
-  const { data: workflow } = useWorkflowQuery({ workflowId: entityId, filterId });
+  const documents = selectDocuments(workflow);
 
-  if (!value.data?.length) return;
+  if (!value.data?.length) return null;
```
Omri-Levy, line 16: Was this change due to an issue?

chesterkmr: `undefined` isnt considered as ReactElement, so with this line TS rendered issues with type missmatch, to fix that i replaced empty return with null to satifsfy TS.

Omri-Levy: Great work. 👍🏼

Omri-Levy (approved):

PR #1735, opened 2023-12-03
fix: fixed deletion of string properties

chesterkmr:
### Description
- Fixed deletion of string document properties in backoffice

Note: This PR doesnt fix issue where `date` properties could not be cleared.


apps/backoffice-v2/src/pages/Entity/components/EditableDetails/EditableDetails.tsx
```diff
@@ -94,14 +94,20 @@ export const EditableDetails: FunctionComponent<IEditableDetails> = ({
...
-          propertyValue?.length === 16
-        ) {
+        // In case when date value is cleared its value should be set to null to perform succesfull merge.
+        // Currently date value could not be cleared because schemas doesnt not allow date values to be nullable.
+        // if (isDateProperty && !propertyValue) {
+        //   acc[curr] = null;
+        //   return acc;
+        // }
+
+        if (isDateProperty && typeof propertyValue === 'string' && propertyValue?.length === 16) {
```
Omri-Levy, line 109: The check for this logic should not include `format` of `date` as they are not the same.

chesterkmr: How its make any difference if they both work with date picker input?

Omri-Levy: Two different types of inputs based on two distinct values. One includes time. (hours and minutes)

Omri-Levy (approved):

PR #1764, opened 2023-12-05
fix: fixed bug where DateInput started losing focus when entering year

chesterkmr:
### Description
- fixed bug where DateInput started losing focus when entering year
- ui bump

### DIFF
--- a/packages/ui/src/components/molecules/inputs/DatePickerInput/DatePickerInput.tsx
+++ b/packages/ui/src/components/molecules/inputs/DatePickerInput/DatePickerInput.tsx
@@ -2,11 +2,10 @@ import { LocalizationProvider } from '@mui/x-date-pickers/LocalizationProvider';
 import { AdapterDayjs } from '@mui/x-date-pickers/AdapterDayjs';
 import { DatePicker } from '@mui/x-date-pickers/DatePicker';
 import { FocusEvent, useCallback, useMemo, useState } from 'react';
-import { CalendarDays, ChevronLeft, ChevronRight } from 'lucide-react';
 import dayjs, { Dayjs } from 'dayjs';
-import { TextField, TextFieldProps, ThemeProvider } from '@mui/material';
+import { Paper, TextField, TextFieldProps, ThemeProvider } from '@mui/material';
 import { muiTheme } from '@/common/mui-theme';
-import { Paper } from '@/components/atoms';
+import { CalendarDays, ChevronLeft, ChevronRight } from 'lucide-react';
 
 export interface DatePickerChangeEvent {
   target: {
@@ -46,12 +45,19 @@ export const DatePickerInput = ({
     (value: Dayjs | null) => {
       if (!value) return onChange({ target: { value: null, name } });
 
-      onChange({
-        target: {
-          value: serializeValue(value),
-          name,
-        },
-      });
+      try {
+        const serializedDateValue = serializeValue(value);
+        onChange({
+          target: {
+            value: serializedDateValue,
+            name,
+          },
+        });
+      } catch (error) {
+        // Ignoring serialization due to partial date input
+        // Attept to serialize partially entered date e.g 12/MM/YYYY will cause exception
+        return null;
+      }
     },
     [name, onChange, serializeValue],
   );
@@ -101,7 +107,7 @@ export const DatePickerInput = ({
         />
       );
     },
-    [isFocused, onBlur],
+    [isFocused],
   );
 
   return (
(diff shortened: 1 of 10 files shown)


packages/ui/src/components/organisms/DynamicForm/components/RSJVInputAdaters/DateInputAdater/DateInputAdater.tsx
```diff
@@ -1,17 +1,47 @@
-import { DatePickerInput } from '@/components/molecules';
+import { DatePickerChangeEvent, DatePickerInput } from '@/components/molecules';
 import { RJSFInputAdapter } from '@/components/organisms/DynamicForm/components/RSJVInputAdaters/types';
 import { useCallback } from 'react';
 
-export const DateInputAdater: RJSFInputAdapter = ({ id, formData, disabled, onBlur, onChange }) => {
+export const isDateValid = (dateString: string): boolean => {
```
Omri-Levy: Also consider how this function doesn't check if the date is valid per se. 
```suggestion
export const isValidDate = (dateString: string): boolean => {
```

packages/ui/src/components/organisms/DynamicForm/components/RSJVInputAdaters/DateInputAdater/DateInputAdater.tsx
```diff
@@ -1,17 +1,47 @@
...
     // @ts-ignore
     onBlur && onBlur(id, formData);
   }, [id, onBlur, formData]);
 
+  const handleChange = useCallback(
+    (event: DatePickerChangeEvent) => {
+      const dateValue: string | null = event.target.value;
+      if (dateValue === null) return onChange(null);
+
+      const dateString = String(event.target.value);
```
Omri-Levy: What purpose does `String` serve here?

packages/ui/src/components/organisms/DynamicForm/components/RSJVInputAdaters/DateInputAdater/DateInputAdater.tsx
```diff
@@ -1,17 +1,47 @@
...
+  onChange,
+}) => {
   const handleBlur = useCallback(() => {
     // @ts-ignore
     onBlur && onBlur(id, formData);
   }, [id, onBlur, formData]);
 
+  const handleChange = useCallback(
+    (event: DatePickerChangeEvent) => {
+      const dateValue: string | null = event.target.value;
```
Omri-Levy: Note explicit typing over leaning on inference tends to do more harm than good in application code in the macro. This line does not seem to justify the type.
```suggestion
      const dateValue = event.target.value;
      
```

packages/ui/src/components/molecules/inputs/DatePickerInput/DatePickerInput.tsx
```diff
@@ -2,11 +2,10 @@ import { LocalizationProvider } from '@mui/x-date-pickers/LocalizationProvider';
 import { AdapterDayjs } from '@mui/x-date-pickers/AdapterDayjs';
 import { DatePicker } from '@mui/x-date-pickers/DatePicker';
 import { FocusEvent, useCallback, useMemo, useState } from 'react';
-import { CalendarDays, ChevronLeft, ChevronRight } from 'lucide-react';
 import dayjs, { Dayjs } from 'dayjs';
-import { TextField, TextFieldProps, ThemeProvider } from '@mui/material';
+import { Paper, TextField, TextFieldProps, ThemeProvider } from '@mui/material';
```
Omri-Levy: Is the `Paper` change intentional?

packages/ui/src/components/organisms/DynamicForm/components/RSJVInputAdaters/DateInputAdater/DateInputAdater.tsx
```diff
@@ -1,17 +1,47 @@
...
+      if (dateValue === null) return onChange(null);
+
+      const dateString = dateValue;
+
+      // every onChange call in context of DateInput forces re-render of component which eventually dicsonnects user focus from input
+      // in cases when date considered as valid e.g 05/12/0100 but doesnt make sense onChange call should be skipped
+      // validating date with helper below
+      if (!isValidDate(dateString)) return;
+
+      onChange(dateString);
```
Omri-Levy: TypeScript should be enough for where naming is lacking, re-assignment raises questions and adds unintentional meaning.
```suggestion
      const dateValue = event.target.value;
      
      if (dateValue === null) return onChange(null);

      // every onChange call in context of DateInput forces re-render of component which eventually dicsonnects user focus from input
      // in cases when date considered as valid e.g 05/12/0100 but doesnt make sense onChange call should be skipped
      // validating date with helper below
      if (!isValidDate(dateValue)) return;

      onChange(dateValue);
```

Omri-Levy (approved):

PR #1769, opened 2023-12-06
Bal 948

chesterkmr:
### Description
- Update controls on director documents blocks according to new requirements
- Updated breadcrumbs label to use bold text when active

image
image
image
image


apps/kyb-app/src/components/atoms/Stepper/components/atoms/Breadcrumbs/Breadcrumbs.Label.tsx
```diff
@@ -1,19 +1,12 @@
...
-  ) : LabelComponent ? (
-    <LabelComponent text={text} />
-  ) : null;
+  return (
+    <BaseLabel
+      className={ctw({ 'font-bold': active, 'opacity-50': state === 'completed' })}
+      text={text}
+    />
+  );
 };
```
Omri-Levy: Creating a component for two CSS classes seems like overkill. Can't these classes be passed from outside?

apps/backoffice-v2/src/pages/Entity/hooks/useTasks/hooks/useDirectorsBlocks/useDirectorsBlocks.tsx
```diff
@@ -86,6 +78,97 @@ export const useDirectorsBlocks = (
...
+                        id: document.id,
+                        contextUpdateMethod: 'director',
+                        disabled:
+                          (!isDoneWithRevision && Boolean(document?.decision?.status)) || noAction,
+                      },
+                    },
+                  },
+                ]
+              : []),
+          ];
```
Omri-Levy: Can we refactor these spreads + ternaries to `if` statements instead? Seems like we have flows based on the state and that may also change over time.

alonp99 (approved):

apps/backoffice-v2/src/pages/Entity/hooks/useTasks/hooks/useDirectorsBlocks/useDirectorsBlocks.tsx
```diff
@@ -108,6 +82,105 @@ export const useDirectorsBlocks = (
...
+              variant: 'warning',
+              className: 'min-h-8 text-sm font-bold',
+            },
+          };
+
+          const reUploadedNeededBadge = {
+            type: 'badge',
+            value: (
+              <React.Fragment>
+                Re-upload needed
```
alonp99: translation

alonp99 (approved):

apps/backoffice-v2/src/pages/Entity/hooks/useTasks/hooks/useDirectorsBlocks/useDirectorsBlocks.tsx
```diff
@@ -27,8 +26,8 @@ export const useDirectorsBlocks = (
   documentFiles: UseQueryResult[],
   documentImages: Array<Array<string>>,
 ) => {
-  const { mutate } = useRemoveDecisionTaskByIdMutation(
-    workflow?.id,
+  const { mutate: removeDecisionById } = useRemoveDecisionTaskByIdMutation(
+    workflow.id,
```
Omri-Levy: Spots like this should have optional chaining.

apps/backoffice-v2/src/pages/Entity/hooks/useTasks/hooks/useDirectorsBlocks/useDirectorsBlocks.tsx
```diff
@@ -53,37 +51,13 @@ export const useDirectorsBlocks = (
...
-  );
-
   const blocks = useMemo(() => {
     return directors
       .filter(director => Array.isArray(director.additionalInfo?.documents))
       .map(director => {
         const { documents } = director.additionalInfo;
 
-        const multiDocumentsBlocks = documents.map((document, docIndex) => {
+        const multiDocumentsBlocks = documents.map((document: AnyObject, docIndex: number) => {
```
Omri-Levy: Use `DefaultContextSchema['documents'][number]` or the `TDocument` type.

apps/backoffice-v2/src/pages/Entity/hooks/useTasks/hooks/useDirectorsBlocks/useDirectorsBlocks.tsx
```diff
@@ -108,6 +82,105 @@ export const useDirectorsBlocks = (
...
+          // Re-upload blocks end
+
+          // Approve blocks start
+          if (isApproved) {
+            documentHeading.push(approvedBadge);
+          } else {
+            if (!isRevision) {
+              documentHeading.push(approveButton);
+            }
+          }
```
Omri-Levy: `isApproved` and `isRevision` can never be true together. If somehow they do that's a bug or the wrong condition.
```suggestion
          if (isApproved) {
            documentHeading.push(approvedBadge);
          }
          
            if (!isRevision) {
              documentHeading.push(approveButton);
            }
```

apps/backoffice-v2/src/pages/Entity/hooks/useTasks/hooks/useDirectorsBlocks/useDirectorsBlocks.tsx
```diff
@@ -108,6 +82,105 @@ export const useDirectorsBlocks = (
...
+          const documentHeading = [];
+
+          // Re-upload blocks start
+          if (isRevision) {
+            if (workflow?.tags?.includes(StateTag.REVISION)) {
+              documentHeading.push(pendingReUploadBadge);
+            } else {
+              documentHeading.push(reUploadedNeededBadge);
+            }
+          }
```
Omri-Levy: Repeat yourself if it affects clarity and readability positively. These kinds of code blocks may include loops and other nesting levels overtime. This one can be avoided.
```suggestion
          if (isRevision && workflow?.tags?.includes(StateTag.REVISION)) {
               documentHeading.push(pendingReUploadBadge);
           }
           
           if (isRevision && !workflow?.tags?.includes(StateTag.REVISION)) {
              documentHeading.push(reUploadedNeededBadge);
           }
```

apps/backoffice-v2/src/pages/Entity/hooks/useTasks/hooks/useDirectorsBlocks/useDirectorsBlocks.tsx
```diff
@@ -108,6 +82,105 @@ export const useDirectorsBlocks = (
...
+              },
+            },
+          };
+
+          const documentHeading = [];
+
+          // Re-upload blocks start
+          if (isRevision) {
+            if (workflow?.tags?.includes(StateTag.REVISION)) {
+              documentHeading.push(pendingReUploadBadge);
```
Omri-Levy: In other places we have a function that returns an array of cells depending on the condition, see if it makes sense here as a pattern over using `Array.prototype.push`. Don't need to change for now but for future code have this in mind for consistency.

apps/kyb-app/src/components/atoms/Stepper/Stepper.tsx
Omri-Levy, line 1: Do we need an abstraction for a div with flex, width, and height? Also, check out a class called `d-full`.

apps/kyb-app/src/components/atoms/Stepper/components/atoms/Breadcrumbs/Breadcrumbs.Label.tsx
```diff
@@ -1,19 +1,15 @@
...
-import { StepperLabelsMap } from '@/components/atoms/Stepper/types';
-
-export const baseLabelsMap: StepperLabelsMap = {
-  idle: ({ text }) => <BaseLabel variant="idle" text={text} />,
-  warning: ({ text }) => <BaseLabel variant="warning" text={text} />,
-  completed: ({ text }) => <BaseLabel variant="completed" text={text} />,
-};
+import { ctw } from '@ballerine/ui';
 
 export const Label = ({ active, text, state }: BreadcrumbsLabelProps) => {
```
Omri-Levy, line 4: Using a `span` for a component called `Label` will cause wrong use. If its in use in forms and other places which expect semantic `label` tag we've hurt native benefits and accessiibility.

apps/backoffice-v2/src/pages/Entity/hooks/useTasks/hooks/useDirectorsBlocks/useDirectorsBlocks.tsx
chesterkmr: This is not relevant anymore. Reworked after integration with Blocks API

apps/kyb-app/src/components/atoms/Stepper/components/atoms/Breadcrumbs/Breadcrumbs.Label.tsx
chesterkmr: This label has nothing to do about forms. Its for breadcrumbs.

apps/backoffice-v2/src/lib/blocks/hooks/useDirectorsBlocks/helpers.ts
Omri-Levy, line 1: Don't think these reasons should be in the client or in the backoffice. Is there a reason these are not part of a schema or at the very least received from the workflows service?

apps/backoffice-v2/src/lib/blocks/hooks/useDirectorsBlocks/useDirectorsBlocks.tsx
```diff
@@ -17,10 +22,14 @@ import { selectDirectorsDocuments } from '@/pages/Entity/selectors/selectDirecto
...
 import { getPostDecisionEventName } from '../../components/CallToActionLegacy/hooks/useCallToActionLegacyLogic/useCallToActionLegacyLogic';
-import { valueOrNA } from '@/common/utils/value-or-na/value-or-na';
 import { createBlocksTyped } from '@/lib/blocks/create-blocks-typed/create-blocks-typed';
+import { X } from 'lucide-react';
+import { getRevisionReasonsForDocument } from '@/lib/blocks/components/DirectorsCallToAction/helpers';
+import { valueOrNA } from '@/common/utils/value-or-na/value-or-na';
+import { toTitleCase } from 'string-ts';
 
 export type Director = AnyObject;
+type DecisionStatus = NonNullable<DefaultContextSchema['documents'][number]['decision']>['status'];
```
Omri-Levy: Move to a co-located `types.ts`.

apps/backoffice-v2/src/lib/blocks/hooks/useDirectorsBlocks/helpers.ts
chesterkmr: They should be in schema indeed. But it will take huge effort to implement it now. Its out of this scope.

apps/backoffice-v2/src/lib/blocks/hooks/useDirectorsBlocks/useDirectorsBlocks.tsx
```diff
@@ -129,13 +138,100 @@ export const useDirectorsBlocks = ({
...
+                    },
+                  },
+                })
+                .build()
+                .flat(1);
+
+              return reUploadNeededBlock;
+            }
+
+            return undefined;
```
Omri-Levy, line 219: I understand this is to be explicit but I don't think in the macro `return undefined` adds enough value.

apps/backoffice-v2/src/lib/blocks/hooks/useDirectorsBlocks/useDirectorsBlocks.tsx
```diff
@@ -335,11 +407,6 @@ export const useDirectorsBlocks = ({
                     type: 'heading',
                     value: `Director - ${director.firstName} ${director.lastName}`,
                   })
-                  .addCell({
-                    id: 'actions',
-                    type: 'container',
-                    value: getReuploadActionOrBadge(workflow?.tags ?? []),
-                  })
```
Omri-Levy, line 342: I don't understand why this block was removed?

apps/backoffice-v2/src/lib/blocks/hooks/useDirectorsBlocks/useDirectorsBlocks.tsx
```diff
@@ -148,33 +244,45 @@ export const useDirectorsBlocks = ({
                 })
                 .build()
                 .flat(1);
+
+              return approvedBadgeBlock;
+            } else {
```
Omri-Levy, line 242: I think its important for these functions to work like a switch case with a default over using `else`. Explicitly define a cell/block for each `decisionStatus` and return nothing, some fallback, or throw an error if no condition matched.

apps/backoffice-v2/src/lib/blocks/hooks/useDirectorsBlocks/useDirectorsBlocks.tsx
```diff
@@ -148,33 +244,45 @@ export const useDirectorsBlocks = ({
...
+            return undefined;
           };
 
+          const documentHeading = [
+            getReuploadStatusOrAction(document?.decision?.status, workflow),
+            getReUploadedNeededAction(document?.decision?.status, workflow),
+            getDecisionStatusOrAction(document?.decision?.status, workflow),
+          ]
+            .filter(Boolean)
+            .map(block => block?.flat(1)[0]);
```
Omri-Levy, line 277: Can you explain this code?

chesterkmr: Functions are returning `undefined` in some cases. So they are should be filtered out. You can see example at 1st and 2nd screenshots.

apps/backoffice-v2/src/lib/blocks/hooks/useDirectorsBlocks/useDirectorsBlocks.tsx
chesterkmr: Switch case won't work here because some conditions having multiple levels of check. `default` not needed as well, its either return control/badge or return nothing.

apps/backoffice-v2/src/lib/blocks/hooks/useDirectorsBlocks/useDirectorsBlocks.tsx
chesterkmr: First iteration used global actions for all documents. Now each document has its own actions.

apps/backoffice-v2/src/pages/Entity/hooks/useTasks/hooks/useDirectorsBlocks/useDirectorsBlocks.tsx
chesterkmr: Non relevant anymore

apps/backoffice-v2/src/lib/blocks/hooks/useDirectorsBlocks/useDirectorsBlocks.tsx
chesterkmr: This return just shows that undefined expected as well. Im aware that JS functions return undefined by default but this could be not obvious for someone else.

apps/backoffice-v2/src/lib/blocks/hooks/useDirectorsBlocks/helpers.ts
Omri-Levy: Let's make sure to open a tech debt task for this one.

apps/backoffice-v2/src/lib/blocks/hooks/useDirectorsBlocks/useDirectorsBlocks.tsx
Omri-Levy: Would like this to be a consistent pattern in our monorepo and one we decided on. Same for implicit returns.

apps/backoffice-v2/src/lib/blocks/hooks/useDirectorsBlocks/useDirectorsBlocks.tsx
Omri-Levy: Will be refactored once more statuses are introduced.

apps/backoffice-v2/src/lib/blocks/hooks/useDirectorsBlocks/useDirectorsBlocks.tsx
Omri-Levy: Let's revisit the control flow of state to block.

PR #1797, opened 2023-12-11
feat: add hpa support for ballerine apps

pratapalakshmi:
Add hpa support for ballerine apps


alonp99 (approved):

liorzam:
Please add value example for hpa on `deploy/helm/example.values.yaml`

PR #1801, opened 2023-12-11
Opening PDF in new tab

tomer-shvadron:
### Description
Opening PDF in new tab


apps/backoffice-v2/src/pages/Entity/components/Case/Case.Documents.Toolbar.tsx
```diff
@@ -26,20 +26,20 @@ export const DocumentsToolbar: FunctionComponent<{
 }) => {
   return (
     <div className={`absolute z-50 flex space-x-2 bottom-right-6`}>
+      {!hideOpenExternalButton && !isLoading && image?.id && (
```
Omri-Levy, line 29: Lets disable the button when loading (not 100% on the image id)

Omri-Levy (approved):

PR #1807, opened 2023-12-11
fix: fixed validation error duplicate on business address

chesterkmr:
### Description
- Fixed validation error duplicates on business address page

### DIFF
--- a/services/workflows-service/src/ui-definition/utils/collection-flow/3-business-address-info-page.ts
+++ b/services/workflows-service/src/ui-definition/utils/collection-flow/3-business-address-info-page.ts
@@ -29,10 +29,18 @@ const validationSchema = {
                     properties: {
                       isDifferentFromPhysical: { const: true },
                     },
-                    required: ['isDifferentFromPhysical'],
                   },
                   then: {
-                    required: ['street', 'streetNumber', 'city', 'country', 'physical'],
+                    required: ['physical'],
+                    errorMessage: {
+                      required: {
+                        street: 'errorMessage.required.street',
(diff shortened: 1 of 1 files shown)


services/workflows-service/src/ui-definition/utils/collection-flow/3-business-address-info-page.ts
```diff
@@ -29,10 +29,18 @@ const validationSchema = {
...
                       isDifferentFromPhysical: { const: true },
                     },
-                    required: ['isDifferentFromPhysical'],
                   },
                   then: {
-                    required: ['street', 'streetNumber', 'city', 'country', 'physical'],
+                    required: ['physical'],
+                    errorMessage: {
+                      required: {
+                        street: 'errorMessage.required.street',
```
tomer-shvadron, line 37: Are these error messages shown after the change of the `required` array?

tomer-shvadron (approved):

services/workflows-service/src/ui-definition/utils/collection-flow/3-business-address-info-page.ts
chesterkmr: @tomer-shvadron 
Duplicating property case in this case is not needed, json-schema extends values in then statement. So i removed them.

Issue was under missing default, due to missing default it generated error with path `entity.something.blabla.physical` which was incorrectly parsed in CollectionFlow and generated duplicated errors(In fact there were not duplicates, just recieved incorrect path to property and rendered in wrong place.)

alonp99 (approved):

Blokh (approved):

PR #1820, opened 2023-12-13
fix: fixed bug where unfocus of input generated additional space under input

chesterkmr:
### Description
- fixed bug where unfocus of input generated additional space under input


apps/kyb-app/src/components/organisms/UIRenderer/elements/JSONForm/components/DocumentField/DocumentField.tsx
```diff
@@ -224,7 +224,9 @@ export const DocumentField = (
         onChange={handleChange}
       />
       {warnings.length ? <ErrorsList errors={warnings.map(err => err.message)} /> : null}
-      {isTouched ? <ErrorsList errors={validationErrors.map(error => error.message)} /> : null}
+      {isTouched && validationErrors.length ? (
+        <ErrorsList errors={validationErrors.map(error => error.message)} />
+      ) : null}
```
Omri-Levy: Please consider using `&&` over a ternary for JSX or at least the `null` use-case. In the macro its a lot of noise and hurts readability. for `warnings.length` using `!!` would avoid rendering `0`. The point about JSX is due to awkward formatting when using ternaries.

tomer-shvadron (approved):

apps/kyb-app/src/components/organisms/UIRenderer/elements/JSONForm/components/DocumentField/DocumentField.tsx
chesterkmr: Will stick to it.

PR #1822, opened 2023-12-13
blokh/feat/workflow-defintion-variants

Blokh:
added workflow defintion variants 
added common workflow variants


Omri-Levy (approved):

services/workflows-service/prisma/migrations/20231218141329_add_workflow_definition_variant/migration.sql
```diff
@@ -0,0 +1,2 @@
+-- AlterTable
+ALTER TABLE "WorkflowDefinition" ADD COLUMN     "variant" TEXT NOT NULL DEFAULT 'DEFAULT';
```
liorzam, line 2: NOT NULL for existing env is a bit problematic because default wont take in place for existing rows in workflow definitions, so this migration will fail for our deployed environments

liorzam (changes requested):

services/workflows-service/prisma/migrations/20231218141329_add_workflow_definition_variant/migration.sql
Blokh: I am not getting it.
so it will be "DEFAULT" until something else is set.

what is wrong here?

PR #1839, opened 2023-12-17
fix(ts): fix vscode issue with tsconifg path and eslint resolver

alonp99:


services/workflows-service/.eslintrc.cjs
```diff
@@ -10,9 +10,8 @@ module.exports = {
     'import/resolver': {
       typescript: {
         alwaysTryTypes: true,
-        project: './tsconfig.json',
+        project: ['**/tsconfig.json'],
```
Omri-Levy, line 13: Does this change leave a chance for the wrong tsconfig to be used?

services/workflows-service/.eslintrc.cjs
```diff
@@ -10,9 +10,8 @@ module.exports = {
...
       },
-      node: true,
```
Omri-Levy, line 15: Is changing `project` enough for the fix?

services/workflows-service/.eslintrc.cjs
alonp99: I dont think so, from what I read it seems to fit monorepos

Omri-Levy (approved):

PR #1843, opened 2023-12-18
Backoffice blocks variants (dependant on other PRs)

Omri-Levy:
### Description
Breaks combination of blocks to different "variants", allowing to render blocks based on workflow definition and have different actions for different variants.


apps/backoffice-v2/package.json
```diff
@@ -50,7 +50,7 @@
     "preview": "vite preview"
   },
   "dependencies": {
-    "@ballerine/common": "0.7.30",
```
alonp99: you might need to run changeset again since tomer pushed a version

apps/backoffice-v2/src/pages/Entity/Entity.page.tsx
```diff
@@ -33,81 +19,14 @@ export const Entity = () => {
...
-          })}
-        {kybChildWorkflows?.map(childWorkflow => (
-          <ChildDocumentBlocks
-            parentWorkflowId={workflow?.id}
-            childWorkflow={childWorkflow}
-            parentMachine={parentMachine}
-            key={childWorkflow?.id}
+        {workflow?.workflowDefinition && (
+          <BlocksVariant
+            workflowDefinition={{
```
alonp99, line 24: The naming is right here? when seeing "workflowDefinition" I would expect different content

apps/backoffice-v2/package.json
Blokh: +1

alonp99 (approved):

apps/backoffice-v2/src/pages/Entity/Entity.page.tsx
Omri-Levy: 1. Over time when more props are introduced I think its important to group related properties
2. I'd like to avoid passing a prop called `config`, `version`, and `variant` as its not clear enough where they should come from
3. It is the same type of `workflowDefinition` with `Pick`

apps/backoffice-v2/src/lib/blocks/variants/BlocksVariant/BlocksVariant.tsx
```diff
@@ -0,0 +1,28 @@
...
+import { DefaultBlocks } from '@/lib/blocks/variants/DefaultBlocks/DefaultBlocks';
+import { FunctionComponent } from 'react';
+import { TWorkflowById } from '@/domains/workflows/fetchers';
+import { WorkflowDefinitionVariant } from '@ballerine/common';
+
+export const BlocksVariant: FunctionComponent<{
+  workflowDefinition: Pick<TWorkflowById['workflowDefinition'], 'variant' | 'config' | 'version'>;
+}> = ({ workflowDefinition }) => {
+  if (
+    workflowDefinition?.version >= 0 &&
```
tomer-shvadron, line 12: For readability purposes (and assuming we're gonna have more variants in the future), I think it would be best if you extract the conditions in each `if` statement to a function whose name would describe what we're checking for

apps/backoffice-v2/src/lib/blocks/variants/DefaultBlocks/hooks/useDefaultBlocksLogic/useDefaultBlocksLogic.tsx
```diff
@@ -0,0 +1,396 @@
+import { useParams } from 'react-router-dom';
```
tomer-shvadron, line 1: This is a new 400 lines file with a lot of logic. Can we think of a way to simplify and shorten this file without compromising usability?

apps/backoffice-v2/src/lib/blocks/variants/KybDemoBlocks/hooks/useKybDemoBlocksLogic/useKybDemoBlocksLogic.tsx
```diff
@@ -0,0 +1,294 @@
+import { useParams } from 'react-router-dom';
```
tomer-shvadron, line 1: Same comment here as in `useDefaultBlocksLogic.tsx`, I really think that if we manage to minimize and better control the code in this files then we'd be able to manage them much more easily.

apps/backoffice-v2/src/lib/blocks/variants/ManualReviewBlocks/ManualReviewBlocks.tsx
```diff
@@ -0,0 +1,14 @@
+import { BlocksComponent } from '@/lib/blocks/components/BlocksComponent/BlocksComponent';
+import { useManualReviewBlocksLogic } from '@/lib/blocks/variants/ManualReviewBlocks/hooks/useManualReviewBlocksLogic/useManualReviewBlocksLogic';
+import { NoBlocks } from '@/lib/blocks/components/NoBlocks/NoBlocks';
+
+export const ManualReviewBlocks = () => {
+  const { blocks, isLoading } = useManualReviewBlocksLogic();
+
+  return (
+    <>
+      <BlocksComponent blocks={blocks} />
```
tomer-shvadron: Should we show `BlocksComponent` only if `blocks?.length` is truthy?

apps/backoffice-v2/src/pages/Entity/components/Case/hooks/useDocumentsToolbarLogic/useDocumentsToolbarLogic.tsx
```diff
@@ -15,7 +17,9 @@ export const useDocumentsToolbarLogic = ({
   hideOpenExternalButton,
   onOpenDocumentInNewTab,
 }: IUseDocumentsToolbarProps) => {
-  const { workflow } = useEntity();
+  const { entityId: workflowId } = useParams();
+  const filterId = useFilterId();
+  const { data: workflow } = useWorkflowQuery({ workflowId: workflowId, filterId });
```
tomer-shvadron, line 22: Thanks ❤️

tomer-shvadron (approved):

apps/backoffice-v2/src/lib/blocks/variants/ManualReviewBlocks/ManualReviewBlocks.tsx
Omri-Levy: Handled by the package.

PR #1877, opened 2023-12-26
Blokh/feat/execute all button refactoring

Blokh:
updated logic of execute all button to all be fulfilled in front-end


apps/backoffice-v2/src/pages/Entity/components/Case/hooks/usePendingRevisionEvents/usePendingRevisionEvents.tsx
```diff
@@ -0,0 +1,60 @@
+import { TWorkflowById } from '@/domains/workflows/fetchers';
+import { useCallback, useMemo } from 'react';
+import { calculateAllWorkflowEvents } from '@/pages/Entity/components/Case/hooks/usePendingRevisionEvents/utils/calculate-pending-workflow-events';
+import { CommonWorkflowEvent } from '@ballerine/common';
+import { checkIsKybExampleVariant } from '@/lib/blocks/variants/variant-checkers';
+import { useRevisionCaseMutation } from '@/domains/workflows/hooks/mutations/useRevisionCaseMutation/useRevisionCaseMutation';
+import { IPendingEvent } from './interfaces';
+
+function composeUniqueWorkflowEvents(
```
Omri-Levy: Please ensure consistency for syntax used for functions. Prefer arrow functions over function expressions.

apps/backoffice-v2/src/lib/blocks/components/ChildDocumentBlocks/hooks/useChildDocumentBlocksLogic/useChildDocumentBlocksLogic.tsx
```diff
@@ -40,11 +39,7 @@ export const useChildDocumentBlocksLogic = ({
   const isWorkflowLevelResolution =
     parentWorkflow?.workflowDefinition?.config?.workflowLevelResolution ??
     parentWorkflow?.context?.entity?.type === 'business';
-  const isKybExampleVariant = checkIsKybExampleVariant({
-    variant: parentWorkflow?.workflowDefinition?.variant,
-    config: parentWorkflow?.workflowDefinition?.config,
-    version: parentWorkflow?.workflowDefinition?.version,
-  });
+
```
Omri-Levy, line 44: Revert so associated company KYB block has its revision button disabled.

apps/backoffice-v2/src/lib/blocks/variants/variant-checkers.ts
```diff
@@ -2,11 +2,12 @@ import { TWorkflowById } from '@/domains/workflows/fetchers';
 import { WorkflowDefinitionVariant } from '@ballerine/common';
 
 export const checkIsKybExampleVariant = (
-  workflowDefinition: Pick<TWorkflowById['workflowDefinition'], 'variant' | 'config' | 'version'>,
+  workflowDefinition?: TWorkflowById['workflowDefinition'],
```
Omri-Levy: Revert. This function makes no sense with no arguments. If for some reason workflow definition is undefined it is a concern of the scope using the function and not the function itself.

apps/backoffice-v2/src/pages/Entity/components/Case/hooks/useCaseActionsLogic/useCaseActionsLogic.tsx
```diff
@@ -57,38 +55,10 @@ export const useCaseActionsLogic = ({ workflowId, fullName }: IUseActions) => {
 
   // Avoid passing the onClick event to mutate
   const onMutateApproveEntity = useCallback(() => mutateApproveEntity(), [mutateApproveEntity]);
-  const childPendingWorkflowEvents = useMemo(
-    () =>
-      workflow?.childWorkflows?.flatMap(
-        childWorkflow => childWorkflow?.context?.pendingWorkflowEvents,
-      ) ?? [],
-    [workflow?.childWorkflows],
+  const { onMutateRevisionCase, pendingWorkflowEvents } = usePendingRevisionEvents(
```
Omri-Levy, line 58: I understand the need to extract logic from `useCaseActionsLogic` but `usePendingRevisionEvents` should only return `pendingWorkflowEvents`. We can extract `onMutateRevisionCase` in some other way or at least achieve the same result via composition. (some hook consuming `usePendingRevisionEvents` etc.

apps/backoffice-v2/src/pages/Entity/components/Case/hooks/usePendingRevisionEvents/interfaces.ts
```diff
@@ -0,0 +1,6 @@
+export interface IPendingEvent {
+  workflowId: string;
+  documentId: string;
+  pendingEvent: string;
```
Omri-Levy: I think the pending event is the entire object and not the specific property. This naming will likely result in instances of `pendingEvent.pendingEvent`.
```suggestion
  event: string; // or name
```

apps/backoffice-v2/src/pages/Entity/components/Case/hooks/usePendingRevisionEvents/usePendingRevisionEvents.tsx
```diff
@@ -0,0 +1,60 @@
...
+  acc: Record<string, IPendingEvent>,
+  pendingWorkflowEvent: IPendingEvent,
+) {
+  acc[`${pendingWorkflowEvent?.workflowId}-${pendingWorkflowEvent?.pendingEvent}`] =
+    pendingWorkflowEvent;
+
+  return acc;
+}
+
+function isRevisionEvent(pendingWorkflowEvent: IPendingEvent) {
```
Omri-Levy: I think the problem here is the name of the function itself. It's better that there are several conditions and that they may evolve over time so the implementation is not reflected directly in the naming but `isRevisionEvent` gives the wrong impression.

apps/backoffice-v2/src/pages/Entity/components/Case/hooks/usePendingRevisionEvents/usePendingRevisionEvents.tsx
```diff
@@ -0,0 +1,60 @@
...
+  }, [workflow]);
+
+  const onMutateRevisionCase = useCallback(() => {
+    if (!pendingWorkflowEvents || !workflow) return;
+
+    const uniqueWorkflowEvents = pendingWorkflowEvents
+      .filter((pendingWorkflowEvent: IPendingEvent) => isRevisionEvent(pendingWorkflowEvent))
+      .reduce((acc: Record<string, IPendingEvent>, pendingWorkflowEvent: IPendingEvent) => {
+        return composeUniqueWorkflowEvents(acc, pendingWorkflowEvent);
+      }, {});
```
Omri-Levy: The logic here is very short and simple. Readability can remain the same if the logic is inline and not abstracted as long as curly braces are used with explicit returns.

Omri-Levy: Note `Array.prototype.reduce` is effectively the same as `Array.prototype.filter`.
```suggestion
const uniqueWorkflowEvents = pendingWorkflowEvents
    .reduce((acc: Record<string, IPendingEvent>, pendingWorkflowEvent: IPendingEvent) => {
        if (!isRevisionEvent(pendingWorkflowEvent)) {
            return acc;
        }
        
        return composeUniqueWorkflowEvents(acc, pendingWorkflowEvent);
    }, {});
```

apps/backoffice-v2/src/pages/Entity/components/Case/hooks/usePendingRevisionEvents/usePendingRevisionEvents.tsx
```diff
@@ -0,0 +1,60 @@
...
+    Object.keys(uniqueWorkflowEvents).forEach(pendingWorkflowKeys => {
+      const pendingWorkflowEvent = uniqueWorkflowEvents[pendingWorkflowKeys];
+      mutateRevisionCase({ workflowId: pendingWorkflowEvent!.workflowId });
+
+      const isKybExampleVariant = checkIsKybExampleVariant(workflow.workflowDefinition);
+      if (!isKybExampleVariant) return;
+
+      window.open(
+        `${workflow?.context?.metadata?.collectionFlowUrl}/?token=${pendingWorkflowEvent?.token}`,
+        pendingWorkflowEvent?.token,
```
Omri-Levy, line 55: Can you explain this second argument?

apps/backoffice-v2/src/pages/Entity/components/Case/hooks/usePendingRevisionEvents/utils/calculate-pending-workflow-events.ts
```diff
@@ -0,0 +1,27 @@
...
+  return workflow.context.documents
+    .map(document => {
+      if (!!document.decision.status && workflow?.tags?.includes(StateTag.MANUAL_REVIEW))
+        return {
+          workflowId: workflow.id,
+          documentId: document.id as string,
+          pendingEvent: calculateWorkflowRevisionableEvent(workflow, document.decision.status),
+          token: workflow?.context?.metadata?.token,
+        };
+    })
```
Omri-Levy: If you don't want an early return here at least add curly braces for the if statement. 🙏🏼

apps/backoffice-v2/src/pages/Entity/components/Case/hooks/usePendingRevisionEvents/utils/calculate-pending-workflow-events.ts
```diff
@@ -0,0 +1,27 @@
+import { TWorkflowById } from '@/domains/workflows/fetchers';
+import { StateTag } from '@ballerine/common';
+import { calculateWorkflowRevisionableEvent } from '@/pages/Entity/components/Case/hooks/usePendingRevisionEvents/utils/calculate-workflow-revisionable-event';
+import { IPendingEvent } from '@/pages/Entity/components/Case/hooks/usePendingRevisionEvents/interfaces';
+
+export const calculatePendingWorkflowEvents = (workflow: TWorkflowById): Array<IPendingEvent> => {
+  return workflow.context.documents
```
Omri-Levy, line 7: Consider `Array.prototype.reduce` for similar logic. It's possible to create an array of items matching the specified conditions.

apps/backoffice-v2/src/pages/Entity/components/Case/hooks/usePendingRevisionEvents/utils/calculate-pending-workflow-events.ts
```diff
@@ -0,0 +1,27 @@
...
+          workflowId: workflow.id,
+          documentId: document.id as string,
+          pendingEvent: calculateWorkflowRevisionableEvent(workflow, document.decision.status),
+          token: workflow?.context?.metadata?.token,
+        };
+    })
+    .filter((a): a is NonNullable<IPendingEvent> => !!a && !!a.pendingEvent);
+};
+
+export const calculateAllWorkflowEvents = (workflow: TWorkflowById): Array<IPendingEvent> => {
```
Omri-Levy: I think an indication for a recursive function is a good idea. 🙂 
```suggestion
export const calculateWorkflowEventsDeep = (workflow: TWorkflowById): Array<IPendingEvent> => {
```

apps/backoffice-v2/src/pages/Entity/components/Case/hooks/usePendingRevisionEvents/utils/calculate-workflow-revisionable-event.tsx
```diff
@@ -0,0 +1,20 @@
+import { TWorkflowById } from '@/domains/workflows/fetchers';
+import { getPostDecisionEventName } from '@/lib/blocks/components/CallToActionLegacy/hooks/useCallToActionLegacyLogic/useCallToActionLegacyLogic';
+import { CommonWorkflowEvent } from '@ballerine/common';
+
+export const calculateWorkflowRevisionableEvent = (
+  workflow: TWorkflowById,
+  documentStatus: string,
```
Omri-Levy, line 7: In general for types around context and documents `DefaultContextSchema` is now pretty accurate. Here you'd see `documentStatus` is "revision" | "approve" and so on. No change required, something to have in mind for the future.

apps/backoffice-v2/src/pages/Entity/components/Case/hooks/usePendingRevisionEvents/utils/calculate-workflow-revisionable-event.tsx
```diff
@@ -0,0 +1,20 @@
...
+) => {
+  const postDecisionEvent = getPostDecisionEventName(workflow);
+  const dispatchableEvent =
+    postDecisionEvent || documentStatus === CommonWorkflowEvent.REVISION
+      ? (CommonWorkflowEvent.REVISION as string)
+      : undefined;
```
Omri-Levy: Consider if statements over ternaries for similar logic. It does not seem the variable is in use beyond returning it.

examples/headless-example/playwright-report/index.html
Omri-Levy, line 1: Please add `playwright-report` to `.gitignore` and remove this file from version control.

apps/backoffice-v2/src/lib/blocks/variants/variant-checkers.ts
Blokh: yeah but the usage of this makes it unnecessary hard.

I agree that providing all the attributes as the required in order to generate a proper signature for the method.
e.g. 
 {
  variant: (typeof WorkflowDefinitionVariant)[keyof typeof WorkflowDefinitionVariant];
  config: TWorkflowById['workflowDefinition']['config'];
  version: number;
}

and yet it's unnecessarily difficult to pick specific attributes for a helper method regardless of what attributes are being used for 

and yet i will revert it because it has no real meaning.

apps/backoffice-v2/src/pages/Entity/components/Case/hooks/useCaseActionsLogic/useCaseActionsLogic.tsx
Blokh: so what exactly do you suggest? removing the hook of pending revision from useCaseAction?

apps/backoffice-v2/src/pages/Entity/components/Case/hooks/usePendingRevisionEvents/usePendingRevisionEvents.tsx
Blokh: i updated it to isPendingEventIsRevision, but I dont think it's still proper

apps/backoffice-v2/src/pages/Entity/components/Case/hooks/usePendingRevisionEvents/usePendingRevisionEvents.tsx
Blokh: I removed the curlies

Blokh: I disagree.
I don't think reduce should also filter. that's why we have filter for

apps/backoffice-v2/src/pages/Entity/components/Case/hooks/usePendingRevisionEvents/usePendingRevisionEvents.tsx
Blokh: the second argument is "target"

when working with tabs.
without target it will always open a new tab.

with target. if the tab with the same target is already opened it will just switch you to the specific tab.

apps/backoffice-v2/src/pages/Entity/components/Case/hooks/usePendingRevisionEvents/utils/calculate-pending-workflow-events.ts
Blokh: true you should have filter first before the map.

apps/backoffice-v2/src/pages/Entity/components/Case/hooks/usePendingRevisionEvents/utils/calculate-pending-workflow-events.ts
Blokh: i prefer to write business logic naming for a method instead of implementation details.
I updated it to:
calculateAllWorkflowPendingEvents

examples/headless-example/playwright-report/index.html
Blokh: it is for me the change came from Ilya probably. reverted it

Omri-Levy (approved):

PR #1908, opened 2024-01-08
Search term case status querying

tomer-shvadron:
### Description
* Search term querying
* Case Status querying


services/workflows-service/src/workflow/workflow.service.ts
```diff
@@ -433,24 +436,77 @@ export class WorkflowService {
...
+    const workflowDefinitionIdsParam = Prisma.join(
+      workflowDefinitionIds.map(id => Prisma.sql`${id}`),
     );
 
-    if (page.number > 1 && totalWorkflowsCount < (page.number - 1) * page.size + 1) {
+    const statusesParam = statuses.length
+      ? Prisma.join(statuses.map(status => Prisma.sql`${status}`))
+      : Prisma.sql``;
+
+    const projectIdsParam = Prisma.join((projectIds || []).map(id => Prisma.sql`${id}`));
```
alonp99: The default to an empty array here doesn't seem right, we should probably fail this action when there is no project filter

services/workflows-service/src/workflow/workflow.service.ts
```diff
@@ -433,24 +436,77 @@ export class WorkflowService {
...
+            array[${workflowDefinitionIdsParam}]::text[],
+            array[${statusesParam}]::text[],
+            array[${projectIdsParam}]::text[],
+            array[${assigneeIdsParam}]::text[],
+            array[${caseStatusParam}]::text[],
+            ${includeUnassigned}::boolean
+        )
+    `;
+
+    const workflowIds = await this.workflowRuntimeDataRepository.queryRawUnscoped<
```
alonp99: We shouldn't use unscoped repo methods, build a scoped function that enforces data isolation and remove these methods for the repository

services/workflows-service/src/workflow/workflow.service.ts
```diff
@@ -433,24 +436,77 @@ export class WorkflowService {
...
     );
 
-    const totalWorkflowsCount = await this.workflowRuntimeDataRepository.count(
-      {
-        where: query.where,
-      },
-      projectIds,
+    const { workflowDefinitionIds, statuses, assigneeIds, includeUnassigned } = {
+      workflowDefinitionIds:
+        query.where.workflowDefinitionId &&
```
alonp99: It's hard to understand why this code is needed, also its repeating itself - I think extracting it into a function will make it more readable and concise

apps/backoffice-v2/src/pages/Entities/Entities.page.tsx
```diff
@@ -40,7 +40,7 @@ export const Entities: FunctionComponent = () => {
               ? skeletonEntities.map(index => (
                   <Cases.SkeletonItem key={`cases-list-skeleton-${index}`} />
                 ))
-              : cases?.map(case_ => (
+              : (Array.isArray(cases) ? cases : []).map(case_ => (
```
Omri-Levy: Why this change? React doesn't render undefined and optional chaining prevents map from failing.

apps/backoffice-v2/src/pages/Entities/components/Cases/hooks/useCases/useCases.tsx
```diff
@@ -104,15 +152,16 @@ export const useCases = () => {
         if (!sortRef.current) break;
 
         sortRef.current.focus();
+        break;
       case 'f':
```
Omri-Levy: Adding curly braces allows using `const`. Up to you if to leave in these changes.
```suggestion
      case 'f': {
```

services/workflows-service/src/workflow/dtos/find-workflows-list.dto.ts
```diff
@@ -66,6 +73,7 @@ export const FindWorkflowsListSchema = z.object({
         .array(z.union([z.literal('').transform(() => null), z.string().nonempty()]))
         .optional(),
       status: z.array(z.nativeEnum(WorkflowRuntimeDataStatus)).optional(),
+      caseStatus: z.array(z.string()).optional(),
```
Omri-Levy, line 76: Do we need `caseStatus` each page? If so, maybe it should be named `state` or `caseState` for parity sake.

services/workflows-service/src/workflow/types/index.ts
```diff
@@ -140,3 +141,18 @@ export interface KYBParentKYCSessionExampleContext {
   };
   documents: any[];
 }
+
+export type FilterQuery = {
```
Omri-Levy, line 145: I think we can use the type from the filter zod schema.

services/workflows-service/src/workflow/workflow.service.ts
```diff
@@ -433,24 +436,77 @@ export class WorkflowService {
...
+        typeof query.where.workflowDefinitionId === 'object' &&
+        'in' in query.where.workflowDefinitionId &&
+        Array.isArray(query.where.workflowDefinitionId.in)
```
Omri-Levy: Pretty sure this code can be cut down to `Array.isArray(query.where.workflowDefinitionId.in)` as it also checks the previous conditions.

apps/backoffice-v2/src/pages/Entities/Entities.page.tsx
tomer-shvadron: I don't know why, but when developing I had ESLint issues there, but now I don't. Weird.

services/workflows-service/src/workflow/workflow.service.ts
tomer-shvadron: As discussed, it is safe

PR #1921, opened 2024-01-09
fix(*): fixed validation errors not propagating to the backoffice

Omri-Levy:
### Description

- Added a custom error for validation errors - formats messages
- Updated property names and paths to the new ones for accessing validation errors
- Usage of the new custom error now passes an errors array to logs and server response with a structure of message and path

### DIFF
--- a/services/workflows-service/src/common/filters/AllExceptions.filter.ts
+++ b/services/workflows-service/src/common/filters/AllExceptions.filter.ts
@@ -1,11 +1,11 @@
 import { env } from '@/env';
-import { isObject } from '@ballerine/common';
 import { AppLoggerService } from '@/common/app-logger/app-logger.service';
 import { inspect } from 'util';
 import { ArgumentsHost, Catch, HttpException, InternalServerErrorException } from '@nestjs/common';
 import { BaseExceptionFilter, HttpAdapterHost } from '@nestjs/core';
 import type { Request, Response } from 'express';
 import { HttpStatusCode } from 'axios';
+import { AjvValidationError } from '@/errors';
 
 @Catch()
 export class AllExceptionsFilter extends BaseExceptionFilter {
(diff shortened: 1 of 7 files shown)


services/workflows-service/src/common/filters/AllExceptions.filter.ts
```diff
@@ -34,49 +35,45 @@ export class AllExceptionsFilter extends BaseExceptionFilter {
...
+    });
   }
 
   isHttpException(error: unknown): error is HttpException {
-    return (
-      isObject(error) &&
-      'status' in error &&
-      typeof (error as HttpException).getStatus() === 'number'
-    );
+    return error instanceof HttpException;
```
Omri-Levy, line 57: People also reported an error due to `typeof (error as HttpException).getStatus() === 'number'` since non HttpException errors don't have `getStatus`.

alonp99 (approved):

services/workflows-service/src/common/filters/AllExceptions.filter.ts
```diff
@@ -23,6 +23,7 @@ export class AllExceptionsFilter extends BaseExceptionFilter {
 
   private _handleHttpErrorResponse(exception: unknown, request: Request, response: Response) {
     const serverError = this.getHttpException(exception);
+    const errors = exception instanceof AjvValidationError ? exception.serializeErrors() : [];
```
liorzam, line 26: why dont we use ajv.errorsText?
i've seen also seen this usage : services/workflows-service/src/workflow/workflow.service.ts:861
image

Omri-Levy: Good catch. Will see what I can do.

liorzam: So it would be better to use our custom error with a factory method for each provider...

PR #1933, opened 2024-01-09
Fix validation error message

liorzam:


liorzam:
@Omri-Levy You can remove this PR in case you start working on it 

1. Create a Nest interceptor for validation errors
2. Handle Zod and Ajv the same way
3. Change serializeErrors to -> serialize
4. Change output of serialize to return an object with a property of errors
5. Change the response message of validation errors to be Validation Error
6. Update the KYB app to handle the latest error changes
7. Ensure Sentry handles latest changes
8. Validation class should not extend Nest class - map status code and message elsewhere
9. Consider https://www.npmjs.com/package/http-status-codes to get HTTP messages and statuses - it exports enums and helpful methods

PR #1949, opened 2024-01-11
feat: case creation

chesterkmr:
### Description
- added Sheet & Switch components
- Implemented case generation


apps/backoffice-v2/src/common/hooks/useCurrentFilter/useCurrentFilter.ts
```diff
@@ -0,0 +1,16 @@
+import { useFilterId } from '@/common/hooks/useFilterId/useFilterId';
+import { useFiltersQuery } from '@/domains/filters/hooks/queries/useFiltersQuery/useFiltersQuery';
+import { useMemo } from 'react';
+
+export const useCurrentFilter = () => {
```
Omri-Levy, line 5: Sync with @tomer-shvadron he is working on something similar.

Omri-Levy, line 5: A lot of cases where "current" is used the current is implied. What do you think about "selected" or an indication its related to the URL?
```suggestion
export const useSelectedFilter = () => {
```

apps/backoffice-v2/src/common/hooks/useCurrentFilter/useCurrentFilter.ts
```diff
@@ -0,0 +1,16 @@
...
+  const filterId = useFilterId();
+  const { data: filters } = useFiltersQuery();
+
+  const currentFilter = useMemo(() => {
+    if (!Array.isArray(filters) || !filterId) return null;
```
Omri-Levy: Please use `undefined` over `null` unless there's a use of the `null` value for the use-case.

apps/backoffice-v2/src/domains/filters/fetchers.ts
```diff
@@ -1,14 +1,24 @@
-import { apiClient } from '../../common/api-client/api-client';
 import { z } from 'zod';
-import { handleZodError } from '../../common/utils/handle-zod-error/handle-zod-error';
+import { apiClient } from '../../common/api-client/api-client';
 import { Method } from '../../common/enums';
+import { handleZodError } from '../../common/utils/handle-zod-error/handle-zod-error';
 import { ObjectWithIdSchema } from '../../lib/zod/utils/object-with-id/object-with-id';
 
+export const BaseFilter = ObjectWithIdSchema.extend({
```
Omri-Levy: Keep schema naming convention.
```suggestion
export const BaseFilterSchema = ObjectWithIdSchema.extend({
```

apps/backoffice-v2/src/domains/filters/fetchers.ts
```diff
@@ -1,14 +1,24 @@
...
+});
+
+export type TFilter = z.output<typeof BaseFilter>;
+
 export const fetchFilterById = async (filterId: string) => {
   const [filter, error] = await apiClient({
     endpoint: `filters/${filterId}`,
     method: Method.GET,
-    schema: z.any(),
+    schema: BaseFilter,
```
Omri-Levy: Greatly appreciated schema change!

apps/backoffice-v2/src/domains/filters/fetchers.ts
Omri-Levy: The `Base` prefix is only needed for schemas that include self-references as `z.lazy` tends to break types. So you want to have a "base" with no relationships and a non-base for the more complicated and "real" schema.

apps/backoffice-v2/src/domains/workflows/fetchers.ts
```diff
@@ -54,20 +57,35 @@ export const fetchWorkflows = async (params: {
...
+  transitionSchema: z
+    .array(
+      z.object({
+        state: z.string(),
+        schema: z.record(z.any()),
+        additionalParameters: z.record(z.any()).optional(),
+      }),
+    )
+    .nullable(),
+  isManualCreationEnabled: z.boolean().optional(),
```
Omri-Levy: This should be inside `config`.

apps/backoffice-v2/src/domains/workflows/fetchers.ts
```diff
@@ -54,20 +57,35 @@ export const fetchWorkflows = async (params: {
...
+  definition: z.record(z.any()),
+  transitionSchema: z
+    .array(
+      z.object({
+        state: z.string(),
+        schema: z.record(z.any()),
+        additionalParameters: z.record(z.any()).optional(),
+      }),
+    )
+    .nullable(),
```
Omri-Levy: Should be passed by more team members but this seems to be a good point to group schemas under a single property..

apps/backoffice-v2/src/domains/workflows/fetchers.ts
```diff
@@ -54,20 +57,35 @@ export const fetchWorkflows = async (params: {
...
+  variant: z.string().default(WorkflowDefinitionVariant.DEFAULT),
+  contextSchema: z.record(z.any(), z.any()).nullable(),
+  documentsSchema: z.array(z.any()).optional().nullable(),
+  config: z.record(z.any(), z.any()).nullable(),
+  definition: z.record(z.any()),
+  transitionSchema: z
+    .array(
+      z.object({
+        state: z.string(),
+        schema: z.record(z.any()),
```
Omri-Levy: This schema is invalid. `z.record` expects two arguments.
```suggestion
        schema: z.record(z.string(), z.unknown()),
```

apps/backoffice-v2/src/domains/workflows/fetchers.ts
```diff
@@ -54,20 +57,35 @@ export const fetchWorkflows = async (params: {
...
+        additionalParameters: z.record(z.any()).optional(),
```
Omri-Levy: Same as comment above.

apps/backoffice-v2/src/domains/workflows/fetchers.ts
```diff
@@ -54,20 +57,35 @@ export const fetchWorkflows = async (params: {
 
 export type TWorkflowById = z.output<typeof WorkflowByIdSchema>;
 
+export const BaseWorkflowDefinition = ObjectWithIdSchema.extend({
```
Omri-Levy: Usage does not seem to require two schemas.
```suggestion
export const WorkflowDefinition = ObjectWithIdSchema.extend({
```

apps/backoffice-v2/src/domains/workflows/fetchers.ts
```diff
@@ -254,3 +272,45 @@ export const fetchWorkflowEventDecision = async ({
 
   return handleZodError(error, workflow);
 };
+
+export const createWorkflow = async ({
```
Omri-Levy: `createWorkflow` is a name used in the system as the way to use our SDKs, this might cause confusion.

apps/backoffice-v2/src/domains/workflows/fetchers.ts
```diff
@@ -254,3 +272,45 @@ export const fetchWorkflowEventDecision = async ({
...
+  workflowDefinitionId,
+  context,
+}: {
+  workflowDefinitionId: string;
+  context: AnyObject;
```
Omri-Levy:
```suggestion
  context: TWorkflowById['context'];
```

apps/backoffice-v2/src/domains/workflows/fetchers.ts
```diff
@@ -254,3 +272,45 @@ export const fetchWorkflowEventDecision = async ({
...
+    method: Method.POST,
+    url: `${getOriginUrl(env.VITE_API_URL)}/api/v1/external/workflows/run`,
+    body: {
+      workflowId: workflowDefinitionId,
+      context,
+    },
+    options: {
+      headers: { Authorization: 'Bearer clipspay_secret' },
+    },
+    schema: z.any(),
```
Omri-Levy, line 293: Can existing schemas be used here? i.e. the workflow by id schema.

apps/backoffice-v2/src/domains/workflows/hooks/mutations/useCreateWorkflowMutation/useCreateWorkflowMutation.ts
```diff
@@ -0,0 +1,35 @@
...
+        context,
+      }),
+    onSuccess: () => {
+      void queryClient.invalidateQueries();
+
+      toast.success('Case created successfully.');
+    },
+    onError: () => {
+      toast.error(`Failed to create case.`);
+    },
```
Omri-Levy: Use translations and use existing toast logic as a reference. Make sure to use same structure of words.

apps/backoffice-v2/src/domains/workflows/hooks/mutations/useCreateWorkflowMutation/useCreateWorkflowMutation.ts
```diff
@@ -0,0 +1,35 @@
...
+      context,
+    }: {
+      workflowDefinitionId: string;
+      context: AnyObject;
+    }) =>
+      createWorkflow({
+        workflowDefinitionId: workflowDefinitionId,
+        context,
+      }),
+    onSuccess: () => {
```
Omri-Levy, line 27: See if adding the new workflow to the workflows list as an optimistic update is low effort. Can copy existing optimistic logic to do so.

apps/backoffice-v2/src/domains/workflows/hooks/queries/useWorkflowDefinitionQuery/useWorkflowDefinitionQuery.ts
```diff
@@ -0,0 +1,14 @@
+import { useQuery } from '@tanstack/react-query';
+import { workflowsQueryKeys } from '../../../query-keys';
+
+export const useWorkflowDefinitionQuery = ({
```
Omri-Levy: Keep consistency with other queries.
```suggestion
export const useWorkflowDefinitionByIdQuery = ({
```

apps/backoffice-v2/src/domains/workflows/hooks/queries/useWorkflowDefinitionQuery/useWorkflowDefinitionQuery.ts
```diff
@@ -0,0 +1,14 @@
...
+  workflowDefinitionId,
+}: {
+  workflowDefinitionId: string;
+}) => {
+  return useQuery({
+    ...workflowsQueryKeys.workflowDefinition({ workflowDefinitionId }),
```
Omri-Levy: Ideally there'll be `workflowDefinitionQueryKeys` so the structure can remain the same of `list` and `byId` and open the possibility of creating and updating definitions without mixing into runtime.

apps/backoffice-v2/src/domains/workflows/query-keys.ts
```diff
@@ -30,6 +30,10 @@ export const workflowsQueryKeys = createQueryKeys('workflows', {
       queryFn: () => fetchWorkflows(data),
     };
   },
+  workflowDefinition: ({ workflowDefinitionId }: { workflowDefinitionId: string }) => ({
```
Omri-Levy: If separating to a different keys object is too much let's at least add a `ById` suffix.

apps/backoffice-v2/src/domains/workflows/utils/get-origin-url.ts
```diff
@@ -0,0 +1,5 @@
+export const getOriginUrl = (link: string) => {
```
Omri-Levy, line 1: Why is this needed?

apps/backoffice-v2/src/pages/Entities/components/CaseGeneration/CaseGeneration.tsx
```diff
@@ -0,0 +1,82 @@
+import { Button } from '@/common/components/atoms/Button/Button';
+import { SheetContent, SheetTrigger } from '@/common/components/atoms/Sheet';
+import { Sheet } from '@/common/components/atoms/Sheet/Sheet';
+import { CaseGenerationForm } from '@/pages/Entities/components/CaseGeneration/components/CaseGenerationForm';
+import { withCaseGeneration } from '@/pages/Entities/components/CaseGeneration/context/case-generation-context/hocs/withCaseGeneration';
+import { useCaseGenerationContext } from '@/pages/Entities/components/CaseGeneration/context/case-generation-context/hooks/useCaseGenerationContext';
+import { useCaseGenerationWorkflowDefinition } from '@/pages/Entities/components/CaseGeneration/hooks/useCaseGenerationWorkflowDefinition';
+import { PlusIcon } from '@radix-ui/react-icons';
+
+export const CaseGeneration = withCaseGeneration(() => {
```
Omri-Levy: Changes thus far use terminology around "create" and "creation" let's keep that consistent.
```suggestion
export const CaseCreation = withCaseCreation(() => {
```

apps/backoffice-v2/src/pages/Entities/components/CaseGeneration/CaseGeneration.tsx
```diff
@@ -0,0 +1,82 @@
...
+  const { workflowDefinition, isLoading, error } = useCaseGenerationWorkflowDefinition();
+  const { isOpen, setOpen } = useCaseGenerationContext();
+
+  return (
+    <Sheet open={isOpen} onOpenChange={setOpen}>
+      <SheetTrigger>
```
Omri-Levy: Might be missing `asChild`. Make sure we don't render a button within a button.

apps/backoffice-v2/src/pages/Entities/components/CaseGeneration/CaseGeneration.tsx
```diff
@@ -0,0 +1,82 @@
+import { Button } from '@/common/components/atoms/Button/Button';
+import { SheetContent, SheetTrigger } from '@/common/components/atoms/Sheet';
+import { Sheet } from '@/common/components/atoms/Sheet/Sheet';
+import { CaseGenerationForm } from '@/pages/Entities/components/CaseGeneration/components/CaseGenerationForm';
+import { withCaseGeneration } from '@/pages/Entities/components/CaseGeneration/context/case-generation-context/hocs/withCaseGeneration';
+import { useCaseGenerationContext } from '@/pages/Entities/components/CaseGeneration/context/case-generation-context/hooks/useCaseGenerationContext';
+import { useCaseGenerationWorkflowDefinition } from '@/pages/Entities/components/CaseGeneration/hooks/useCaseGenerationWorkflowDefinition';
+import { PlusIcon } from '@radix-ui/react-icons';
```
Omri-Levy: We use Lucide for icons in the backoffice.

apps/backoffice-v2/src/pages/Entities/components/CaseGeneration/CaseGeneration.tsx
```diff
@@ -0,0 +1,82 @@
...
+            <div>
+              {workflowDefinition ? (
+                <CaseGenerationForm workflowDefinition={workflowDefinition} />
+              ) : (
+                <p>Workflow definition is missing.</p>
+              )}
+            </div>
+          </div>
+        ) : (
+          <div>Loading workflow definition...</div>
```
Omri-Levy: Should we use any skeletons here?

apps/backoffice-v2/src/pages/Entities/components/CaseGeneration/CaseGeneration.tsx
```diff
@@ -0,0 +1,82 @@
...
+        )}
+        {!!error && <div>Failed to load workflow definition</div>}
```
Omri-Levy: Note React Query exposes `isSuccess` for places the error is not in use.

apps/backoffice-v2/src/pages/Entities/components/CaseGeneration/CaseGeneration.tsx
```diff
@@ -0,0 +1,82 @@
...
+      </SheetContent>
+    </Sheet>
+  );
+});
+
+const q = {
```
Omri-Levy: Is this object supposed to be removed?

apps/backoffice-v2/src/pages/Entities/components/CaseGeneration/components/CaseGenerationForm/CaseGenerationForm.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+const layouts = {
+  ...baseLayouts,
+  ButtonTemplates: {
+    ...baseLayouts.ButtonTemplates,
+    SubmitButton: SubmitSection,
+  },
+};
+
+export const CaseGenerationForm: FunctionComponent<FormProps> = ({ workflowDefinition }) => {
```
Omri-Levy: Same comment about "create" terminology.

apps/backoffice-v2/src/pages/Entities/components/CaseGeneration/components/CaseGenerationForm/CaseGenerationForm.tsx
```diff
@@ -0,0 +1,35 @@
...
+  workflowDefinition: TWorkflowDefinition;
+}
+
+const layouts = {
+  ...baseLayouts,
+  ButtonTemplates: {
+    ...baseLayouts.ButtonTemplates,
+    SubmitButton: SubmitSection,
+  },
+};
```
Omri-Levy: Might help with the type assertion below.
```suggestion
} satisfies typeof baseLayouts;
```

apps/backoffice-v2/src/pages/Entities/components/CaseGeneration/components/CaseGenerationForm/components/SubmitSection/SubmitSection.tsx
```diff
@@ -0,0 +1,31 @@
...
+    <div className="flex flex-row justify-end gap-6">
+      <div className="flex items-center space-x-2">
+        <Switch
+          checked={caseGenerationContext.isMultipleCasesCreation}
+          onClick={caseGenerationContext.toggleMultiCaseCreation}
+          id="add_more_switch"
+        />
+        <Label htmlFor="add_more_switch">Add more</Label>
+      </div>
+      {!norender && (
```
Omri-Levy: I don't understand why is this needed if there's an early return above and a disabled prop on the button?

apps/backoffice-v2/src/pages/Entities/components/CaseGeneration/components/CaseGenerationForm/hooks/useCaseGenerationForm/useCaseGenerationForm.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export const useCaseGenerationForm = (workflowDefinition: TWorkflowDefinition) => {
+  const { isLoading, mutateAsync } = useCreateWorkflowMutation();
+  const { isMultipleCasesCreation, setOpen } = useCaseGenerationContext();
+
+  const handleSubmit = useCallback(
+    async (formData: AnyObject) => {
+      const context = prepareContext(formData);
+
+      await mutateAsync({ workflowDefinitionId: workflowDefinition.id, context });
```
Omri-Levy: Why `mutateAsync` over `mutate`?

apps/backoffice-v2/src/pages/Entities/components/CaseGeneration/components/CaseGenerationForm/hooks/useCaseGenerationForm/utils/prepare-context.ts
```diff
@@ -0,0 +1,15 @@
+import { generateEntityId } from '@/pages/Entities/components/CaseGeneration/components/CaseGenerationForm/hooks/useCaseGenerationForm/utils/generate-entity-id';
+import { AnyObject } from '@ballerine/ui';
+
+export const prepareContext = (formData: AnyObject): AnyObject => {
```
Omri-Levy: We already use `create`, `compose`, `generate`, and `build`. Let's stick to ideally around one. 🙏🏼 
```suggestion
export const createInitialContext = (formData: AnyObject): AnyObject => {
```

apps/backoffice-v2/src/pages/Entities/components/CaseGeneration/components/CaseGenerationForm/hooks/useFormSchema/useFormSchema.ts
```diff
@@ -0,0 +1,26 @@
+import { TWorkflowDefinition } from '@/domains/workflows/fetchers';
+import { useMemo } from 'react';
+
+export const useFormSchema = (workflowDefinition: TWorkflowDefinition) => {
+  const transitionSchema = useMemo(
+    () =>
+      workflowDefinition?.transitionSchema?.find(
+        schema => schema.state === workflowDefinition?.definition?.initial,
+      ) ?? null,
```
Omri-Levy: Prefer `undefined` to `null` unless an is an exception. `null` from React Query and Prisma are just instances out of our control.
```suggestion
      ),
```

apps/backoffice-v2/src/pages/Entities/components/CaseGeneration/components/CaseGenerationForm/hooks/useFormSchema/useFormSchema.ts
```diff
@@ -0,0 +1,26 @@
...
+  );
+
+  const jsonSchema = useMemo(() => {
+    return transitionSchema?.schema || { type: 'object' };
+  }, [transitionSchema]);
+
+  const uiSchema = useMemo(
+    () => transitionSchema?.additionalParameters?.uiSchema || {},
+    [transitionSchema],
+  );
```
Omri-Levy: What's the benefit of `useMemo` here?

apps/backoffice-v2/src/pages/Entities/components/CaseGeneration/components/CaseGenerationForm/utils/transformErrors.ts
```diff
@@ -0,0 +1,48 @@
+import { AnyObject } from '@ballerine/ui';
+import { RJSFValidationError } from '@rjsf/utils';
+
+export const transformErrors = (errors: RJSFValidationError[]): RJSFValidationError[] => {
+  return errors.map(error => {
+    console.log('error', error);
```
Omri-Levy: Is this log supposed to be removed? If not, let's use `terminal` here or `console.error` like we do with Zod.

Omri-Levy: Why not log `errors` or the output of `transformErrors`?

apps/backoffice-v2/src/pages/Entities/components/CaseGeneration/components/CaseGenerationForm/utils/transformErrors.ts
```diff
@@ -0,0 +1,48 @@
...
+    if (error.name === 'minLength' || error.name === 'required') {
+      error.message = 'This field is required.';
```
Omri-Levy: Really don't think this function should mutate the data passed into it..

apps/backoffice-v2/src/pages/Entities/components/CaseGeneration/components/CaseGenerationForm/utils/transformErrors.ts
```diff
@@ -0,0 +1,48 @@
+import { AnyObject } from '@ballerine/ui';
+import { RJSFValidationError } from '@rjsf/utils';
+
+export const transformErrors = (errors: RJSFValidationError[]): RJSFValidationError[] => {
```
Omri-Levy: Do we transform errors differently in the backoffice compared to the KYB app or workflows service? Also, check out `Ajv.errorsText` as it might already provide a good format.

apps/backoffice-v2/src/pages/Entities/components/CaseGeneration/context/case-generation-context/case-generation.context.ts
```diff
@@ -0,0 +1,11 @@
+import { CaseGenerationContext } from '@/pages/Entities/components/CaseGeneration/context/case-generation-context/types';
+import { createContext } from 'react';
+
+export const initialCaseGenerationContext = {
+  isMultipleCasesCreation: false,
+  isOpen: false,
+} as CaseGenerationContext;
```
Omri-Levy: Prefer `satisfies` or `initialCaseGenerationContext: CaseGenerationContext` for this use.
```suggestion
} satisfies CaseGenerationContext;
```

apps/backoffice-v2/src/pages/Entities/components/CaseGeneration/context/case-generation-context/hocs/withCaseGeneration/withCaseGeneration.tsx
```diff
@@ -0,0 +1,17 @@
+import { CaseGenerationContextProvider } from '@/pages/Entities/components/CaseGeneration/context/case-generation-context/Provider';
+
+export function withCaseGeneration<TComponentProps extends object>(
```
Omri-Levy: Why do we need an HoC for a context provider or at least for the case generation?

apps/backoffice-v2/src/pages/Entities/components/CaseGeneration/context/case-generation-context/hooks/useCaseGenerationState/useCaseGenerationState.ts
```diff
@@ -0,0 +1,39 @@
...
+  >({
+    isMultipleCasesCreation: initialCaseGenerationContext.isMultipleCasesCreation,
+    isOpen: initialCaseGenerationContext.isOpen,
+  });
+
+  const toggleMultiCaseCreation = useCallback(() => {
+    setState(prev => ({ ...prev, isMultipleCasesCreation: !prev.isMultipleCasesCreation }));
+  }, []);
+
+  const setOpen = useCallback((open: boolean) => {
```
Omri-Levy: A setter should be named the same as the value it sets. 🙂 
```suggestion
  const setIsOpen = useCallback((open: boolean) => {
```

Omri-Levy: Then there won't be a need for `isOpen: open` either.
```suggestion
  const setOpen = useCallback((isOpen: boolean) => {
```

apps/backoffice-v2/src/pages/Entities/components/CaseGeneration/context/case-generation-context/hooks/useCaseGenerationState/useCaseGenerationState.ts
```diff
@@ -0,0 +1,39 @@
+import { initialCaseGenerationContext } from '@/pages/Entities/components/CaseGeneration/context/case-generation-context/case-generation.context';
+import {
+  CaseGenerationApi,
+  CaseGenerationContext,
+} from '@/pages/Entities/components/CaseGeneration/context/case-generation-context/types';
+import { useCallback, useMemo, useState } from 'react';
+
+export const useCaseGenerationAPI = () => {
```
Omri-Levy: My initial thought was this hook makes HTTP requests. Is using our existing naming for hooks for components business logic applicable here?
```suggestion
export const useCaseGenerationLogic = () => {
```

apps/backoffice-v2/src/pages/Entities/components/CaseGeneration/hooks/useCaseGenerationWorkflowDefinition/selectors/select-workflow-definition-id.ts
```diff
@@ -0,0 +1,8 @@
+import { TFilter } from '@/domains/filters/types';
+
+export const selectWorkflowDefinitionId = (obj: TFilter | null) => {
+  if (!obj) return null;
+
+  // TO DO: Find better place to aquire workflowDefinitionId
+  return obj?.query?.where?.workflowDefinitionId?.in[0] ?? null;
```
Omri-Levy: Remove nulls. 🙏🏼

apps/backoffice-v2/src/pages/Entities/components/CaseGeneration/hooks/useCaseGenerationWorkflowDefinition/useCaseGenerationWorkflowDefinition.ts
```diff
@@ -0,0 +1,26 @@
...
+  const {
+    data: workflowDefinition,
+    isLoading,
+    error,
+  } = useWorkflowDefinitionQuery({
+    workflowDefinitionId,
+  });
+
+  return {
+    workflowDefinition: workflowDefinition ?? null,
```
Omri-Levy: Remove `null`.

apps/backoffice-v2/src/common/hooks/useCurrentFilter/useCurrentFilter.ts
chesterkmr: `undefined` - Is a compiler value and means that something isnt exists. Why we should use it if variable is declared and in use?

apps/backoffice-v2/src/common/hooks/useCurrentFilter/useCurrentFilter.ts
chesterkmr: In current context im assuming that some filter is exists and i can retrieve it. I dont care what is source of this filter, if its from URL or selected somewhere by user. So in my opinion `useCurrent` fits well.

apps/backoffice-v2/src/domains/workflows/fetchers.ts
chesterkmr: @alonp99

apps/backoffice-v2/src/domains/workflows/fetchers.ts
chesterkmr: image

apps/backoffice-v2/src/common/hooks/useCurrentFilter/useCurrentFilter.ts
Omri-Levy: In this context `currentFilter` doesn't exist. `null` introduces extra code, and mental overhead, and makes data more awkward to work with. We should ideally stick to either `undefined` or `null`, not both, `undefined` makes more sense since it is the returned value from operations like `find` and is already in use throughout the system.

apps/backoffice-v2/src/domains/workflows/fetchers.ts
Omri-Levy: https://zod.dev/?id=record-key-type

apps/backoffice-v2/src/common/hooks/useCurrentFilter/useCurrentFilter.ts
chesterkmr: In current context value of `currentFilter` not exists but variable is declared and being processed by code. Also `null` return indicates that this value was returned by human and not compilier.

Methods such as `find` return `undefined` because returning `null` doesnt make sense.
`[].find(val => val === null)` // will return null in both cases no matter if value found or no.

If this is really concerning you then feel free to create poll & add such eslint rules

apps/backoffice-v2/src/domains/workflows/fetchers.ts
chesterkmr: image

chesterkmr: Fixed

apps/backoffice-v2/src/domains/workflows/hooks/mutations/useCreateWorkflowMutation/useCreateWorkflowMutation.ts
chesterkmr: Its not needed, queries are invalidated after creation so list is automatically reloaded. If i understood correctly what you ment.

apps/backoffice-v2/src/domains/workflows/query-keys.ts
chesterkmr: It has to be in its own query key file indeed. But i dont think its worth it to generate lots of boilerplate just for single endpoint in current case. Rename should work.

apps/backoffice-v2/src/domains/workflows/utils/get-origin-url.ts
chesterkmr: This helper needed to extract host name of api url so i can build link to any other endpoint excpect `api/v1/internal` as it is now. Currently this is somewhat issue, all endpoinds are binded to /internal endpoints. We should update all fetchers accordingly to endpoint destination.

apps/backoffice-v2/src/pages/Entities/components/CaseGeneration/CaseGeneration.tsx
chesterkmr: @alonp99 wdyt?

apps/backoffice-v2/src/pages/Entities/components/CaseGeneration/components/CaseGenerationForm/components/SubmitSection/SubmitSection.tsx
chesterkmr: It was mistake.

apps/backoffice-v2/src/pages/Entities/components/CaseGeneration/components/CaseGenerationForm/utils/transformErrors.ts
chesterkmr: KYB using AJV directly when json-form using some additional logic around json-schema. So currently i dont know how to use AJV and custom errors messages here.

apps/backoffice-v2/src/pages/Entities/components/CaseGeneration/context/case-generation-context/case-generation.context.ts
chesterkmr: Won't work here. Initial context missing methods.

apps/backoffice-v2/src/pages/Entities/components/CaseGeneration/context/case-generation-context/hocs/withCaseGeneration/withCaseGeneration.tsx
chesterkmr: HoC is needed to avoid generating additional component layers around CaseCreation so it could have access to context. Usage is here `src/pages/Entities/components/CaseCreation/CaseCreation.tsx`

apps/backoffice-v2/public/locales/en/toast.json
```diff
@@ -45,5 +45,9 @@
   "event": {
     "success": "Event sent successfully.",
     "error": "Error occurred while sending event."
+  },
+  "case_creation": {
+    "success": "Case created successfully.",
+    "error": "Failed to create case."
```
Omri-Levy:
```suggestion
    "error": "Error occurred while creating a case."
```

apps/backoffice-v2/src/domains/workflows/fetchers.ts
```diff
@@ -254,3 +271,39 @@ export const fetchWorkflowEventDecision = async ({
...
+      workflowId: workflowDefinitionId,
+      context,
+    },
+    schema: z.any(),
+  });
+
+  return handleZodError(error, workflow);
+};
+
+export const fetchWorkflowDefinition = async ({
```
Omri-Levy: Let's ensure consistency with other fetchers is kept. 🙏🏼 
```suggestion
export const fetchWorkflowDefinitionById = async ({
```

apps/backoffice-v2/src/domains/workflows/hooks/mutations/useCreateWorkflowMutation/useCreateWorkflowMutation.ts
Omri-Levy: The effort is key. If its low it can be nice to see a case immediately added to the UI on the cases list.

apps/backoffice-v2/src/pages/Entities/Entities.page.tsx
```diff
@@ -22,6 +24,7 @@ export const Entities: FunctionComponent = () => {
     totalPages,
     caseCount,
     skeletonEntities,
+    showCaseGeneration,
```
Omri-Levy:
```suggestion
    showCaseCreation,
```

apps/backoffice-v2/src/pages/Entities/components/CaseCreation/hooks/useCaseCreationWorkflowDefinition/selectors/select-workflow-definition-id.ts
```diff
@@ -0,0 +1,8 @@
+import { TFilter } from '@/domains/filters/types';
+
+export const selectWorkflowDefinitionId = (obj: TFilter | null) => {
```
Omri-Levy:
```suggestion
export const selectWorkflowDefinitionId = (filter: TFilter | null) => {
```

apps/backoffice-v2/src/pages/Entities/components/CaseCreation/context/case-creation-context/hocs/withCaseCreation/withCaseCreation.tsx
```diff
@@ -0,0 +1,17 @@
+import { CaseGenerationContextProvider } from '@/pages/Entities/components/CaseCreation/context/case-creation-context/Provider';
```
Omri-Levy:
```suggestion
import { CaseCreationContextProvider } from '@/pages/Entities/components/CaseCreation/context/case-creation-context/Provider';
```

apps/backoffice-v2/src/pages/Entities/components/CaseCreation/context/case-creation-context/case-creation.context.ts
```diff
@@ -0,0 +1,9 @@
+import { CaseGenerationContext } from '@/pages/Entities/components/CaseCreation/context/case-creation-context/types';
```
Omri-Levy: Just make sure to map all areas that have the word "generation". 🙏🏼

apps/backoffice-v2/src/domains/workflows/utils/get-origin-url.ts
Omri-Levy: Isn't this solved by the case management backend module?

apps/backoffice-v2/src/domains/workflows/fetchers.ts
alonp99: Yes, makes sense it will be in config - @chesterkmr you have other thoughts?

chesterkmr: Nope, sounds good. Wanted confirmation.

apps/backoffice-v2/src/domains/workflows/utils/get-origin-url.ts
chesterkmr: Nope, you cant reach it by using `internal` path piece.

apps/backoffice-v2/src/common/hooks/useCurrentFilter/useCurrentFilter.ts
Omri-Levy: Revisit PR post poll.

apps/backoffice-v2/src/pages/Entities/components/CaseCreation/components/CaseCreationForm/hooks/useCaseCreationForm/useCaseCreationForm.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export const useCaseCreationForm = (workflowDefinition: TWorkflowDefinition) => {
+  const { isLoading, mutateAsync } = useCreateWorkflowMutation();
+  const { isMultipleCasesCreation, setIsOpen: setOpen } = useCaseCreationContext();
+
+  const handleSubmit = useCallback(
+    async (formData: AnyObject) => {
+      const context = prepareContext(formData);
+
+      await mutateAsync({ workflowDefinitionId: workflowDefinition.id, context });
```
Omri-Levy, line 16: Any reason for using `mutateAsync` over `mutate`?

apps/backoffice-v2/src/pages/Entities/components/CaseCreation/components/CaseCreationForm/hooks/useCaseCreationForm/useCaseCreationForm.ts
```diff
@@ -0,0 +1,29 @@
...
+
+      if (!isMultipleCasesCreation) {
+        setOpen(false);
```
Omri-Levy: Why the alias from `setIsOpen`?

apps/backoffice-v2/src/pages/Entities/components/CaseCreation/components/CaseCreationForm/hooks/useCaseCreationForm/utils/prepare-context.ts
```diff
@@ -0,0 +1,16 @@
+import { TWorkflowById } from '@/domains/workflows/fetchers';
+import { generateEntityId } from '@/pages/Entities/components/CaseCreation/components/CaseCreationForm/hooks/useCaseCreationForm/utils/generate-entity-id';
+import { AnyObject } from '@ballerine/ui';
+
+export const prepareContext = (formData: AnyObject): TWorkflowById['context'] => {
```
Omri-Levy: I think the function's name can be clearer. We already have many variants like "merge", "create", "build", and "compose" just to name a few. What does "prepare" do?

services/workflows-service/src/case-management/types/transition-schema.ts
```diff
@@ -0,0 +1,5 @@
+export interface TransitionSchema {
```
Omri-Levy: Is there a schema we can infer the type from? Having an interface named like a schema is a bit confusing.
```suggestion
export interface ITransitionSchema {
```

services/workflows-service/src/workflow/workflow-service.module.ts
Omri-Levy, line 1: Any reason we can't move properties from `workflow.module.ts` into `workflow-service.module.ts` and rename the file to `workflow.module.ts`? Its a bit awkward having a file with both "service" and "module" in its name.

apps/backoffice-v2/src/domains/workflows/fetchers.ts
chesterkmr: done

Omri-Levy (approved):

services/workflows-service/src/workflow/workflow-controller.module.ts
```diff
@@ -0,0 +1,11 @@
+import { Module } from '@nestjs/common';
```
alonp99: `services/workflows-service/src/workflow/workflow-controller.module.ts` ? dosent make sense

services/workflows-service/src/case-management/types/transition-schema.ts
```diff
@@ -0,0 +1,5 @@
+export interface ITransitionSchema {
```
alonp99: transition schema in case managment types? should be in the workflow lib or at least workflow type in WF service

services/workflows-service/prisma/schema.prisma
```diff
@@ -155,6 +155,7 @@ model WorkflowDefinition {
 
   definitionType     String
   definition         Json
+  transitionSchema   Json?
```
alonp99: Why here and not inside the definition, near the transition itself? it will make us manage it in different places and can be hard to track

apps/backoffice-v2/src/common/hooks/useCurrentFilter/useCurrentFilter.ts
```diff
@@ -0,0 +1,16 @@
+import { useFilterId } from '@/common/hooks/useFilterId/useFilterId';
+import { useFiltersQuery } from '@/domains/filters/hooks/queries/useFiltersQuery/useFiltersQuery';
+import { useMemo } from 'react';
+
+export const useCurrentFilter = () => {
+  const filterId = useFilterId();
+  const { data: filters } = useFiltersQuery();
+
+  const currentFilter = useMemo(() => {
```
tomer-shvadron: inline variable and return

apps/backoffice-v2/src/pages/Entities/components/CaseCreation/hooks/useCaseCreationWorkflowDefinition/selectors/select-workflow-definition-id.ts
```diff
@@ -0,0 +1,8 @@
+import { TFilter } from '@/domains/filters/fetchers';
+
+export const selectWorkflowDefinitionId = (filter: TFilter | null) => {
+  if (!filter) return null;
+
+  // TO DO: Find better place to aquire workflowDefinitionId
```
tomer-shvadron: I know it sounds silly, but we should remove the space in the "TO DO" so we could always find it by searching "TODO"

services/workflows-service/prisma/migrations/20240110184558_add_transition_schema_to_workflow_definition/migration.sql
```diff
@@ -0,0 +1,2 @@
+-- AlterTable
+ALTER TABLE "WorkflowDefinition" ADD COLUMN     "transitionSchema" JSONB;
```
tomer-shvadron: What's `transitionSchema`?

tomer-shvadron (approved):

services/workflows-service/src/workflow/workflow-controller.module.ts
chesterkmr: Separated wf-service & controller to different modules to simplify re-using of wf-service logic in different modules. Instead of re-import everything in order to make wf-service run now we can import just single workflow module which exports service.

services/workflows-service/prisma/migrations/20240110184558_add_transition_schema_to_workflow_definition/migration.sql
chesterkmr: Transition schema will be used to validate context data between transitions or initial.(Only for creation currently).

apps/backoffice-v2/src/domains/workflows/hooks/queries/useWorkflowDefinitionQuery/useWorkflowDefinitionQuery.ts
Omri-Levy, line 1: The file's name should also be `useWorkflowDefinitionByIdQuery`.

apps/backoffice-v2/src/pages/Entities/components/CaseCreation/components/CaseCreationForm/hooks/useCaseCreationForm/utils/create-context-from-form-data.ts
```diff
@@ -0,0 +1,16 @@
+import { TWorkflowById } from '@/domains/workflows/fetchers';
+import { generateEntityId } from '@/pages/Entities/components/CaseCreation/components/CaseCreationForm/hooks/useCaseCreationForm/utils/generate-entity-id';
+import { AnyObject } from '@ballerine/ui';
+
+export const createContextFromFormData = (formData: AnyObject): TWorkflowById['context'] => {
+  const context = {
+    entity: {
+      id: generateEntityId(),
```
liorzam: In case you generate this id on client side we should take care for uniqueness on our backend

services/workflows-service/src/app.module.ts
```diff
@@ -1,7 +1,7 @@
 import { MiddlewareConsumer, Module } from '@nestjs/common';
 import { APP_GUARD, APP_INTERCEPTOR } from '@nestjs/core';
 import { UserModule } from './user/user.module';
-import { WorkflowModule } from './workflow/workflow.module';
+import { WorkflowControllerModule } from './workflow/workflow-controller.module';
```
liorzam: export controller from the original module, there is no reason for additional module just for the controllers

services/workflows-service/src/workflow/schemas/zod-schemas.ts
```diff
@@ -56,6 +56,10 @@ export const ConfigSchema = z
       .optional()
       .describe('Main representative of the company'),
     customerName: z.string().optional().describe('The customer (tenant) display name'),
+    isManualCreation: z
```
liorzam: instead a boolean why dont we just add a createdBy field?

services/workflows-service/src/workflow/workflow-controller.module.ts
```diff
@@ -0,0 +1,11 @@
...
+import { WorkflowControllerExternal } from './workflow.controller.external';
+import { WorkflowControllerInternal } from './workflow.controller.internal';
+import { WorkflowModule } from '@/workflow/workflow.module';
+
+@Module({
+  imports: [WorkflowModule],
+  controllers: [WorkflowControllerExternal, WorkflowControllerInternal],
+  exports: [WorkflowModule],
+})
+export class WorkflowControllerModule {}
```
liorzam: its redundant you can set 'controllers' in WorkflowModule

services/workflows-service/src/workflow/schemas/zod-schemas.ts
alonp99: Good point, we already have createdBy - it will help to add the agent who created it there for auditing.
But, we wont know its a user id without marking its a manual creation. @liorzam wdyt?

alonp99 (approved):

services/workflows-service/src/workflow/schemas/zod-schemas.ts
chesterkmr: @liorzam 
`isManualCreation` - Is set not after creation, its parameter which indicates if manual creation is available within backoffice.

Though assigning id of agent to`createBy` on manual creation is a good point.

apps/backoffice-v2/src/domains/workflows/fetchers.ts
```diff
@@ -54,28 +56,32 @@ export const fetchWorkflows = async (params: {
...
+  .nullable();
+
+export const WorkflowDefinitionSchema = ObjectWithIdSchema.extend({
+  name: z.string(),
+  version: z.number(),
+  variant: z.string().default(WorkflowDefinitionVariant.DEFAULT),
+  contextSchema: z.record(z.any(), z.any()).nullable(),
+  documentsSchema: z.array(z.any()).optional().nullable(),
+  config: WorkflowDefinitionConfigSchema,
+  definition: z.record(z.any(), z.any()),
```
Omri-Levy, line 73: `z.record(z.any(), z.any())` seems like a regression from what we had.

Omri-Levy (approved):

services/workflows-service/src/case-management/controllers/case-management.controller.ts
```diff
@@ -0,0 +1,48 @@
...
+    protected readonly caseManagementService: CaseManagementService,
+    protected readonly logger: AppLoggerService,
+  ) {}
+
+  @Get('workflow-definition/:workflowDefinitionId')
+  async getCaseDefinitionById(
+    @Param('workflowDefinitionId') workflowDefinitionId: string,
+    @ProjectIds() projectIds: TProjectId[],
+  ) {
+    return this.workflowDefinitionService.getLatestVersion(workflowDefinitionId, projectIds[0]!);
```
alonp99: It should use an arbitrary project id from an array, the correct logic is to try and fetch this wf -id from all of the allowed projectIds for this user

apps/backoffice-v2/src/pages/Entities/hooks/useEntities/useEntities.tsx
```diff
@@ -88,12 +89,15 @@ export const useEntities = () => {
...
 
+  const { workflowDefinition } = useCaseCreationWorkflowDefinition();
+
   return {
     onPaginate,
     onSearch: onSearchChange,
     onFilter: onFilterChange,
     onSortBy: onSortByChange,
     onSortDirToggle,
+    showCaseCreation: workflowDefinition?.config?.isManualCreation,
```
alonp99: dont forget to change this as well after changing the name

PR #1951, opened 2024-01-11
feat: added case management module & updated backoffice fetchers

chesterkmr:
### Description
- Added case management module
- Case creation endpoint
- validation of case input schema
- moved workflow service & dependencies to module.


services/workflows-service/src/case-management/case-management.service.ts
```diff
@@ -0,0 +1,76 @@
...
+    });
+
+    return {
+      workflowDefinitionId: actionResult[0]!.workflowDefinition.id,
+      workflowRuntimeId: actionResult[0]!.workflowRuntimeData.id,
+      ballerineEntityId: actionResult[0]!.ballerineEntityId,
+    };
+  }
+
+  private validateEntity(workflowDefinition: WorkflowDefinition, entity: unknown) {
```
alonp99, line 53: I don't think this method is specefic for the case management - its more around workflows

services/workflows-service/src/case-management/case-management.service.ts
```diff
@@ -0,0 +1,76 @@
...
+      workflowDefinition.transitionSchema as unknown as TransitionSchema[]
+    )?.find(schema => schema.state === inputState);
+
+    if (!transitionSchema || !transitionSchema.schema) return;
+
+    const validate = ajv.compile(transitionSchema.schema);
+
+    const isValid = validate(entity);
+
+    if (!isValid) {
```
alonp99, line 66: Is this properly handled in the front end? @Omri-Levy I know that mapping of error changes lately, @chesterkmr please sync with Omri about it, make sure it maps correctly to the input fields

services/workflows-service/src/case-management/case-management.service.ts
```diff
@@ -0,0 +1,76 @@
...
+      throw new BadRequestException({
+        message: 'Entity validation failed.',
+        errors: validate.errors?.map(({ instancePath, message, ...rest }) => ({
+          ...rest,
+          message: `${instancePath} ${message}`,
+          instancePath,
+        })),
+      });
```
Omri-Levy: Use `AjvValidationError`.

services/workflows-service/src/case-management/controllers/case-management.controller.ts
```diff
@@ -0,0 +1,47 @@
...
+export class CaseManagementController {
+  constructor(
+    protected readonly workflowDefinitionService: WorkflowDefinitionService,
+    protected readonly workflowService: WorkflowService,
+    protected readonly caseManagementService: CaseManagementService,
+    protected readonly logger: AppLoggerService,
+  ) {}
+
+  @Get('workflow-definition/:workflowDefinitionId')
+  async getCaseDefinition(
```
Omri-Levy:
```suggestion
  async getCaseDefinitionById(
```

services/workflows-service/src/case-management/controllers/case-management.controller.ts
```diff
@@ -0,0 +1,47 @@
...
+    return this.workflowDefinitionService.getLatestVersion(workflowDefinitionId, projectIds[0]!);
+  }
+
+  @Post()
+  @ApiOkResponse()
+  @HttpCode(200)
+  @ApiForbiddenResponse({ type: ForbiddenException })
+  async createCase(
+    @Body() body: WorkflowRunDto,
+    @UserData() user: { user: User },
```
Omri-Levy: `user` is of type `AuthenticatedEntity` as far as I know.

services/workflows-service/src/case-management/types/transition-schema.ts
```diff
@@ -0,0 +1,5 @@
+export interface TransitionSchema {
+  state: string;
+  schema: object;
+  additionalParameters?: object;
```
Omri-Levy: Prefer `Record<PropertyKey, unknown>` over `object`. or `string` instead of `PropertyKey` if its an object received/sent via network.

services/workflows-service/src/workflow/utils/ajv.validator.ts
Omri-Levy, line 1: Move to `src/common/ajv` or `src/lib/ajv`.

services/workflows-service/src/workflow/workflow.module.ts
Omri-Levy, line 1: Can we do something with having two workflow module files? It's not clear if its needed.

alonp99 (approved):

services/workflows-service/src/case-management/case-management.service.ts
chesterkmr: This method performs validation specificly against entity. Rest of validation is handled within createOrUpdate workflow.

PR #1966, opened 2024-01-16
Create collection flow URL endpoint

tomer-shvadron:
### Description
* Created a new endpoint that gets `workflowRuntimeDataId`, `endUserId` and `expiry` and returns a new token and the URL to the collection flow with that token
* Added entities object to the response of `/run` endpoint


services/workflows-service/src/workflow/dtos/create-collection-flow-url.ts
```diff
@@ -0,0 +1,26 @@
+import { ApiProperty } from '@nestjs/swagger';
+import { IsNotEmpty, IsNumber, IsOptional } from 'class-validator';
+
+export class CreateCollectionFlowUrlDto {
+  @ApiProperty({
+    required: true,
+  })
+  @IsNotEmpty()
```
Omri-Levy, line 9: Isn't `@IsString()` missing? If not, adding it just for clarity sake might be a good idea. Same case for `endUserId` @15

services/workflows-service/src/workflow/workflow.controller.external.ts
```diff
@@ -138,8 +142,9 @@ export class WorkflowControllerExternal {
     @ProjectIds() projectIds: TProjectIds,
     @CurrentProject() currentProjectId: TProjectId,
   ) {
-    // Rename to intent or getRunnableWorkflowDataByIntent?
+    // @TODO: Rename to intent or getRunnableWorkflowDataByIntent?
```
Omri-Levy: Can you please move the comment to line 147? 🙏🏼

services/workflows-service/src/workflow/workflow.controller.external.ts
```diff
@@ -190,6 +195,31 @@ export class WorkflowControllerExternal {
...
+  ) {
+    const expiresAt = new Date(Date.now() + (expiry || 30) * 24 * 60 * 60 * 1000);
+
+    const { token } = await this.workflowTokenService.create(currentProjectId, {
+      workflowRuntimeDataId: workflowRuntimeDataId,
+      expiresAt,
+      endUserId,
+    });
+
+    return res.json({
```
Omri-Levy: Note NestJS does not require `res.json` in most cases. Usually returning the object itself suffices.

Omri-Levy (approved):

PR #1990, opened 2024-01-21
add: automation shell script to deploy ballerine

pratapalakshmi:


alonp99 (approved):

deploy/scripts/boot.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+set -o errexit
+
```
liorzam, line 4: add some checks before running installation that our minimum requirements are matched 
OS, Arch, Available memory 

im not fan of sudo command, when use the boot.sh, the caller should provide relevant permission to run the script as administator you can add it to the requirements 
```
if (( $EUID != 0 )); then
    echo "Please run as root"
    exit
fi
```

deploy/scripts/boot.sh
```diff
@@ -0,0 +1,21 @@
...
+  sudo apt update;
+  sudo apt install -y docker.io
+  mkdir -p ~/.docker/cli-plugins/
+  curl -SL https://github.com/docker/compose/releases/download/v2.3.3/docker-compose-linux-x86_64 -o ~/.docker/cli-plugins/docker-compose
+  chmod +x ~/.docker/cli-plugins/docker-compose
+  sudo mv ~/.docker/cli-plugins/docker-compose /usr/bin/docker-compose
+}
+
+deploy_ballerine(){
+	mkdir -p /home/ubuntu/ballerine; cd /home/ubuntu/ballerine
```
liorzam: dont use hard coded path, prefer to use $HOME env var

alonp99 (approved):

alonp99 (approved):

PR #1994, opened 2024-01-23
Web-UI-SDK updates

tomer-shvadron:
### Description
* Updated `Web-UI-SDK` to use `workflows-service`


services/workflows-service/src/main.ts
```diff
@@ -27,6 +27,7 @@ const corsOrigins = [
   ...env.BACKOFFICE_CORS_ORIGIN,
   ...env.WORKFLOW_DASHBOARD_CORS_ORIGIN,
   ...env.KYB_EXAMPLE_CORS_ORIGIN,
+  'http://localhost:5202',
```
alonp99: add it as a const like all other cors configurations

alonp99 (approved):

PR #1996, opened 2024-01-23
add: automation tests for ballerine

pratapalakshmi:
Depends-on: [2007](https://github.com/ballerine-io/ballerine/pull/2007)


.github/workflows/test-ballerine-wf.yml
```diff
@@ -0,0 +1,45 @@
...
+          shell: bash
+    # Service containers to run with this job. Required for running tests
+    services:
+        # Label used to access the service container
+        postgres:
+          # Docker Hub image for Redis
+          image: sibedge/postgres-plv8:15.3-3.1.7
+          ports:
+            # Opens tcp port 5432 on the host and service container
+            - 5432:5432
```
liorzam, line 31: have you check the option to use dind?(Docker in docker?)

So, it should be easy to just run docker compose file and than the command for test job 
https://github.com/peter-evans/docker-compose-actions-workflow

liorzam:
Why dont just use docker compose and start relevant services as configured ?
https://gist.github.com/cecilemuller/437d7340b9f095cf5635dc9780a05092

pratapalakshmi:
> Why dont just use docker compose and start relevant services as configured ? https://gist.github.com/cecilemuller/437d7340b9f095cf5635dc9780a05092

That is what we are doing  here

pratapalakshmi:
please merge [this](https://github.com/ballerine-io/ballerine/pull/2007)
and this is the output of sample job
https://github.com/pratapalakshmi/ballerine/actions/runs/7685815163/job/20943974929

.github/workflows/test-ballerine-wf.yml
```diff
@@ -0,0 +1,57 @@
+name: Publish workflows-service image
```
alonp99: rename

alonp99 (approved):

.github/workflows/test-ballerine-wf.yml
pratapalakshmi: my bad

.github/workflows/test-ballerine-wf.yml
```diff
@@ -0,0 +1,57 @@
...
+
+on:
+  workflow_dispatch:
+  push:
+    paths:
+      # Run this pipeline only if there are changes in specified path
+      - "services/workflows-service/**"
+    branches:
```
liorzam, line 9: One we pushed it to dev and ran e2e test on this env,
The same result should be placed when running on others as well because we promote the image tag to another env

.github/workflows/test-ballerine-wf.yml
```diff
@@ -0,0 +1,57 @@
...
+        run: |
+           mkdir ballerine;
+           cd ballerine;
+           curl https://raw.githubusercontent.com/ballerine-io/ballerine/dev/deploy/docker-compose.yml -o docker-compose.yml;
+           docker-compose up -d  --force-recreate;
+           sleep 120
+
+      - name : Run health check if workflow service is running
+        run: |
+           ready_http_code=$(curl -o /dev/null -s -w "%{http_code}\n" http://localhost:3000/_health/ready)
```
liorzam, line 44: Add a retry mechanism which sleep 2 seconds in order to wait for warm up otherwise the service won't be ready right after you started the docker compose

pratapalakshmi: True that's why i kept a 2 minute sleep  [here](https://github.com/ballerine-io/ballerine/pull/1996/files#diff-8edd32e57210235974f774e1aea603f54d426f2f2a8c4d795860d0199b1b006cR40)

.github/workflows/test-ballerine-wf.yml
pratapalakshmi: Ohh ok got it then the changes should be a little different . I will address it

alonp99 (approved):

PR #2011, opened 2024-01-28
feat(backoffice-v2): created a component to handle uneditable inputs with long values

Omri-Levy:
### Description
Now breaks to multiple lines
image


apps/backoffice-v2/.eslintrc.cjs
```diff
@@ -6,6 +6,9 @@ module.exports = {
       callees: ['ctw'],
     },
   },
+  parserOptions: {
+    project: './tsconfig.json',
+  },
```
Omri-Levy, line 11: Fixes ESLint error.

apps/backoffice-v2/src/lib/blocks/components/EditableDetails/EditableDetails.tsx
```diff
@@ -54,14 +118,14 @@ export const EditableDetails: FunctionComponent<IEditableDetails> = ({
       return false;
     }
 
-    return isDecisionComponent && value && POSITIVE_VALUE_INDICATOR.includes(value.toLowerCase());
+    return isDecisionComponent && !!value && POSITIVE_VALUE_INDICATOR.includes(value.toLowerCase());
```
Omri-Levy, line 121: TypeScript says that before the return type was `boolean | ""`.

tomer-shvadron (approved):

PR #2028, opened 2024-01-30
bal 1282 tm subscriptions on customer v2

liorzam:
## Description
This pull request introduces a new API endpoint to enable the addition of subscriptions to a customer.
Additionally, updates have been made to the DB in order to support config column for customer.

## Changes Include:

* New Endpoint: Added a new POST /api/v1/external/customers/subscriptions endpoint in the relevant service to facilitate the creation of customer subscriptions config.
* schema.prisma: Customer model has been changed, added new optional Json column

* Performed unit tests to ensure the correct functioning of the new endpoint.
Conducted integration tests to verify the successful integration of subscription configurations into the customer's JSONB field. - BAL-1284

image

```sh
curl --location 'http://localhost:3001/api/v1/external/customers/subscriptions' \
--header 'Content-Type: application/json' \
--header 'Accept: application/json' \
--header 'Authorization: Bearer <SECRET>' \
--data '{
    "subscriptions": [
        {
            "url": "http://stam.com",
            "events": ["aaa"],
            "type": "webhook1"
        }
    ]
}'
```


alonp99 (approved):

services/workflows-service/src/common/filters/AllExceptions.filter.ts
```diff
@@ -34,11 +35,14 @@ export class AllExceptionsFilter extends BaseExceptionFilter {
       .setHeader('Content-Type', 'application/json')
       .json({
         errorCode: String(HttpStatusCode[serverError.getStatus()]),
-        message: serverError.message,
-        errors,
+        message:
+          typeof serverError.getResponse() === 'string'
+            ? serverError.getResponse()
+            : serverError.message,
```
Omri-Levy, line 41: I recall we came to an agreement `message` would be along the lines of `Bad Request` and `Validation Error` and `errors` would consist of the details.

services/workflows-service/src/common/types.ts
```diff
@@ -1,6 +1,17 @@
...
+
+export const SubscriptionSchema = z.discriminatedUnion('type', [
+  z
+    .object({
+      type: z.literal('webhook'),
+      url: z.string().url(),
+      events: z.array(z.string()),
+    })
+    .strict(),
+]);
```
Omri-Levy, line 17: Love it. 💪🏼

services/workflows-service/src/customer/customer.controller.external.intg.test.ts
```diff
@@ -0,0 +1,205 @@
...
+            "customerId": "${customerId}",
+            "id": "${projectId}",
+            "name": "Project ${projectId}",
+          },
+        ],
+        "subscriptions": null,
+        "websiteUrl": null,
+      }
+    `);
+    });
```
Omri-Levy: Why not expect an object?

services/workflows-service/src/errors.ts
```diff
@@ -25,24 +26,55 @@ export class SessionExpiredException extends common.UnauthorizedException {
...
+  @ApiProperty()
+  path!: string;
+}
+
+export class ValidationError extends common.BadRequestException {
   @ApiProperty()
   statusCode!: number;
   @ApiProperty()
-  message!: string;
+  message: string = 'Validation error';
```
Omri-Levy: Let's match the casing with "Bad Request" and the rest.
```suggestion
  message: string = 'Validation Error';
```

services/workflows-service/src/errors.ts
```diff
@@ -25,24 +26,55 @@ export class SessionExpiredException extends common.UnauthorizedException {
...
 
-  constructor(
-    public error: ErrorObject<string, Record<string, any>, unknown>[] | null | undefined,
-  ) {
-    super();
+  constructor(errors: { message: string; path: string }[]) {
+    super(
+      {
+        statusCode: common.HttpStatus.BAD_REQUEST,
+        message: errors,
```
Omri-Levy: Why are we passing errors to message?

services/workflows-service/src/errors.ts
```diff
@@ -25,24 +26,55 @@ export class SessionExpiredException extends common.UnauthorizedException {
...
+      path: instancePath,
+    }));
+
+    return new ValidationError(errors);
+  }
+
+  static fromZodError(error: ZodError) {
+    const errors = error.errors.map(zodIssue => ({
+      message: zodIssue.message,
+      path: zodIssue.path.join('.'), // Backwards compatibility - Legacy code message excepts array
```
Omri-Levy, line 75: Are we aware of what `path` can be? Wondering if joining with "." always outputs something like "user.details.email" and not something unexpected.

services/workflows-service/src/global.d.ts
```diff
@@ -2,6 +2,7 @@ declare module '@prisma/client' {
   import type {
     WorkflowRuntimeData as _WorkflowRuntimeData,
     WorkflowDefinition as _WorkflowDefinition,
+    Customer as _Customer,
```
Omri-Levy: `_` is our convention for unused. Until we vote on it in the past we've used a `_` suffix.
```suggestion
    Customer as Customer_,
```

services/workflows-service/src/global.d.ts
```diff
@@ -12,4 +13,8 @@ declare module '@prisma/client' {
   export type WorkflowDefinition = Omit<_WorkflowDefinition, 'config'> & {
     config: WorkflowConfig | any;
   };
+
+  export type Customer = Omit<_Customer, 'subscriptions'> & {
+    subscriptions: TCustomerSubscription | any;
```
Omri-Levy, line 18: A union with `any` just results in `any`.

services/workflows-service/src/common/filters/AllExceptions.filter.ts
liorzam: message always should present as it was until these days for backward compatibility which already used to pass errors to BadRequest and read a list of errors from message field,

services/workflows-service/src/customer/customer.controller.external.intg.test.ts
liorzam: easier way to get this object using inlineSnapshot

services/workflows-service/src/errors.ts
liorzam: a message should starts with capital letter and all the other should be lower

services/workflows-service/src/common/filters/AllExceptions.filter.ts
Omri-Levy: This has not been true for the backoffice for a while and in the collection flow error handling is limited and focused. Considering a change is needed in either of the apps making that change to handle `errors` instead of `message` makes more sense as I see it.

services/workflows-service/src/customer/customer.controller.external.intg.test.ts
Omri-Levy: Not sure I follow how its easier?
```suggestion
      expect(restCustomerFields).toMatchObject({
  country: 'GB',
  displayName: 'Customer ${customerId}',
  faviconImageUri: '',
  language: 'en',
  logoImageUri: '',
  name: 'Customer ${customerId}',
  projects: [
    {
      customerId: '${customerId}',
      id: '${projectId}',
      name: 'Project ${projectId}',
    },
  ],
  subscriptions: null,
  websiteUrl: null,
});
```

services/workflows-service/src/errors.ts
Omri-Levy: Is that true for a majority of messages in our system? Having in mind the conclusion of the errors thread and meeting we had.

services/workflows-service/src/customer/customer.controller.external.intg.test.ts
MatanYadaev: Using a real object instead of a string object is better IMO cuz you have autocompletion and have less room to mistake.

services/workflows-service/src/customer/customer.controller.external.intg.test.ts
```diff
@@ -0,0 +1,205 @@
...
+      [CustomerControllerExternal],
+      [PrismaModule, ClsModule],
+    );
+  });
+
+  describe('POST /subscriptions', () => {
+    let customerId: string;
+    let projectId: string;
+
+    beforeEach(async () => {
```
Omri-Levy, line 103: Why is there an `expect` in a `beforeEach` and the `beforeEach` is in a `describe`?

services/workflows-service/src/errors.ts
liorzam: changed and set . also for ajv

services/workflows-service/src/global.d.ts
liorzam: i cant tell, as you can see i follow the existing WorkflowDefinition

services/workflows-service/src/customer/customer.controller.external.intg.test.ts
liorzam: to make sure the client which should add customer did what he actually should do

PR #2049, opened 2024-02-06
Empty checkbox in filter by options

tomer-shvadron:
### Description
* Fixed empty checkbox in the status section in the filter by options

image


apps/backoffice-v2/src/pages/Entities/components/Cases/hooks/useCases/useCases.tsx
```diff
@@ -54,7 +54,7 @@ export const useCases = () => {
     () => [
       ...new Set(
         Object.keys(states || {})
-          .filter(key => Object.keys(states?.[key] || {}).length)
+          .filter(key => states?.[key]?.tags)
```
Omri-Levy: Do we need to check for empty arrays?

Omri-Levy (approved):

PR #2059, opened 2024-02-10
[docs] update ballerine installation on aws

pratapalakshmi:


apps/backoffice-v2/.env.example
```diff
@@ -1,4 +1,4 @@
-VITE_API_URL=http://localhost:3000/api/v1/internal
+VITE_API_URL=api/v1/internal
```
alonp99, line 1: It won't work on local environment since the API server is on port 3001

deploy/docker-compose-build.yml
```diff
@@ -21,9 +23,10 @@ services:
       - ballerine-workflow-service
     restart: on-failure
     environment:
-      VITE_API_URL: 'http://localhost:3000/api/v1/'
+      VITE_API_URL: 'http://${DOMAIN_NAME}:3000/api/v1/'
```
alonp99: if the domain name is undefined it should probably use just a path, in this case "/api/v1/"

deploy/docker-compose-build.yml
```diff
@@ -40,31 +43,32 @@ services:
...
       API_KEY: ${API_KEY}
       NODE_ENV: ${NODE_ENV}
       COMPOSE_PROJECT_NAME: ${COMPOSE_PROJECT_NAME}
       DB_PORT: ${DB_PORT}
       DB_USER: ${DB_USER}
       DB_PASSWORD: ${DB_PASSWORD}
       SESSION_SECRET: ${SESSION_SECRET}
-      BACKOFFICE_CORS_ORIGIN: http://localhost:${BACKOFFICE_PORT}
-      WORKFLOW_DASHBOARD_CORS_ORIGIN: http://localhost:${WORKFLOW_DASHBOARD_PORT}
+      BACKOFFICE_CORS_ORIGIN: http://${DOMAIN_NAME}:${BACKOFFICE_PORT}
```
alonp99: In this case if the domain name id undefined, it should probably fallback to localhost

apps/backoffice-v2/.env.example
pratapalakshmi: Ok, anything you can recommend here then ?

alonp99 (approved):

PR #2060, opened 2024-02-10
Fixed @ballerine/blocks types resolving to any

Omri-Levy:
### Description

1. At least temporarily resolved circular imports by moving `types.ts` into `blocks.ts`
2. Renamed the `Blocks` type to `BlocksClient` as it was a typo


packages/blocks/src/types.ts
Omri-Levy, line 1: Fix 1/2

packages/blocks/src/blocks.ts
```diff
@@ -1,19 +1,125 @@
...
+  ? TCell
+  : never;
+
+/**
+ * Infer the generic passed into the instance of `BlocksBuilder` registered by the consumer of `@ballerine/blocks`.
+ */
+export type Cells =
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
+  InferBlocksBuilderGeneric<BlocksClient['cells']>;
```
Omri-Levy: Fix 2/2. `BlocksClient` was previously `Blocks`, `Blocks` does not have the `cells` property registered by the package consumer.

packages/blocks/src/blocks.ts
```diff
@@ -1,19 +1,125 @@
...
+ * const cells: CellsMap = {
+ *    text: TextCell,
+ *    image: ImageCell,
+ *  }
+ */
+export type CellsMap = {
+  [TType in CellType]: FunctionComponent<ExtractCellProps<TType>>;
+};
+
+export type InferAllButLastArrayElements<T extends Array<any>> = T extends [...infer U, any]
```
MatanYadaev: Just a suggestion
```suggestion
export type ExcludeLastElement<T extends Array<any>> = T extends [...infer U, any]
```

packages/blocks/src/blocks.ts
```diff
@@ -1,19 +1,125 @@
...
+  ? U
+  : [];
+
+export type InferLastArrayElement<T extends Array<any>> = T extends [...any, infer U] ? U : never;
```
MatanYadaev: Maybe `get` is more intuitive than `infer`? Your choice :)
```suggestion
export type GetLastArrayElement<T extends Array<any>> = T extends [...any, infer U] ? U : never;
```

MatanYadaev (approved):
Beautiful work Omri

packages/blocks/src/blocks.ts
Omri-Levy: Following existing conventions from libraries I know like `tRPC` and Next.JS where `infer` is preferred.

packages/blocks/src/blocks.ts
Omri-Levy: Name could use some work but I'd like to avoid `Exclude` as its an existing TypeScript concept and could confuse that unions are expected.

PR #2066, opened 2024-02-11
bal 1379 assign alert bulk operation

liorzam:


services/workflows-service/prisma/schema.prisma
```diff
@@ -676,6 +676,7 @@ model Alert {
 
   assignee   User?   @relation(fields: [assigneeId], references: [id], onUpdate: Cascade, onDelete: NoAction)
   assigneeId String?
+  assigneeAt DateTime?
```
alonp99: assignedAt*

alonp99: assignedBy may also be vauleable

alonp99 (approved):

services/workflows-service/src/alert/alert.controller.external.ts
```diff
@@ -6,42 +6,107 @@ import { PrismaService } from '@/prisma/prisma.service';
...
 
-  @Get('/')
+  @common.Get('/')
   @swagger.ApiOkResponse({ type: Array<Object> }) // TODO: Set type
   @swagger.ApiNotFoundResponse({ type: errors.NotFoundException })
   @swagger.ApiForbiddenResponse({ type: errors.ForbiddenException })
-  @UsePipes(new ZodValidationPipe(FindAlertsSchema, 'query'))
-  async getAll(@Query() findAlertsDto: FindAlertsDto, @ProjectIds() projectIds: TProjectId[]) {
+  @common.UsePipes(new ZodValidationPipe(FindAlertsSchema, 'query'))
+  async getAll(
```
Omri-Levy: Please follow existing naming conventions, we're open to changing them but in an organized manner. Can be `listAlerts` or similar too.
```suggestion
  async list(
```

services/workflows-service/src/alert/alert.repository.ts
```diff
@@ -42,6 +42,33 @@ export class AlertRepository {
...
   }
 
+  // Method to update an alerts
+  async updateMany<T extends Omit<Prisma.AlertUpdateManyArgs, 'where'>>(
+    alertIds: string[],
+    projectId: TProjectId,
+    args: Prisma.SelectSubset<T, Omit<Prisma.AlertUpdateManyArgs, 'where'>>,
+  ): Promise<Alert[]> {
+    if (!projectId)
+      throw new Error('Project ID is required to perform an update opeartion on Alerts');
```
Omri-Levy: Should it be `Error` and not one of our custom classes?
```suggestion
    if (!projectId) {
      throw new Error('Project ID is required to perform an update opeartion on Alerts');
   };
```

Omri-Levy:
```suggestion
    if (!projectId)
      throw new Error('Project ID is required to perform an update operation on Alerts');
```

services/workflows-service/src/alert/alert.service.ts
```diff
@@ -20,6 +22,32 @@ export class AlertService {
...
+  ): Promise<Alert[]> {
+    try {
+      return await this.alertRepository.updateMany(alertIds, projectId, {
+        data: {
+          assigneeId: assigneeDto.assigneeId,
+        },
+      });
+    } catch (error: unknown) {
+      // Should be handled by ProjectAssigneeGuard on controller level
+      if (error instanceof Prisma.PrismaClientKnownRequestError) {
```
Omri-Levy: Replace with a method similar to `isRecordNotFoundError` found in `prisma.util.ts`.

services/workflows-service/src/alert/alert.service.ts
```diff
@@ -20,6 +22,32 @@ export class AlertService {
...
+    projectId: string,
+    assigneeDto: AlertAssigneeUniqueDto,
+  ): Promise<Alert[]> {
+    try {
+      return await this.alertRepository.updateMany(alertIds, projectId, {
+        data: {
+          assigneeId: assigneeDto.assigneeId,
+        },
+      });
+    } catch (error: unknown) {
```
Omri-Levy: `unknown` is supposed to be the default.
```suggestion
    } catch (error) {
```

services/workflows-service/src/alert/dtos/assign-alert.dto.ts
```diff
@@ -0,0 +1,81 @@
...
+  })
+  status!: string;
+}
+
+export class BulkAssignAlertsResponse {
+  @ApiProperty({
+    required: true,
+    enum: BulkStatus,
+  })
+  overallStatus!: TBulkStatus;
```
Omri-Levy, line 62: @alonp99 did we agree that the HTTP code can be used over `overallStatus`?

services/workflows-service/src/alert/guards/project-assignee.guard.ts
```diff
@@ -0,0 +1,60 @@
...
+      assigneeId = request.body.assigneeId;
+    } else if (request.query.assigneeId && typeof request.query.assigneeId === 'string') {
+      assigneeId = request.query.assigneeId;
+    } else if (request.params.assigneeId) {
+      assigneeId = request.params.assigneeId;
+    }
+
+    if (!assigneeId) {
+      this.logger.warn('Assignee id not found in request');
+      return false;
```
Omri-Levy: A line break like other places would be appreciated. 🙏🏼

services/workflows-service/src/alert/types.ts
```diff
@@ -5,3 +5,33 @@ export type TAuthenticationConfiguration = {
   isValid: boolean;
   webhookSharedSecret: string;
 };
+
+export const BulkStatus = {
+  success: 'success',
+  partial: 'partial',
+  failed: 'failed',
```
Omri-Levy: Let's keep at least the keys consistent with the rest of the system. 🙏🏼 
```suggestion
  SUCCESS: 'success',
  PARTIAL: 'partial',
  FAILED: 'failed',
```

services/workflows-service/src/alert/types.ts
```diff
@@ -5,3 +5,33 @@ export type TAuthenticationConfiguration = {
...
+  alertId: string;
+  status: 'success';
+};
+
+type FailedResponse = {
+  alertId: string;
+  status: 'failed';
+  error: {
+    message: string;
+  };
```
Omri-Levy: Don't think errors should be handled differently in different places.
```suggestion
  errors: {
    path?: string;
    message: string;
  }[];
```

services/workflows-service/src/case-management/case-management.service.ts
```diff
@@ -66,7 +66,7 @@ export class CaseManagementService {
     const isValid = validate(entity);
 
     if (!isValid) {
-      throw ValidationError.fromAjvError(validate.errors!);
+      throw BadValidationException.fromAjvError(validate.errors!);
```
Omri-Levy, line 69: Is the rename due to the naming conflict with class-validator?

services/workflows-service/src/errors.ts
```diff
@@ -75,6 +76,15 @@ export class ValidationError extends common.BadRequestException {
       path: zodIssue.path.join('.'), // Backwards compatibility - Legacy code message excepts array
     }));
 
-    return new ValidationError(errors);
+    return new BadValidationException(errors);
+  }
+
+  static fromClassValidator(error: ValidationError[]) {
```
Omri-Levy, line 82: Please consider checking out of dev for a moment and making unrelated changes on a separate PR. 🙏🏼

services/workflows-service/prisma/schema.prisma
liorzam: assignedBy currently is out of scope

services/workflows-service/src/alert/dtos/assign-alert.dto.ts
liorzam: this schema provided by @alonp99 
let me know what status code would you expect to ?

services/workflows-service/src/alert/types.ts
liorzam: its not a validation error, this kind of error handling by class-validator before reaching the Controller api

services/workflows-service/src/case-management/case-management.service.ts
liorzam: yeah, i could use 'as' which its confusing to use it and rename this class name in some places

services/workflows-service/src/errors.ts
liorzam: its related to my pr when i couldnt see validation errors, the api didnt work as expected

services/workflows-service/src/alert/types.ts
Omri-Levy: I understand where you're coming from but as a client it would be unpredictable and require more inside knowledge (even if from Swagger) if sometimes its `error` sometimes `errors` with different data structures.

services/workflows-service/src/errors.ts
Omri-Levy: You're fixing class validator's errors. Its affecting your PR but its unrelated.

services/workflows-service/src/alert/alert.service.ts
liorzam: done

services/workflows-service/src/alert/dtos/assign-alert.dto.ts
alonp99: We can use only status code or both status code and overallStatus

alonp99 (approved):

services/workflows-service/src/alert/alert.service.ts
```diff
@@ -20,6 +23,29 @@ export class AlertService {
...
+  ): Promise<Alert[]> {
+    try {
+      return await this.alertRepository.updateMany(alertIds, projectId, {
+        data: {
+          assigneeId: assigneeDto.assigneeId,
+        },
+      });
+    } catch (error) {
+      // Should be handled by ProjectAssigneeGuard on controller level
+      if (error instanceof Prisma.PrismaClientKnownRequestError) {
```
Omri-Levy: Looks like this condition is part of `isFkConstraintError`, can we remove this if?

services/workflows-service/src/alert/types.ts
liorzam: Please resolve and we will discuss about it later with alon

services/workflows-service/src/alert/dtos/assign-alert.dto.ts
liorzam: Can you explain what status code would you expect when overallStatus is `partial`?

Omri-Levy (approved):

PR #2099, opened 2024-02-19
Alerts drawer heading and summary

Omri-Levy:
- feat(*): the transactions drawer no longer uses hardcoded heading and summary
- feat(backoffice-v2): added counterparty id to search query params


services/workflows-service/src/alert-definition/alert-definition.repository.ts
```diff
@@ -0,0 +1,33 @@
...
+
+    queryArgs.where = {
+      ...queryArgs.where,
+      alert: {
+        some: {
+          id: alertId,
+        },
+      },
+      projectId: { in: projectIds! },
+    };
```
Omri-Levy: Need to refactor this and one more place that do not use the scope service.

services/workflows-service/src/alert/alert.controller.external.ts
```diff
@@ -201,4 +207,15 @@ export class AlertControllerExternal {
       }),
     };
   }
+
+  @common.Get('/:id/alert-definition')
```
Omri-Levy, line 211: @alonp99 @liorzam let me know if the path is fine with you.

alonp99: LGTM

alonp99 (approved):

PR #2132, opened 2024-02-26
Bal 1405 check alert def service

liorzam:


services/workflows-service/src/alert/alert.service.ts
```diff
@@ -103,33 +112,136 @@ export class AlertService {
...
   // Specific alert check logic based on the definition
-  private async checkAlert(definition: AlertDefinition): Promise<boolean> {
-    // ...
-    return true;
+  private async checkAlert(alertDefinition: AlertDefinition): Promise<boolean> {
+    let inlineRule: InlineRule;
+
+    const unknownData: unknown = alertDefinition.inlineRule;
+
+    inlineRule = unknownData as InlineRule;
```
Omri-Levy: Why are these re-assignments and type changes needed?

services/workflows-service/src/alert/alert.service.ts
```diff
@@ -103,33 +112,136 @@ export class AlertService {
...
+          alertResponse.rejected.push({
+            status: AlertExecutionStatus.FAILED,
+            error,
+            alertDefinition,
+            aggregatedRow,
+          });
+        }
+      });
+
+    await Promise.all(alertsPromises);
```
Omri-Levy, line 202: Does this need to be `allSettled`?

services/workflows-service/src/alert/consts.ts
```diff
@@ -0,0 +1,5 @@
+export const AlertExecutionStatus = {
+  SUCCEEDED: 'SUCCEEDED',
+  SKIPPED: 'SKIPPED',
+  FAILED: 'FAILED',
+};
```
Omri-Levy:
```suggestion
} as const;
```

services/workflows-service/src/data-analytics/types.ts
```diff
@@ -19,3 +25,11 @@ export type TransactionsAgainstDynamicRulesType = {
...
   groupByCounterparty?: boolean;
 };
+
+export type AnyArray = any[];
+
+export type AnyRecord = Record<PropertyKey, any>;
+
+export type GenericFunction = (...args: AnyArray) => any;
+
+export type GenericAsyncFunction = (...args: AnyArray) => Promise<any>;
```
Omri-Levy: Let's move these types to `workflows-service/src/types.ts` for now. I'll add a tech debt task to have some way to share these sort of types in our system.

Omri-Levy (approved):

services/workflows-service/src/alert/alert.service.ts
liorzam: nope the promise in handled since we have high level try..catch in the map function

services/workflows-service/src/alert/alert.service.ts
liorzam: i just got an ts issue about it
```
Conversion of type 'string | number | boolean | JsonObject | JsonArray | null' to type 'InlineRule' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
  Type 'JsonValue[]' is missing the following properties from type 'InlineRule': id, fnName, options, subjectsts(2352)
  ```

PR #2152, opened 2024-03-04
Fixes for AML block

tomer-shvadron:
### Description
Fixes for AML block and usages


apps/backoffice-v2/src/common/components/organisms/Header/Header.Navbar.tsx
```diff
@@ -23,13 +24,9 @@ export const Navbar: FunctionComponent = () => {
         const isActiveFilterGroup = checkIsActiveFilterGroup(navItem);
 
         return (
-          <>
+          <Fragment key={`${navItem.key}-${isActiveFilterGroup}`}>
```
tomer-shvadron, line 27: Boyscouting, fixed `missing key` warning

MatanYadaev: Great. Thanks for the comment, it definitely reduces cognitive load.

MatanYadaev (approved):

apps/backoffice-v2/src/lib/blocks/variants/OngoingBlocks/hooks/useOngoingBlocksLogic/useOngoingBlocksLogic.tsx
```diff
@@ -1,23 +1,23 @@
...
     filterId: filterId ?? '',
   });
-  const kycSessionKeys = Object.keys(workflow?.context?.pluginsOutput?.kyc_session ?? {});
-  const getAmlData = useCallback((key: string) => workflow?.context?.aml, [workflow?.context?.aml]);
-  const amlBlock = useAmlBlock({
-    sessionKeys: kycSessionKeys,
-    getAmlData,
-  });
+
+  const amlData = useMemo(() => workflow?.context?.aml, [workflow?.context?.aml]);
```
Omri-Levy: Redundant `useMemo`. Would make a bit more sense if it included the array wrapping.

apps/backoffice-v2/src/domains/workflows/fetchers.ts
```diff
@@ -65,6 +66,7 @@ export const BaseWorkflowByIdSchema = z.object({
   workflowDefinition: WorkflowDefinitionByIdSchema,
   createdAt: z.string().datetime(),
   context: z.object({
+    aml: AmlSchema.optional(),
```
Omri-Levy, line 69: Greatly appreciated! 🙏🏼

apps/backoffice-v2/src/lib/blocks/components/AmlBlock/utils/aml-adapter.ts
```diff
@@ -1,38 +1,34 @@
...
+  listingsRelatedToMatch: ListingRelatedToMatchSchema.optional().nullable(),
+});
+
+export const AmlSchema = z.object({
+  hits: z.array(HitSchema).optional().nullable(),
+  createdAt: z.string().optional().nullable(),
+  totalHits: z.number().optional().nullable(),
+});
+
+export const amlAdapter = (aml: z.infer<typeof AmlSchema>) => {
```
Omri-Levy: Note in most cases `output` is probably what intended as it considers transforms done to the data.

```suggestion
export type TAml = z.output<typeof AmlSchema>;

export const amlAdapter = (aml: TAml) => {
```

Omri-Levy: `input` fits in the context of forms for example where the schema is used at different stages of the data.

Omri-Levy (approved):

apps/backoffice-v2/src/lib/blocks/variants/OngoingBlocks/hooks/useOngoingBlocksLogic/useOngoingBlocksLogic.tsx
tomer-shvadron: You're right, I added the array into it.

PR #2187, opened 2024-03-11
bal 1589 rules after schema changes

liorzam:


services/workflows-service/src/prisma/prisma.service.ts
```diff
@@ -44,6 +44,10 @@ export class PrismaService extends PrismaClient implements OnModuleInit, OnModul
         },
       ],
     });
+
+    BigInt.prototype.toJSON = function () {
```
MatanYadaev, line 51: Maybe leave a comment here so it'll be clear why you did it?

services/workflows-service/scripts/alerts/generate-alerts.ts
```diff
@@ -91,7 +104,7 @@ export const ALERT_INLINE_RULES = [
         havingAggregate: AggregateType.COUNT,
 
         direction: 'inbound',
-        excludedCounterpartyIds: ['9999999999999999', '999999******9999'],
+        // excludedCounterparty: ['9999999999999999', '999999******9999'],
```
MatanYadaev, line 111: Consider deleting this line instead of commenting it.

services/workflows-service/scripts/alerts/generate-alerts.ts
```diff
@@ -119,7 +132,7 @@ export const ALERT_INLINE_RULES = [
...
```
MatanYadaev, line 141: Same

services/workflows-service/src/alert/alert.service.ts
```diff
@@ -16,6 +16,15 @@ import { AlertExecutionStatus } from './consts';
 
 // TODO: move to utils
 
+const findValNoCase = (obj: { [key: string]: unknown }, prop: string) => {
+  prop = (prop + '').toLowerCase();
+  for (let p in obj) {
+    if (p in obj && prop === `${p}`.toLowerCase()) {
```
MatanYadaev: Use `.hasOwnProperty`
```suggestion
  for (const p in obj) {
    if (obj.hasOwnProperty(p) && prop === p.toLowerCase()) {
```
Or use `Object.keys()`
```suggestion
  for (const p of Object.keys(obj)) {
    if (prop === p.toLowerCase()) {
```

services/workflows-service/src/alert/alert.service.ts
```diff
@@ -16,6 +16,15 @@ import { AlertExecutionStatus } from './consts';
 
 // TODO: move to utils
 
+const findValNoCase = (obj: { [key: string]: unknown }, prop: string) => {
```
MatanYadaev:
```suggestion
const findByKeyCaseInsensitive = (obj: { [key: string]: unknown }, prop: string) => {
```
```suggestion
const findByKeyNoCase = (obj: { [key: string]: unknown }, prop: string) => {
```

services/workflows-service/src/alert/alert.service.ts
```diff
@@ -152,10 +168,10 @@ export class AlertService {
       async (executionRow: Record<string, unknown>) => {
         try {
           if (
-            _.some(
-              inlineRule.subjects,
-              field => _.isNull(executionRow[field]) || _.isUndefined(executionRow[field]),
-            )
+            _.some(inlineRule.subjects, field => {
+              const val = findValNoCase(executionRow, field);
+              _.isNull(val) || _.isUndefined(val);
```
MatanYadaev: I think you missed a `return` statement.
```suggestion
              return _.isNull(val) || _.isUndefined(val);
```

services/workflows-service/src/alert/alert.service.ts
```diff
@@ -152,10 +168,10 @@ export class AlertService {
...
+            })
```
MatanYadaev: Executing multiline functions inside `if` statement is hard to read. Consider extracting it into a variable.

MatanYadaev (approved):

services/workflows-service/src/alert/alert.service.ts
liorzam: property can be also non string 
that's why i've used `${p}`

services/workflows-service/scripts/alerts/generate-alerts.ts
liorzam: we need to modify this exclusion from other rules.
so once i changed the type i had to take care for it cross all other rules

services/workflows-service/src/alert/alert.service.ts
liorzam: eslint rule
  7:13  error    Do not access Object.prototype method 'hasOwnProperty' from target object  no-prototype-builtins

PR #2191, opened 2024-03-12
Add alert labels

MatanYadaev:
https://github.com/ballerine-io/ballerine/assets/13586343/7b2a38d2-c494-403f-b3d4-c19cc1a9694b

When the text is too long:
CleanShot 2024-03-12 at 12 56 28

### DIFF
--- a/apps/backoffice-v2/src/domains/alerts/query-keys.ts
+++ b/apps/backoffice-v2/src/domains/alerts/query-keys.ts
@@ -1,5 +1,9 @@
 import { createQueryKeys } from '@lukemorales/query-key-factory';
-import { fetchAlertDefinitionByAlertId, fetchAlerts } from '@/domains/alerts/fetchers';
+import {
+  fetchAlertDefinitionByAlertId,
+  fetchAlertLabels,
+  fetchAlerts,
+} from '@/domains/alerts/fetchers';
 
 export const alertsQueryKeys = createQueryKeys('alerts', {
   list: ({ sortBy, sortDir, page, pageSize, ...params }) => {
@@ -35,4 +39,8 @@ export const alertsQueryKeys = createQueryKeys('alerts', {
       queryFn: () => fetchAlertDefinitionByAlertId({ alertId }),
     };
   },
+  alertLabels: () => ({
+    queryKey: ['alertLabels'],
+    queryFn: () => fetchAlertLabels(),
+  }),
 });
--- a/apps/backoffice-v2/src/domains/alerts/query-keys.ts
+++ b/apps/backoffice-v2/src/domains/alerts/query-keys.ts
@@ -40,7 +40,7 @@ export const alertsQueryKeys = createQueryKeys('alerts', {
     };
   },
   alertLabels: () => ({
-    queryKey: ['alertLabels'],
+    queryKey: [{}],
     queryFn: () => fetchAlertLabels(),
   }),
 });
--- a/apps/backoffice-v2/src/pages/TransactionMonitoringAlerts/components/AlertsTable/columns.tsx
+++ b/apps/backoffice-v2/src/pages/TransactionMonitoringAlerts/components/AlertsTable/columns.tsx
@@ -45,11 +45,15 @@ export const columns = [
     header: 'Date & Time',
   }),
   columnHelper.accessor('label', {
-    cell: info => (
-      <Badge variant="secondary" className="max-w-[8rem]" title={info.getValue()}>
-        <div className="truncate">{info.getValue()}</div>
+    cell: info => {
+    const label = info.getValue();
+    
+    return (
+      <Badge variant="secondary" className="max-w-[8rem]" title={label}>
+        <div className="truncate">{label}</div>
       </Badge>
-    ),
+    )
+    },
     header: 'Label',
   }),
   columnHelper.accessor('merchant.name', {
--- a/apps/backoffice-v2/src/pages/TransactionMonitoringAlerts/components/AlertsTable/columns.tsx
+++ b/apps/backoffice-v2/src/pages/TransactionMonitoringAlerts/components/AlertsTable/columns.tsx
@@ -45,7 +45,11 @@ export const columns = [
     header: 'Date & Time',
   }),
   columnHelper.accessor('label', {
-    cell: info => <Badge variant={'secondary'}>{info.getValue()}</Badge>,
+    cell: info => (
+      <Badge variant="secondary" className="max-w-[8rem]" title={info.getValue()}>
+        <div className="truncate">{info.getValue()}</div>
+      </Badge>
+    ),
     header: 'Label',
   }),
   columnHelper.accessor('merchant.name', {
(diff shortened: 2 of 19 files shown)


services/workflows-service/prisma/migrations/20240311133449_add_label_to_alert_definitions/migration.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE "AlertDefinition" ADD COLUMN "label" TEXT NOT NULL DEFAULT substr(md5(random()::text), 0, 10);
```
MatanYadaev, line 1: Added a default to not break local environments that already have alert definitions

apps/backoffice-v2/src/domains/alerts/query-keys.ts
```diff
@@ -35,4 +39,8 @@ export const alertsQueryKeys = createQueryKeys('alerts', {
       queryFn: () => fetchAlertDefinitionByAlertId({ alertId }),
     };
   },
+  alertLabels: () => ({
+    queryKey: ['alertLabels'],
```
Omri-Levy: Pretty sure the key of the method in this case `alertLabels` is already applied by the library.
```suggestion
    queryKey: [{}],
```

apps/backoffice-v2/src/pages/TransactionMonitoringAlerts/components/AlertsTable/columns.tsx
```diff
@@ -44,6 +44,10 @@ export const columns = [
     },
     header: 'Date & Time',
   }),
+  columnHelper.accessor('label', {
+    cell: info => <Badge variant={'secondary'}>{info.getValue()}</Badge>,
```
Omri-Levy: Consider if `valueOrNA` or the component wrapper is needed here. Can label be falsy? And how does it look like when invalid data or no data returns from the API.

apps/backoffice-v2/src/pages/TransactionMonitoringAlerts/hooks/useTransactionMonitoringAlertsLogic/useTransactionMonitoringAlertsLogic.tsx
```diff
@@ -29,6 +30,8 @@ export const useTransactionMonitoringAlertsLogic = () => {
         ?.sort((a, b) => (a?.id === session?.user?.id ? -1 : b?.id === session?.user?.id ? 1 : 0)),
     [assignees, session?.user?.id],
   );
+  const { data: labels } = useAlertLabelsQuery();
+  const sortedLabels = useMemo(() => labels?.sort(), [labels]);
```
Omri-Levy: `Array.prototype.sort` mutates.
```suggestion
  const sortedLabels = useMemo(() => labels?.slice()?.sort(), [labels]);
```

Omri-Levy: Also is there a need for the labels to be sorted?

apps/backoffice-v2/src/pages/TransactionMonitoringAlerts/components/AlertsTable/columns.tsx
MatanYadaev: `label` is always a string

apps/backoffice-v2/src/pages/TransactionMonitoringAlerts/hooks/useTransactionMonitoringAlertsLogic/useTransactionMonitoringAlertsLogic.tsx
MatanYadaev: There's no clear reason for sorting the labels. Just kept the logic we have with `assignees`, and it's kinda makes sense to sort the label, but not important.

services/workflows-service/scripts/alerts/generate-alerts.ts
```diff
@@ -133,7 +137,39 @@ export const ALERT_INLINE_RULES = [
       } as TransactionsAgainstDynamicRulesType,
     },
   },
-] as const satisfies readonly { inlineRule: InlineRule; defaultSeverity: AlertSeverity }[];
+] as const satisfies readonly {
+  inlineRule: InlineRule;
+  label: string;
+  defaultSeverity: AlertSeverity;
+}[];
```
Omri-Levy: `ReadonlyArray` might be more readable.

services/workflows-service/scripts/alerts/generate-alerts.ts
```diff
@@ -133,7 +137,39 @@ export const ALERT_INLINE_RULES = [
...
+
+function createData(
```
Omri-Levy: Does ESLint not error on this function in the IDE?

MatanYadaev: No error on my machine. But I'll change it to an arrow function :)

apps/backoffice-v2/src/pages/TransactionMonitoringAlerts/components/AlertsTable/columns.tsx
```diff
@@ -44,6 +44,14 @@ export const columns = [
     },
     header: 'Date & Time',
   }),
+  columnHelper.accessor('label', {
+    cell: info => (
+      <Badge variant="secondary" className="max-w-[8rem]" title={info.getValue()}>
+        <div className="truncate">{info.getValue()}</div>
+      </Badge>
+    ),
```
Omri-Levy:
```suggestion
    cell: info => {
    const label = info.getValue();
    
    return (
      <Badge variant="secondary" className="max-w-[8rem]" title={label}>
        <div className="truncate">{label}</div>
      </Badge>
    )
    },
```

Omri-Levy: Can't the `truncate` class work on the badge directly?

apps/backoffice-v2/src/pages/TransactionMonitoringAlerts/hooks/useTransactionMonitoringAlertsLogic/useTransactionMonitoringAlertsLogic.tsx
Omri-Levy: Note the assignees is sorted so the signed in user is always at the top.

apps/backoffice-v2/src/pages/TransactionMonitoringAlerts/components/AlertsTable/columns.tsx
MatanYadaev: When the `truncate` is on the `Badge` component it doesn't look good, it ignores the padding, etc.

Omri-Levy (approved):

PR #2281, opened 2024-04-03
Added: Navigation and Routing

### LINKED ISSUE #2263: Welcome - Navigation and Routing
Description

Add a "Home" page to the navigation bar, represented by the Home icon.
Integrate the new "Statistics" and "Cases" routes into the application's routing structure within the authenticated layout.
Modify authentication redirects to point to the "Statistics" page.
Code References (Example):

```
 // useNavbarLogic.tsx
 import { Home } from 'lucide-react';

 const navItems = [ 
     {
         label: 'Home',
         href: '/en/statistics',
         icon: Home 
     }, 
     // ...existing navItems
 ]; 

// Router.tsx

{
        element: <AuthenticatedLayout />,
        loader: authenticatedLayoutLoader,
        errorElement: <RouteError />,
        children: [
          {
            path: '/:locale',
            element: <Locale />,
            errorElement: <RouteError />,
            children: [
                // ... existing children
               {
                path: '/:locale/statistics',
                element: <Statistics />,
                errorElement: <RouteError />,
               },
              {
                path: '/:locale/workflows',
                element: <Workflows />,
                errorElement: <RouteError />,
               }
            ]
           }
        ]
}

```

```
// Providers.tsx

      <AuthProvider
        redirectAuthenticatedTo={'/en/statistics'}
        signInOptions={{
          callbackUrl: state?.from
            ? ...
            : '/en/statistics',
        }}
       // ...rest of props
      >
```

shashankvish0010:
### Description
Added a "Home" page to the navigation bar, represented by the 'Home' icon.
Integrated the "Statistics" page and "Workflows" page routes into the application's routing structure within the authenticated layout for now.
I modified the authentication redirect; now it redirects to statistics.

fixes: #2263 

 Tested with local development setup.

### Examples and references
Screenshot (319)

### DIFF
--- a/apps/backoffice-v2/src/Router/Router.tsx
+++ b/apps/backoffice-v2/src/Router/Router.tsx
@@ -21,6 +21,8 @@ import { NotFoundRedirect } from '@/pages/NotFound/NotFound';
 import { TransactionMonitoringAlerts } from '@/pages/TransactionMonitoringAlerts/TransactionMonitoringAlerts.page';
 import { TransactionMonitoring } from '@/pages/TransactionMonitoring/TransactionMonitoring';
 import { TransactionMonitoringAlertsAnalysisPage } from '@/pages/TransactionMonitoringAlertsAnalysis/TransactionMonitoringAlertsAnalysis.page';
+import { Statistics } from '@/pages/Statistics/Statistics.page'
+import { Workflows } from '@/pages/Workflows/Workflows.page'
 
 const router = createBrowserRouter([
   {
@@ -46,12 +48,12 @@ const router = createBrowserRouter([
             children: [
               ...(env.VITE_AUTH_ENABLED
                 ? [
-                    {
-                      path: '/:locale/auth/sign-in',
-                      element: <SignIn />,
-                      errorElement: <RouteError />,
-                    },
-                  ]
+                  {
+                    path: '/:locale/auth/sign-in',
+                    element: <SignIn />,
+                    errorElement: <RouteError />,
+                  },
+                ]
                 : []),
             ],
           },
@@ -106,6 +108,16 @@ const router = createBrowserRouter([
                     ],
                   },
                 ],
+              },
+              {
+                path: '/:locale/statistics',
+                element: <Statistics />,
+                errorElement: <RouteError />,
+              },
+              {
+                path: '/:locale/workflows',
+                element: <Workflows />,
+                errorElement: <RouteError />,
               },
             ],
           },
--- a/apps/backoffice-v2/src/Router/types.ts
+++ b/apps/backoffice-v2/src/Router/types.ts
@@ -10,6 +10,7 @@ export type TRouteWithoutChildren = {
 
 export type TRouteWithChildren = {
   text: ReactNode | ReactNode[];
+  href?: string;
   children: Array<Omit<TRouteWithoutChildren, 'icon'>>;
   key: string;
 };
--- a/apps/backoffice-v2/src/Router/types.ts
+++ b/apps/backoffice-v2/src/Router/types.ts
@@ -10,7 +10,6 @@ export type TRouteWithoutChildren = {
 
 export type TRouteWithChildren = {
   text: ReactNode | ReactNode[];
-  href?: string;
   children: Array<Omit<TRouteWithoutChildren, 'icon'>>;
   key: string;
 };
--- a/apps/backoffice-v2/src/common/components/organisms/Header/Header.Navbar.tsx
+++ b/apps/backoffice-v2/src/common/components/organisms/Header/Header.Navbar.tsx
@@ -59,11 +59,11 @@ export const Navbar: FunctionComponent = () => {
                             `gap-x-1 px-1.5 py-2 text-xs capitalize hover:bg-[#EBEEF9] hover:text-[#5E688E] active:bg-[#e0e4f6] [&:not([aria-current=page])]:text-[#8990AC]`,
                             childNavItem.filterId
                               ? {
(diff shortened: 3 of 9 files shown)


Omri-Levy:
Great work! Before merging try and see if you can understand why the nav item is not styled as the active item. Notice its a shade of gray instead of the expected black.

shashankvish0010:
> Great work! Before merging try and see if you can understand why the nav item is not styled as the active item. Notice its a shade of gray instead of the expected black.

Thanks for addressing the issue, I checked it and got to know that the 'Home' nav-item is not styled as active item when it's active, and added changes to make sure it works perfectly fine.

apps/backoffice-v2/src/Router/Router.tsx
```diff
@@ -46,12 +48,12 @@ const router = createBrowserRouter([
...
-                      element: <SignIn />,
-                      errorElement: <RouteError />,
-                    },
-                  ]
+                  {
+                    path: '/:locale/auth/sign-in',
+                    element: <SignIn />,
+                    errorElement: <RouteError />,
+                  },
+                ]
```
Omri-Levy, line 56: Please ensure your IDE is configured to run Prettier and ESLint on save. Install the respective plugins if you haven't and make sure you're not running a custom format on save. 🙏🏼

apps/backoffice-v2/src/Router/types.ts
```diff
@@ -10,6 +10,7 @@ export type TRouteWithoutChildren = {
 
 export type TRouteWithChildren = {
   text: ReactNode | ReactNode[];
+  href?: string;
```
Omri-Levy: Revert this change. Routes with children will never have an href. Also, having an optional href will leave room for mistakes.

apps/backoffice-v2/src/common/components/organisms/Header/Header.Navbar.tsx
```diff
@@ -86,7 +86,7 @@ export const Navbar: FunctionComponent = () => {
                   className={ctw(
                     `flex items-center gap-x-1 px-1.5 py-1 text-sm font-semibold capitalize text-[#8990AC] hover:bg-[#EBEEF9] hover:text-[#5E688E] active:bg-[#e0e4f6]`,
                     {
-                      'bg-white text-[#20232E]': navItem.filterId === filterId,
+                      'bg-white text-[#20232E]': navItem.filterId === filterId || (navItem.key === 'nav-item-Home' && navItem.href === window.location.pathname),
```
Omri-Levy: The fix should work for any item without children and not use the item's key. Note since we're in React and using React router here you'll use the pathname from `useLocation` and not from the window directly. 🙂

apps/backoffice-v2/src/common/components/organisms/Header/hooks/useNavbarLogic/useNavbarLogic.tsx
```diff
@@ -57,9 +63,13 @@ export const useNavbarLogic = () => {
...
-      return navItem.children?.some(
-        childNavItem => childNavItem.filterId === filterId || childNavItem.href === pathname,
-      );
+      if (navItem.children) {
+        return navItem.children?.some(
+          childNavItem => childNavItem.filterId === filterId || childNavItem.href === pathname,
+        );
+      } else {
+        return navItem.href === pathname;
+      }
```
Omri-Levy: Let's revert this considering the other suggested changes. 🙂  Also consider the following in the future for readability by early returning on the simpler condition (usually the negative).
```suggestion
     if (!navItem.children) {
        return navItem.href === pathname;
     }

     return navItem.children?.some(
       childNavItem => childNavItem.filterId === filterId || childNavItem.href === pathname,
     );
```

apps/backoffice-v2/src/common/components/organisms/Header/Header.Navbar.tsx
```diff
@@ -86,7 +86,7 @@ export const Navbar: FunctionComponent = () => {
...
                     },
                   )}
```
Omri-Levy:
```suggestion
                  className={({ isActive }) =>
                    ctw(
                      `flex items-center gap-x-1 p-2 text-sm font-semibold capitalize hover:bg-[#EBEEF9] hover:text-[#5E688E] active:bg-[#e0e4f6] [&:not([aria-current=page])]:text-[#8990AC]`,
                      {
                        'bg-white text-[#20232E]': isActive,
                      },
                    )
                  }
```

Omri-Levy: Please remove `className` on `TNavItemProps` to allow this change and update the `NavItem` component with the code below to allow this change. Thanks!
```
        className={props =>
          ctw(
            `flex gap-x-2 rounded-md`,
            className instanceof Function ? className(props) : className,
          )
        }
```

apps/backoffice-v2/src/common/components/organisms/Header/hooks/useNavbarLogic/useNavbarLogic.tsx
```diff
@@ -57,6 +63,10 @@ export const useNavbarLogic = () => {
   const { pathname } = useLocation();
   const checkIsActiveFilterGroup = useCallback(
     (navItem: TRouteWithChildren) => {
+      if (!navItem.children) {
+        return navItem.href === pathname;
+      }
```
Omri-Levy: Revert please. This should never be true since a filter group always has children. The suggestion in the previous comment was general advice for the future.

Issue #2265, opened 2024-03-31
Use ShadCN Tabs Component with React Router NavLink


alonp99:
What?
Integrate ShadCN's Tabs component in the Statistics page, with "Statistics" and "Workflows" tabs.

Why?
To enhance navigation within the Statistics page, allowing users to switch between Statistics and Workflows views.

How?

Utilize the Tabs component from ShadCN with asChild prop set to true on TabsTrigger.
For each tab trigger, use NavLink from react-router-dom to enable routing.
Create two tabs: "Statistics" and "Workflows".
Acceptance Criteria:

Tabs for "Statistics" and "Workflows" are functional and styled according to the design.
Navigating between tabs changes the view without reloading the page.

![Image](https://github.com/ballerine-io/ballerine/assets/61207713/2d418ceb-6674-4b8f-b8fc-c892b5f7f6fb)

Resources;

- https://ui.shadcn.com/docs/components/tabs
- https://reactrouter.com/en/main/components/nav-link

PR #2298, opened 2024-04-11
Added: ShadCN Tabs Component with React Router NavLink

shashankvish0010:
### Description
Integrated ShadCN's Tabs component in the Home component, with functional "Statistics" and "Workflows" tabs.

fixes: #2265 

### Breaking changes
 * Changes are applied to Home, as both the statistics and workflows are the child element of Home.

### Examples and references
Screenshot (324)
Screenshot (325)

### DIFF
--- a/apps/backoffice-v2/src/common/components/atoms/Home/Home.tsx
+++ b/apps/backoffice-v2/src/common/components/atoms/Home/Home.tsx
@@ -11,7 +11,6 @@ import { TabsContent } from '@/common/components/organisms/Tabs/Tabs.Content';
 export const Home: FunctionComponent = () => {
   const { data: session } = useAuthenticatedUserQuery();
   const { firstName, fullName, avatarUrl } = session?.user || {};
-  const isAuthenticated = useIsAuthenticated();
   const { locale } = useParams();
   const { pathname } = useLocation();
   const navigate = useNavigate();
--- a/apps/backoffice-v2/src/common/components/atoms/Home/Home.tsx
+++ b/apps/backoffice-v2/src/common/components/atoms/Home/Home.tsx
@@ -1,15 +1,20 @@
 import React, { FunctionComponent } from 'react';
-import { Outlet } from 'react-router-dom';
+import { NavLink, Outlet, useParams } from 'react-router-dom';
 import { t } from 'i18next';
 import { useAuthenticatedUserQuery } from '../../../../domains/auth/hooks/queries/useAuthenticatedUserQuery/useAuthenticatedUserQuery';
 import { UserAvatar } from '../UserAvatar/UserAvatar';
+import { Tabs } from '@/common/components/organisms/Tabs/Tabs';
+import { TabsList } from '@/common/components/organisms/Tabs/Tabs.List';
+import { TabsTrigger } from '@/common/components/organisms/Tabs/Tabs.Trigger';
+import { TabsContent } from '@/common/components/organisms/Tabs/Tabs.Content';
 
 export const Home: FunctionComponent = () => {
   const { data: session } = useAuthenticatedUserQuery();
   const { firstName, fullName, avatarUrl } = session?.user || {};
+  const { locale } = useParams();
 
   return (
-    <div className={`flex flex-col p-10`}>
+    <div className={`flex flex-col gap-10 p-10`}>
       <div className={`flex items-center`}>
         <UserAvatar
           fullName={fullName ?? ''}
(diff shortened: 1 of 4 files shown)


apps/backoffice-v2/src/common/components/atoms/Home/Home.tsx
```diff
@@ -21,7 +26,24 @@ export const Home: FunctionComponent = () => {
           {firstName && ` ${firstName}`}
         </h3>
       </div>
-      <Outlet />
+      <div>
+        <Tabs defaultValue="statistics">
```
Omri-Levy: The default value should be dependent on the URL. i.e. if the current route is `/:locale/workflows` then `defaultValue` should be `workflows`.

Omri-Levy:
Note the "Home" nav item should be active whenever either of the tabs are selected. Both "Statistics" and "Workflows" in addition to `/:locale`.

apps/backoffice-v2/src/common/components/atoms/Home/Home.tsx
```diff
@@ -1,15 +1,31 @@
...
   const { data: session } = useAuthenticatedUserQuery();
   const { firstName, fullName, avatarUrl } = session?.user || {};
+  const isAuthenticated = useIsAuthenticated();
+  const { locale } = useParams();
+  const { pathname } = useLocation();
+  const navigate = useNavigate();
+  const value = pathname.includes('workflows') ? 'workflows' : 'statistics';
+
+  useEffect(() => {
+    if (isAuthenticated && pathname === `/${locale}`) {
```
Omri-Levy: No need to handle authentication here. It is covered by the layout.

services/workflows-service/src/workflow/cron/ongoing-monitoring.cron.ts
Omri-Levy, line 1: Please make sure to not merge from dev unless you do the same for the epic branch first. 🙏🏼

apps/backoffice-v2/src/common/components/atoms/Home/Home.tsx
```diff
@@ -1,15 +1,29 @@
...
+  const { locale } = useParams();
+  const { pathname } = useLocation();
+  const navigate = useNavigate();
+  const value = pathname.includes('workflows') ? 'workflows' : 'statistics';
+
+  useEffect(() => {
+    if (pathname === `/${locale}`) {
+      navigate('/en/statistics');
+    }
+  }, [pathname]);
```
Omri-Levy: Early return on a falsy (!==, etc) tends to result in better readability. Once we decide to add more checks and logic doing so would also keep indentation levels low.
```suggestion
  useEffect(() => {
    if (pathname !== `/${locale}`) {
         return;
    }
    
    navigate('/en/statistics');
  }, [pathname]);
```

apps/backoffice-v2/src/common/components/atoms/Home/Home.tsx
```diff
@@ -1,15 +1,29 @@
...
   const { data: session } = useAuthenticatedUserQuery();
   const { firstName, fullName, avatarUrl } = session?.user || {};
+  const { locale } = useParams();
+  const { pathname } = useLocation();
+  const navigate = useNavigate();
+  const value = pathname.includes('workflows') ? 'workflows' : 'statistics';
+
+  useEffect(() => {
+    if (pathname === `/${locale}`) {
+      navigate('/en/statistics');
```
Omri-Levy:
```suggestion
      navigate(`/${locale}/statistics`);
```

apps/backoffice-v2/src/common/components/atoms/Home/Home.tsx
```diff
@@ -21,7 +35,21 @@ export const Home: FunctionComponent = () => {
...
+              <NavLink to={`/${locale}/statistics`}>Statistics</NavLink>
+            </TabsTrigger>
+            <TabsTrigger asChild={true} value="workflows">
+              <NavLink to={`/${locale}/workflows`}>Workflows</NavLink>
+            </TabsTrigger>
+          </TabsList>
+          <TabsContent value={value}>
+            <Outlet />
+          </TabsContent>
+        </Tabs>
```
Omri-Levy: What do you think?
```suggestion
        <Tabs defaultValue={value}>
          <TabsList>
            <TabsTrigger asChild={true} value={`/${locale}/statistics`}>
              <NavLink to={`/${locale}/statistics`}>Statistics</NavLink>
            </TabsTrigger>
            <TabsTrigger asChild={true} value={`/${locale}/workflows`}>
              <NavLink to={`/${locale}/workflows`}>Workflows</NavLink>
            </TabsTrigger>
          </TabsList>
          <TabsContent value={pathname}>
            <Outlet />
          </TabsContent>
        </Tabs>
```

apps/backoffice-v2/src/common/components/atoms/Home/Home.tsx
Omri-Levy: Note ESLint warns some items are missing from the dependency array.
```suggestion
  }, [locale, pathname, navigate]);
```

apps/backoffice-v2/src/common/components/atoms/Home/Home.tsx
shashankvish0010: I agree, as it's now more explicit and easier to understand.

tomer-shvadron (approved):

PR #2323, opened 2024-04-25
[feat] - Implemented sendgrid to internal/users endpoint 

Satyam-G5:
### Description
I have implemented SendGrid and configured it for the endpoint - "internal/users" , enabling the system to send emails to internal users efficiently. Introduced the sendWelcomeEmail option within the CreateOptions class, allowing for the handling of the sendMessage condition in the payload.

 * Emails were send successfully to the user email
   Screenshot 2024-04-25 184454

### Screenshots 
* added options with sendWelcomeEmail flag to payload 
Screenshot 2024-04-23 123445


services/workflows-service/src/user/user.controller.internal.ts
```diff
@@ -67,5 +69,29 @@ export class UserControllerInternal {
       },
       projectIds?.[0] || currentProjectId,
     );
+    if (shouldSendWelcomeEmail && createdUser.email) {
```
Omri-Levy: `options?.sendWelcomeEmail` is already false.
```suggestion
    if (options?.sendWelcomeEmail && createdUser.email) {
```

Omri-Levy: A space would help readability here. 🙂 
```suggestion
    );
    
    if (shouldSendWelcomeEmail && createdUser.email) {
```

services/workflows-service/src/user/user.controller.internal.ts
```diff
@@ -67,5 +69,29 @@ export class UserControllerInternal {
...
+      const message: sgMail.MailDataRequired = {
+        to: createdUser.email,
+        from: 'oss@ballerine.com',
+        subject: 'Welcome Message!',
+        text: 'Welcome to our Ballerine!',
+        html: '<strong>Welcome to our Ballerine!</strong>',
+      };
+
+      if (process.env.SENDGRID_API_KEY) {
```
Omri-Levy: See places that use `env.ts` over `process.env` and do the same here.

services/workflows-service/src/user/user.controller.internal.ts
```diff
@@ -67,5 +69,29 @@ export class UserControllerInternal {
...
+        sgMail.setApiKey(process.env.SENDGRID_API_KEY);
+        await sgMail.send(message).then(() => {}, error => {
+          console.error(error);
+          if (error.response) {
+            console.error(error.response.body)
+          }
+        });
```
Omri-Levy: Use try/catch here instead. 🙂

services/workflows-service/src/user/user.controller.internal.ts
Omri-Levy, line 1: Consider negating the if conditions to reduce nesting levels and indentations.
```
  @common.Post()
  @swagger.ApiCreatedResponse({ type: [UserModel] })
  @UseGuards(AdminAuthGuard)
  @swagger.ApiForbiddenResponse()
  async create(
    @common.Body() userCreatInfo: UserCreateDto,
  ) {
    const { projectIds, options, ...userInfo } = userCreatInfo;

    const createdUser = await this.service.create(
      {
        data: userInfo,
        select: {
          id: true,
          firstName: true,
          lastName: true,
          email: true,
          phone: true,
          roles: true,
          workflowRuntimeData: true,
        },
      },
      projectIds?.[0] || currentProjectId,
    );

    if (!options?.sendWelcomeEmail || !createdUser.email) {
      return createdUser;
    }

    const message: sgMail.MailDataRequired = {
      to: createdUser.email,
      from: 'oss@ballerine.com',
      subject: 'Welcome Message!',
      text: 'Welcome to our Ballerine!',
      html: '<strong>Welcome to our Ballerine!</strong>',
    };

    if (!process.env.SENDGRID_API_KEY) {
      console.warn('SendGrid API key not provided. Email will not be not send ');
      console.log('Email:', message);

      return createdUser;
    }

    try {
      sgMail.setApiKey(process.env.SENDGRID_API_KEY);

      await sgMail.send(message);
    } catch (error) {
      console.error(error);
    }

    return createdUser;
  }
```

Omri-Levy (approved):

services/workflows-service/src/user/user.controller.internal.ts
```diff
@@ -67,5 +69,33 @@ export class UserControllerInternal {
...
+    const message: sgMail.MailDataRequired = {
+      to: createdUser.email,
+      from: 'no-reply@ballerine.com',
+      subject: 'Welcome Message!',
+      text: 'Welcome to our Ballerine!',
+      html: '<strong>Welcome to our Ballerine!</strong>',
+    };
+
+    if (!env.SENDGRID_API_KEY) {
+      console.warn('SendGrid API key not provided. Email will not be not send ');
```
alonp99: Use logger service instead `console.log`

services/workflows-service/src/user/user.controller.internal.ts
```diff
@@ -67,5 +70,64 @@ export class UserControllerInternal {
...
+      this.logger.warn('SendGrid API key not provided. Email will not be not send ');
+      this.logger.log('Email:', message);
+
+      return createdUser;
+    }
+
+    try {
+      sgMail.setApiKey(env.SENDGRID_API_KEY);
+      await sgMail.send(message);
+    } catch (error: Error | any) {
```
Omri-Levy: If you union with `any` the type just becomes any. Error tends to be of type `unknown` since in JS even a string or a promise can be thrown, and when an error is thrown it may be of different types. Also, check out this instance I found in the codebase that seem to be closer to what you are trying to do here. (it was for alerts originally)

```
      this.logger.error(`Error sending email with SendGrid: ${isErrorWithMessage(error) ? error.message : ''}`, {
        error,
      });
```

And if you're familiar with type guards we have functions like `isErrorWithMessage`, or`isRecordNotFoundError`. These either validate the error has specific properties or that the error is of a specific instance.

```
export const PRISMA_RECORD_NOT_FOUND_ERROR = 'P2025';

export const isRecordNotFoundError = (
  error: unknown,
): error is Prisma.PrismaClientKnownRequestError & {
  code: typeof PRISMA_RECORD_NOT_FOUND_ERROR;
} =>
  error instanceof Prisma.PrismaClientKnownRequestError &&
  error.code === PRISMA_RECORD_NOT_FOUND_ERROR;
```

```
export interface IErrorWithMessage extends Error {
  message: string;
}

export const isErrorWithMessage = (error: unknown): error is IErrorWithMessage => {
  return isObject(error) && 'message' in error && typeof error.message === 'string';
};
```

Satyam-G5: Thanks for pointing out, using a union type with `any` can diminish the benefits of type checking. I'll change the error type to `unknown` and commit.

services/workflows-service/src/user/user.controller.internal.ts
```diff
@@ -67,5 +70,64 @@ export class UserControllerInternal {
...
+      this.logger.warn('SendGrid API key not provided. Email will not be not send ');
+      this.logger.log('Email:', message);
+
+      return createdUser;
+    }
+
+    try {
+      sgMail.setApiKey(env.SENDGRID_API_KEY);
+      await sgMail.send(message);
+    } catch (error: unknown) {
```
Omri-Levy: `error` should already be of type `unknown` for projects with a strict tsconfig. 🙂 
```suggestion
    } catch (error) {
```

Omri-Levy:
Can you please upload a screenshot of how the email looks after all the changes? Thanks in advance! 🙏🏼

Satyam-G5:
> Can you please upload a screenshot of how the email looks after all the changes? Thanks in advance! 🙏🏼

Here is the screenshot for the same  👍 

Screenshot 2024-05-04 035109

Issue #2270, opened 2024-03-31
Implement Date Range Input


alonp99:
- Add a date range input component using ShadCN's date picker component.
- When the date range is changed, update the `from` and `to` search query parameters.
- Ensure that the displayed statistics are updated based on the selected date range.

PR #2328, opened 2024-04-28
Implemented Date Range Input.

shashankvish0010:
### Description
Added a date range input component using shadcn/ui date picker component and Calendar component to select the dates.

### Related issues
Fixes: #2270 

### Breaking changes
 - The Home page is updated with the date range section, which allows the user to select two dates, 'from' & 'to'.
 - Once the user selects the date, the range is updated within the search parameters as well.


apps/backoffice-v2/src/common/hooks/useHomeLogic/useHomeLogic.ts
```diff
@@ -0,0 +1,21 @@
+import { useSearchParams } from 'react-router-dom';
+import { useZodSearchParams } from '@/common/hooks/useZodSearchParams/useZodSearchParams';
+import { HomeSearchSchema } from '@/pages/Home/home-search-schema';
+
+export const useHomeLogic = () => {
+  const [searchParams, setSearchParams] = useSearchParams();
+  const [dateRange, setDateRange] = useZodSearchParams(HomeSearchSchema);
```
Omri-Levy: Remember we pass `HomeSearchSchema` since it describes not only the date range.
```suggestion
  const [searchParams, setSearchParams] = useZodSearchParams(HomeSearchSchema);
```

apps/backoffice-v2/src/common/hooks/useHomeLogic/useHomeLogic.ts
```diff
@@ -0,0 +1,21 @@
...
+
+  const handleDateRangeChange = (range: any) => {
```
Omri-Levy: Instead of `any` you can do the following. Similarly you can access the type of the first argument of the `onChange` handler of the component.
```suggestion
  const handleDateRangeChange: ComponentProps<typeof DateRangePicker>['onChange'] = (range) => {
```

apps/backoffice-v2/src/common/hooks/useHomeLogic/useHomeLogic.ts
```diff
@@ -0,0 +1,21 @@
...
+    const from = range?.start?.toISOString() || '';
+    const to = range?.end?.toISOString() || '';
+
+    setSearchParams({ from, to });
+    setDateRange({ from, to });
```
Omri-Levy: After handling the previous comments this would be the next change.
```suggestion
    setSearchParams({ from, to });
```

apps/backoffice-v2/src/common/hooks/useHomeLogic/useHomeLogic.ts
```diff
@@ -0,0 +1,21 @@
...
+  };
+
+  return {
+    dateRange,
```
Omri-Levy:
```suggestion
    from,
    to
```

apps/backoffice-v2/src/pages/Home/Home.page.tsx
```diff
@@ -29,16 +29,25 @@ export const Home: FunctionComponent = () => {
...
+            className={`mr-2 d-6`}
+            avatarUrl={avatarUrl ?? undefined}
+          />
+          <h3 className={`flex max-w-[45ch] break-all text-2xl font-semibold`}>
+            {t(`home.greeting`)}
+            {firstName && ` ${firstName}`}
+          </h3>
+        </div>
+        <DateRangePicker
+          onChange={range => handleDateRangeChange(range)}
```
Omri-Levy:
```suggestion
          onChange={handleDateRangeChange}
```

apps/backoffice-v2/src/common/components/organisms/Calendar/Calendar.tsx
```diff
@@ -0,0 +1,62 @@
+import * as React from 'react';
+import { ChevronLeftIcon, ChevronRightIcon } from '@radix-ui/react-icons';
+import { ctw } from '@/common/utils/ctw/ctw';
+import { DayPicker } from 'react-day-picker';
+import { buttonVariants } from '../../atoms/Button/Button';
+
+export type CalendarProps = React.ComponentProps<typeof DayPicker>;
+
+function Calendar({ className, classNames, showOutsideDays = true, ...props }: CalendarProps) {
```
Omri-Levy: Make sure ESLint is properly configured for you. We have a rule that prefers arrow functions over function expressions.

Omri-Levy: Please take consistency into consideration. In a majority of our code this is how we use `export`. Same for any other pasted ShadCN component.
```suggestion
export function Calendar({ className, classNames, showOutsideDays = true, ...props }: CalendarProps) {
```

apps/backoffice-v2/src/common/components/molecules/DateRangePicker/DateRangePicker.tsx
```diff
@@ -0,0 +1,79 @@
...
+import { Popover, PopoverContent, PopoverTrigger } from '@ballerine/ui';
+
+type DateRangePickerProps = {
+  onChange: (range: { start: Date | null; end: Date | null }) => void;
+  value: { start: Date | null; end: Date | null };
+};
+
+export function DateRangePicker(
+  { onChange, value }: DateRangePickerProps,
+  { className }: React.HTMLAttributes<HTMLDivElement>,
```
Omri-Levy: React components don't receive two arguments. See comment above.

apps/backoffice-v2/src/common/components/molecules/DateRangePicker/DateRangePicker.tsx
```diff
@@ -0,0 +1,79 @@
...
+  return (
+    <div className={ctw('grid gap-2', className)}>
+      <Popover>
+        <PopoverTrigger asChild>
+          <Button
+            id="date"
+            variant={'outline'}
+            className={ctw(
+              'w-[300px] justify-start text-left font-normal',
+              !date && 'text-muted-foreground',
```
Omri-Levy: This syntax is part of `ctw`'s and `clsx`'s advantages. 🙂 
```suggestion
             { 'text-muted-foreground': !date },
```

apps/backoffice-v2/src/common/components/molecules/DateRangePicker/DateRangePicker.tsx
```diff
@@ -0,0 +1,79 @@
...
+              date.to ? (
+                <>
+                  {formatDate(date.from, 'LLL dd, y')} - {formatDate(date.to, 'LLL dd, y')}
+                </>
+              ) : (
+                formatDate(date.from, 'LLL dd, y')
+              )
+            ) : (
+              <span>Pick a date</span>
+            )}
```
Omri-Levy: Consider this alternative syntax. While more verbose, nested ternaris tend to be harder to read and follow. Consider how more conditions or states would look like. Also, when JSX starts to look like this something like `ts-pattern` could help. 🙂 
```suggestion
        {date?.from && date?.to && (
          formatDate(date.from, 'LLL dd, y') - formatDate(date.to, 'LLL dd, y')
        )}
        {date?.from && !date?.to && (
          formatDate(date.from, 'LLL dd, y')
        )}
        {!date?.from && !date?.to && (
          <span>Pick a date</span>
        )}
```

apps/backoffice-v2/src/common/components/molecules/DateRangePicker/DateRangePicker.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+  useEffect(() => {
+    if (date !== prevDateRef.current) {
+      onChange({
+        start: date?.from || null,
+        end: date?.to || null,
+      });
+      prevDateRef.current = date;
+    }
+  }, [date, onChange]);
```
Omri-Levy: `useEffect` should be a last resort. Here passing the date from outside as the `useState`'s initial value and using the `onChange` on the component itself would behave in a more robust way and more consistently. Check this if you haven't yet. https://react.dev/learn/you-might-not-need-an-effect

apps/backoffice-v2/src/common/components/molecules/DateRangePicker/DateRangePicker.tsx
```diff
@@ -0,0 +1,79 @@
...
+import { addMonths } from 'date-fns';
+import { ctw } from '@/common/utils/ctw/ctw';
+import { Button } from '../../atoms/Button/Button';
+import { Calendar } from '../../organisms/Calendar/Calendar';
+import { Popover, PopoverContent, PopoverTrigger } from '@ballerine/ui';
+
+type DateRangePickerProps = {
+  onChange: (range: { start: Date | null; end: Date | null }) => void;
+  value: { start: Date | null; end: Date | null };
+};
```
Omri-Levy: Could also utilize `Omit`, `Pick`, and `extends` with interfaces or type intersections to achieve this type.
```suggestion
type TDateRangePickerProps = {
  onChange: ComponentProps<typeof Calender>['onSelect'];
  value: ComponentProps<typeof Calender>['value'];
  className: ComponentProps<'div'>['className'];
};
```

apps/backoffice-v2/src/common/components/molecules/DateRangePicker/DateRangePicker.tsx
Omri-Levy: This is the expected end result. 🙂 
```suggestion
  { onChange, value, className }: DateRangePickerProps,
```

apps/backoffice-v2/src/common/components/molecules/DateRangePicker/DateRangePicker.tsx
Omri-Levy: Considering this logic is needed, we want access to the date from outside, and we utilize the URL as the initial value, let's make this component receive `value` and `onChange` from outside. 🙂

apps/backoffice-v2/src/common/components/molecules/DateRangePicker/DateRangePicker.tsx
```diff
@@ -0,0 +1,79 @@
+import React, { useEffect, useState, useRef } from 'react';
+import { CalendarIcon } from '@radix-ui/react-icons';
+import { formatDate } from '@/common/utils/format-date';
+import { DateRange } from 'react-day-picker';
+import { addMonths } from 'date-fns';
```
Omri-Levy: Considering we use `dayjs` see if we can achieve the same using `dayjs` and not `date-fns`. 🙏🏼

apps/backoffice-v2/src/common/components/molecules/DateRangePicker/DateRangePicker.tsx
Omri-Levy: Note the library we use here probably exports something like `DayPickerRangeProps`.

apps/backoffice-v2/src/common/components/molecules/DateRangePicker/DateRangePicker.tsx
```diff
@@ -0,0 +1,79 @@
...
+            ) : (
+              <span>Pick a date</span>
+            )}
+          </Button>
+        </PopoverTrigger>
+        <PopoverContent className="w-auto p-0" align="start">
+          <Calendar
+            initialFocus
+            mode="range"
+            defaultMonth={date?.from}
```
Omri-Levy: How are day and year handled? Do we need `defaultMonth` if we pass `selected` and `onSelect` and `selected` is from the URL?

apps/backoffice-v2/src/common/components/molecules/DateRangePicker/DateRangePicker.tsx
```diff
@@ -0,0 +1,79 @@
...
+            selected={date}
+            onSelect={setDate}
```
Omri-Levy: This is another alternative to the `useEffect`, also with `value` passed to the `useState` or directly to `selected`. 🙂 (no need to do this)
```suggestion
            onSelect={(selection) => {
               setDate(selection);
               onChange(selection);
            }}
```

apps/backoffice-v2/src/common/components/molecules/DateRangePicker/DateRangePicker.tsx
shashankvish0010: Yes, I understood that defaultMonth will display the initial month that should be displayed in the calendar; when it first renders. since we're passing the selected prop as the date state, which is initialized with the current date, the defaultMonth prop is not strictly necessary. The calendar will automatically display the current month.

apps/backoffice-v2/src/pages/Home/Home.page.tsx
```diff
@@ -29,16 +29,26 @@ export const Home: FunctionComponent = () => {
...
+            {firstName && ` ${firstName}`}
+          </h3>
+        </div>
+        <DateRangePicker
+          onChange={handleDateRangeChange}
+          value={{
+            start: from ? new Date(from) : null,
+            end: to ? new Date(to) : null,
+          }}
+          className={'<div>'}
```
Omri-Levy: Seems like a mistake.

apps/backoffice-v2/src/common/components/molecules/DateRangePicker/DateRangePicker.tsx
```diff
@@ -0,0 +1,64 @@
...
+    from: new Date(currentDate.getFullYear(), currentDate.getMonth(), currentDate.getDate()),
+    to: dayjs(new Date(currentDate.getFullYear(), currentDate.getMonth(), currentDate.getDate(), 1))
+      .add(1, 'month')
+      .toDate(),
+  });
+
+  const handleDateChange = (selection: DateRange | undefined) => {
+    setDate(selection);
+    onChange(selection);
+  };
```
Omri-Levy: I was expecting these lines to be removed and for the component to receive onChange and value as props. 🙂
```suggestion
```

apps/backoffice-v2/src/common/hooks/useHomeLogic/useHomeLogic.ts
```diff
@@ -0,0 +1,24 @@
...
+import { useZodSearchParams } from '@/common/hooks/useZodSearchParams/useZodSearchParams';
+import { HomeSearchSchema } from '@/pages/Home/home-search-schema';
+
+export const useHomeLogic = () => {
+  const [searchParams, setSearchParams] = useZodSearchParams(HomeSearchSchema);
+
+  const handleDateRangeChange: ComponentProps<typeof DateRangePicker>['onChange'] = (range: {
+    start: { toISOString: () => string };
+    end: { toISOString: () => string };
+  }) => {
```
Omri-Levy: ComponentProps<typeof DateRangePicker>['onChange'] gives you the type for `range`. If the type is wrong then we're doing something incorrectly.
```suggestion
  const handleDateRangeChange: ComponentProps<typeof DateRangePicker>['onChange'] = (range) => {
```

apps/backoffice-v2/src/common/components/molecules/DateRangePicker/DateRangePicker.tsx
shashankvish0010: Hey, what if we only receive the onChange as prop, not the value and className?

apps/backoffice-v2/src/common/components/molecules/DateRangePicker/DateRangePicker.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+type TDateRangePickerProps = {
+  onChange: ComponentProps<typeof Calendar>['onSelect'];
+  value: ComponentProps<typeof Calendar>['value'];
+  className: ComponentProps<'div'>['className'];
+};
+
+export const DateRangePicker = ({ onChange, value, className }: TDateRangePickerProps) => {
+  const [date, setDate] = useState<DateRange | undefined>();
+  console.log(value);
```
Omri-Levy:
```suggestion
```

apps/backoffice-v2/src/common/components/molecules/DateRangePicker/DateRangePicker.tsx
```diff
@@ -0,0 +1,56 @@
...
+            {date?.from && !date?.to && formatDate(date.from, 'LLL dd, y')}
+            {!date?.from && !date?.to && <span>Pick a date</span>}
+          </Button>
+        </PopoverTrigger>
+        <PopoverContent className="w-auto p-0" align="start">
+          <Calendar
+            initialFocus
+            mode="range"
+            selected={date}
+            onSelect={(selection: DateRange | undefined) => {
```
Omri-Levy: Why are you overriding the `onSelect` type here?

apps/backoffice-v2/src/common/components/molecules/DateRangePicker/DateRangePicker.tsx
```diff
@@ -0,0 +1,56 @@
...
+              setDate(selection);
+              onChange ?? onChange(selection);
```
Omri-Levy: You meant the opposite `&&`.
```suggestion
              onChange?.(selection);
```

apps/backoffice-v2/src/common/components/molecules/DateRangePicker/DateRangePicker.tsx
```diff
@@ -0,0 +1,56 @@
...
+import { DateRange } from 'react-day-picker';
+
+type TDateRangePickerProps = {
+  onChange: ComponentProps<typeof Calendar>['onSelect'];
+  value: ComponentProps<typeof Calendar>['value'];
+  className: ComponentProps<'div'>['className'];
+};
+
+export const DateRangePicker = ({ onChange, value, className }: TDateRangePickerProps) => {
+  const [date, setDate] = useState<DateRange | undefined>();
```
Omri-Levy: Why are we not using `value` over this state?

apps/backoffice-v2/src/common/components/molecules/DateRangePicker/DateRangePicker.tsx
shashankvish0010: I just used that for debugging, to check if the value is being updated or not. I should exclude that part, but I have made a stupid mistake. I will keep that in mind next time.

apps/backoffice-v2/src/common/components/molecules/DateRangePicker/DateRangePicker.tsx
shashankvish0010: The reason for not using the value prop is because the value is the initial value of the date range, whenever we select a date range in the Calendar component, the value prop is not updated automatically. Instead, we need to update a component state to reflect the new selected date range. Moreover, I tried to use the value prop but on debugging I addressed that the date range set to from and to within the value prop results in undefined just after selecting the date range, it may be an issue in development mode because of React strict mode ?

Omri-Levy: The `value` should update since `onChange` is passed. `onChange` should not be optional. If `value` was used as initial I'd expect it to be passed into the `useState`. 

If we were to make `value` optional (not saying we should) it would be fine to pass it into the `useState` as well since its already initialized with `undefined`.

 I don't understand why we are passing `value` and making it required in the interface but it doesn't do anything. I do see now its used for a conditional class.

If `value` is from the URL and `onChange` updates the URL I'm failing to understand why it wouldn't work. 🙂

shashankvish0010: Yes, I completely understand and agree with you, the issue with the `value`  is that it remains undefined even after selecting the date range, due to which `{!value?.from && !value?.to && <span>Pick a date</span>}` always remains active even after selecting the date range.

Screenshot (333)

And as the type of `onChange`  is `SelectRangeEventHandler | undefined` here, we cannot invoke an object that is possibly `undefined` thus need to use optional chaining.

Omri-Levy: Addressed in https://github.com/ballerine-io/ballerine/pull/2328/commits/38644b7d689519a1a247c0fb28b27e67a6ce187a

apps/backoffice-v2/src/common/hooks/useHomeLogic/useHomeLogic.ts
```diff
@@ -0,0 +1,49 @@
...
+import { useLocation, useNavigate } from 'react-router-dom';
+
+export const useHomeLogic = () => {
+  const locale = useLocale();
+  const { pathname } = useLocation();
+  const navigate = useNavigate();
+  const [{ from, to }, setSearchParams] = useZodSearchParams(HomeSearchSchema);
+  const { data: session } = useAuthenticatedUserQuery();
+  const { firstName, fullName, avatarUrl } = session?.user || {};
+  const queryParams = from || to ? `?from=${from}&to=${to}` : '';
```
Omri-Levy: This one is redundant. Overtime search query params will change, might as well pass the full `search` prop from `useLocation` like done in `Cases.Item.tsx`.

apps/backoffice-v2/src/pages/Home/Home.page.tsx
```diff
@@ -0,0 +1,62 @@
...
+        </div>
+        <DateRangePicker
+          onChange={handleDateRangeChange}
+          value={{ from: from ? new Date(from) : undefined, to: to ? new Date(to) : undefined }}
+        />
+      </div>
+      <div>
+        <Tabs defaultValue={pathname} key={pathname}>
+          <TabsList>
+            <TabsTrigger asChild={true} value={`/${locale}/home/statistics`}>
```
Omri-Levy: Is the active tab state affected by the search query param change? Do we need to also pass `search` into `value`?

shashankvish0010: Addressed in [6f4033d](https://github.com/ballerine-io/ballerine/pull/2328/commits/6f4033dbfd78b42a8835d9a4366a6222aea85fa6)

Omri-Levy (approved):

PR #2408, opened 2024-05-25
Reports view backend integration

Omri-Levy:


apps/backoffice-v2/src/common/components/organisms/DataTable/DataTable.tsx
```diff
@@ -104,7 +104,7 @@ export const DataTable = <TData extends RowData, TValue = any>({
         const newSortingState = sortingUpdaterOrValue(prevSortingState);
 
         onSort({
-          sortBy: newSortingState[0]?.id || sortBy,
+          sortBy: newSortingState[0]?.id?.replace(/_/g, '.') || sortBy,
```
Omri-Levy, line 107: Tanstack table converts columns that are period separated to snake_case. In the URL we want to represent objects so we transform back to a period.

services/workflows-service/prisma/migrations/20240526072522_business_report_status/migration.sql
```diff
@@ -0,0 +1,11 @@
+/*
+  Warnings:
+
+  - Added the required column `status` to the `BusinessReport` table without a default value. This is not possible if the table is not empty.
```
tomer-shvadron, line 4: Please check this thoroughly

tomer-shvadron (approved):

PR #2419, opened 2024-05-27
Fix "create customer" flow

MatanYadaev:
## Summary by CodeRabbit

- **New Features**
  - Automatic API key generation for customer authentication if not provided.

- **Dependency Updates**
  - Updated various dependencies to version `0.6.10` for improved stability and performance.

- **Improvements**
  - Optional fields `customerStatus` and `authenticationConfiguration` in customer creation for more flexible input handling.


services/workflows-service/src/customer/customer.controller.external.ts
```diff
@@ -37,8 +39,18 @@ export class CustomerControllerExternal {
       };
     }
 
+    const apiKey =
```
tomer-shvadron: Please consult with @liorzam around this area, he changed the use of API keys recently, and as far as I remember, `authenticationConfiguration` will be deprecated.

tomer-shvadron (approved):

services/workflows-service/src/customer/customer.controller.external.ts
```diff
@@ -37,8 +39,18 @@ export class CustomerControllerExternal {
...
+      customer.authenticationConfiguration?.authValue ??
+      faker.internet.password(API_KEY_LEN, false, PASSWORD_REGEX);
```
liorzam: use crypto in order to generate random api key

services/workflows-service/src/customer/customer.controller.external.ts
MatanYadaev: It's ok because it still uses this value to generate the new `ApiKey` model record. Thanks

services/workflows-service/src/customer/customer.controller.external.ts
```diff
@@ -37,8 +38,16 @@ export class CustomerControllerExternal {
       };
     }
 
+    const apiKey = customer.authenticationConfiguration?.authValue ?? randomUUID();
```
Omri-Levy, line 41: Isn't a falsy `authValue` something we should throw an error on?

Omri-Levy (approved):

services/workflows-service/src/customer/customer.controller.external.ts
MatanYadaev: No, the ideal is that the client don't provide the API key, but the server will generate a random one and return it.

PR #2421, opened 2024-05-27
bal 2106 - enable jest reports and data migration file path

liorzam:
Link to an report example: https://github.com/ballerine-io/ballerine/actions/runs/9287773639/artifacts/1548657493
image

Report example:
image

## Summary by CodeRabbit

- **New Features**
  - Added Jest HTML reporter for generating detailed test reports.
  - Introduced a new `fileName` column in the `DataMigrationVersion` table for better tracking of data migrations.

- **Improvements**
  - Enhanced logging in data migration scripts to include file names.
  - Updated `.gitignore` to exclude test report files.

- **Refactor**
  - Streamlined argument passing in alert generation scripts.
  - Updated CI workflow to use a specific version of the `test-action`.

- **Chores**
  - Added `jest-html-reporter` as a new dependency in `package.json`.
  - Set `LOG_LEVEL` to 'error' in database setup scripts.


services/workflows-service/src/test/db-setup.ts
```diff
@@ -4,6 +4,8 @@ import console from 'console';
 import { TestGlobal } from '@/test/test-global';
 import { execSync } from 'child_process';
 
+process.env.LOG_LEVEL = 'error';
```
liorzam, line 7: we would like to reduce logs output while running tests

MatanYadaev (approved):

PR #2423, opened 2024-05-28
chore: remove data when running setup again and print env vars for local env

liorzam:
## Summary by CodeRabbit

- **New Features**
  - Added a script to generate a bcrypt salt and set `HASHING_KEY_SECRET` based on the operating system.

- **Improvements**
  - Updated `.env.example` to enclose `HASHING_KEY_SECRET` in double quotes.
  - Added a new script to stop Docker database service before resetting the database.
  - Improved the structure of server environment variables for better maintainability.

- **Chores**
  - Updated subproject commit hash for data migrations.


services/workflows-service/package.json
```diff
@@ -5,7 +5,7 @@
   "description": "workflow-service",
   "scripts": {
     "spellcheck": "cspell \"*\"",
-    "setup": "npm run docker:db && npm run db:reset && npm run db:migrate-dev && npm run seed",
+    "setup": "npm run docker:db:down && npm run docker:db && npm run db:reset && npm run seed",
```
alonp99, line 8: if there is no docker runnning, docker down will fail the script?

liorzam: will work as expected , it just do nothing

PR #2425, opened 2024-05-28
Dispatch Event Plugin

tomer-shvadron:
## Summary by CodeRabbit

- **New Features**
  - Introduced `DispatchEventPlugin` for handling event dispatching within workflows, enhancing event processing and error handling.

- **Bug Fixes**
  - Improved array merging logic to better handle merging based on IDs.

- **Updates**
  - Updated dependencies to the latest versions for better performance and stability.
  - Enhanced event subscription and notification logic in the `WorkflowRunner` class.

- **Refactor**
  - Refactored internal functions and methods to improve code readability and maintainability.


services/workflows-service/src/workflow/workflow.service.ts
```diff
@@ -1928,6 +1935,101 @@ export class WorkflowService {
         },
       });
 
+      service.subscribe(WorkflowEvents.ENTITIES_UPDATE, async ({ payload }) => {
```
MatanYadaev: The code here must be in a different location. It makes the `createOrUpdateWorkflowRuntime` much less readable than it already is.

services/workflows-service/prisma/schema.prisma
```diff
@@ -95,11 +95,18 @@ model EndUser {
   @@index([projectId])
 }
 
+enum BusinessPosition {
```
MatanYadaev, line 98:
```suggestion
enum PositionInBusiness {
```
```suggestion
enum BusinessPosition {
```

services/workflows-service/prisma/schema.prisma
```diff
@@ -95,11 +95,18 @@ model EndUser {
...
+  director
+  authorized_signatory
+}
+
 model EndUsersOnBusinesses {
   endUser    EndUser  @relation(fields: [endUserId], references: [id])
   endUserId  String
   business   Business @relation(fields: [businessId], references: [id])
   businessId String
+  positionInBusiness BusinessPosition[]
```
MatanYadaev: Maybe `position` is a better name than `positionInBusiness`? It's already in the context of the relationship between an end user and a business, it makes sense to me. The same you call `user.email` and not `user.user_email`

services/workflows-service/prisma/schema.prisma
```diff
@@ -95,11 +95,18 @@ model EndUser {
   @@index([projectId])
 }
 
+enum BusinessPosition {
+  ubo
+  director
+  authorized_signatory
```
MatanYadaev, line 102: We also have `representative` option (I'll need it in my PR)
```suggestion
  authorized_signatory
  representative
```

MatanYadaev (approved):

services/workflows-service/prisma/schema.prisma
MatanYadaev: If you change the column name to `position`, then you can ignore this comment. I suggested it just to be consistent with the column name.

PR #2433, opened 2024-06-03
load env from aws secrets

liorzam:
`source ./scripts/load_env.sh`

## Summary by CodeRabbit

- **New Features**
  - Enhanced environment setup script to automatically fetch and export secrets from AWS Secrets Manager based on git remote URL keywords.
  - Implemented filtering logic to exclude specific keys from being exported as environment variables.

### DIFF
--- a/scripts/load_env.sh
+++ b/scripts/load_env.sh
@@ -45,12 +45,12 @@ fi
 echo "Secret fetched successfully. Setting environment variables...$skip_keys"
 
 # Use jq to parse the JSON and export the key-value pairs
-echo $secret_value | jq -r 'to_entries | .[] | "export \(.key)=\(.value)"' | while read -r line; do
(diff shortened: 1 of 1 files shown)


scripts/load_env.sh
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash
```
MatanYadaev, line 1: Why Bash and not JavaScript? Not everyone on the team is proficient with Bash, which makes this script less maintainable than it could be.
I understand you had to use AWS CLI, but you have use Node's `child_process`.

tomer-shvadron:
@liorzam Using bash will turn the experience of using Ballerine even more complex for Windows-based users. I'm having difficulties understanding your decision to go with bash script (opposed to [AWS SDK](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/SecretsManager.html)) and skipping secrets.

PR #2438, opened 2024-06-04
Bad this scope fix

tomer-shvadron:
## Summary by CodeRabbit

- **Refactor**
  - Improved accessibility of workflow context properties for better integration and flexibility.

- **Dependency Updates**
  - Upgraded `@ballerine/workflow-browser-sdk` and `@ballerine/workflow-node-sdk` to version `0.6.13` across multiple applications.
  - Updated versioning for `@ballerine/backoffice-v2`, `@ballerine/kyb-app`, `@ballerine/headless-example`, `@ballerine/workflow-core`, and `@ballerine/workflows-service`.

- **Bug Fixes**
  - Enhanced event subscription logic in `WorkflowService` for more reliable updates.

### DIFF
--- a/packages/workflow-core/src/lib/workflow-runner.ts
+++ b/packages/workflow-core/src/lib/workflow-runner.ts
@@ -69,7 +69,7 @@ export class WorkflowRunner {
   #__subscriptions: Partial<Record<string, Array<(event: WorkflowEvent) => Promise<void>>>>;
   #__workflow: StateMachine<any, any, any>;
   #__currentState: string | undefined | symbol | number | any;
-  #__context: any;
+  context: any;
   #__config: any;
   #__extensions: WorkflowExtensions;
   #__debugMode: boolean;
@@ -78,10 +78,6 @@ export class WorkflowRunner {
 
   public get workflow() {
     return this.#__workflow;
-  }
-
-  public get context() {
-    return this.#__context;
   }
 
   public get state() {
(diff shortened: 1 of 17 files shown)


packages/workflow-core/src/lib/workflow-runner.ts
```diff
@@ -69,7 +69,7 @@ export class WorkflowRunner {
   #__subscriptions: Partial<Record<string, Array<(event: WorkflowEvent) => Promise<void>>>>;
   #__workflow: StateMachine<any, any, any>;
   #__currentState: string | undefined | symbol | number | any;
-  #__context: any;
+  context: any;
```
Omri-Levy: Should still denote its private by convention. Also if you can check if `private` works better it will be appreciated.
```suggestion
  __context: any;
```

tomer-shvadron: I changed the `#__` because the IDE's debugger could not process and show it's value

services/workflows-service/src/workflow/workflow.service.ts
```diff
@@ -1963,7 +1962,7 @@ export class WorkflowService {
           endUserService: this.endUserService,
           projectId: currentProjectId,
           businessId: workflowRuntimeData.businessId,
-          sendEvent: service.sendEvent,
+          sendEvent: e => service.sendEvent(e),
```
Omri-Levy, line 1965:
```suggestion
          sendEvent: (event) => service.sendEvent(event),
```

Omri-Levy (approved):

services/workflows-service/src/workflow/workflow.service.ts
tomer-shvadron: There is already an `event` variable in the scope, that's why I picked `e`, as we just pass it along and it's common practice.

PR #2439, opened 2024-06-05
Fixes to Profiles Table

tomer-shvadron:
## Summary by CodeRabbit

- **New Features**
  - Enhanced role management by allowing multiple roles for individual profiles.
  - Added a new role "Representative" with appropriate display name in profiles table.

- **Improvements**
  - Updated role names to be displayed in lowercase for consistency.
  - Improved end-user data handling with additional business and role information.

### DIFF
--- a/services/workflows-service/src/case-management/controllers/case-management.controller.ts
+++ b/services/workflows-service/src/case-management/controllers/case-management.controller.ts
@@ -27,6 +27,7 @@ import { ZodValidationPipe } from '@/common/pipes/zod.pipe';
 import { ListIndividualsProfilesSchema } from '@/case-management/dtos/list-individuals-profiles.dto';
 import { z } from 'zod';
 import { EndUserAmlHitsSchema } from '@/end-user/end-user.schema';
+import { Business, EndUsersOnBusinesses } from '@prisma/client';
 
 @Controller('case-management')
 @ApiExcludeController()
@@ -118,19 +119,17 @@ export class CaseManagementController {
       },
       [projectId],
     );
+
     const typedEndUsers = endUsers as Array<
       (typeof endUsers)[number] & {
         endUsersOnBusinesses: Array<{
-          position: string[];
-          business: {
-            companyName: string;
-          };
+          position: EndUsersOnBusinesses['position'];
+          business: Pick<Business, 'companyName'>;
         }>;
-        businesses: Array<{
-          companyName: string;
-        }>;
+        businesses: Array<Pick<Business, 'companyName'>>;
       }
     >;
+
     const tagToKyc = {
       [StateTag.COLLECTION_FLOW]: 'PENDING',
       [StateTag.APPROVED]: 'APPROVED',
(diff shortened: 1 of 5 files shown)


services/workflows-service/src/case-management/controllers/case-management.controller.ts
```diff
@@ -87,6 +87,16 @@ export class CaseManagementController {
...
+              business: {
+                select: {
+                  companyName: true,
+                },
+              },
+            },
+          },
           businesses: {
             select: {
               companyName: true,
```
Omri-Levy, line 103: Is this still needed?

services/workflows-service/src/case-management/controllers/case-management.controller.ts
```diff
@@ -96,13 +106,26 @@ export class CaseManagementController {
...
       (typeof endUsers)[number] & {
+        endUsersOnBusinesses: Array<{
+          position: string[];
+          business: {
+            companyName: string;
+          };
+        }>;
         businesses: Array<{
           companyName: string;
         }>;
```
Omri-Levy: Do you mind using `Pick` on the Prisma types of `Business` and `EndUsersOnBusinesses`?

services/workflows-service/src/case-management/controllers/case-management.controller.ts
tomer-shvadron: I kept it as a backward compatibility measure in case we have other places that do not create a record in `EndUsersOnBusinesses`

services/workflows-service/src/case-management/controllers/case-management.controller.ts
```diff
@@ -138,12 +162,20 @@ export class CaseManagementController {
         const isMonitored = checkIsMonitored();
         const matches = getMatches();
 
+        const businesses = endUser.businesses?.length
+          ? endUser.businesses.map(business => business.companyName).join(', ')
+          : endUser.endUsersOnBusinesses
+              ?.map(endUserOnBusiness => endUserOnBusiness.business.companyName)
+              .join(', ');
```
Omri-Levy, line 168: I think using `endUsersOnBusinesses` here would suffice.

services/workflows-service/src/case-management/controllers/case-management.controller.ts
```diff
@@ -138,12 +162,20 @@ export class CaseManagementController {
...
+
         return {
           correlationId: endUser.correlationId,
           createdAt: endUser.createdAt,
           name: `${endUser.firstName} ${endUser.lastName}`,
-          businesses: endUser.businesses?.map(business => business.companyName).join(', '),
-          role: 'UBO',
+          businesses,
+          roles: endUser.endUsersOnBusinesses?.flatMap(
+            endUserOnBusiness => endUserOnBusiness.position,
```
Omri-Levy, line 176: Can `position` be `undefined` or an empty string?

Omri-Levy (approved):

services/workflows-service/src/case-management/controllers/case-management.controller.ts
tomer-shvadron: Not sure it would be backwards compatible.

services/workflows-service/src/case-management/controllers/case-management.controller.ts
tomer-shvadron: I didn't want to force the use of position as in some cases we might not have it

PR #2459, opened 2024-06-16
feature/ruleset_plugin - Design review - (WIP)

Blokh:
Snippet code for workflow ruleset

## Summary by CodeRabbit

- **New Features**
  - Introduced a new custom rule engine capable of evaluating complex rules.
  - Added support for evaluating JMESPath rules against context objects.
  - Added `RiskCalculationInput` type for enhanced risk calculation.

- **Enhancements**
  - Improved flexibility with new casting options and rule operations.
  - Enhanced rule violation formatting with grouping and criteria-based formatting.

- **Integration**
  - Integrated `NotionService` for fetching and transforming Notion database records.

- **Dependencies**
  - Updated and added several dependencies to support new features.


packages/rules-engine/src/lib/custom-rule-engine/consts/default-context-paths.ts
```diff
@@ -0,0 +1,3 @@
+export const DefaultContextPaths = {
+  ubos_default_path: 'entity.ubos',
```
MatanYadaev: Why snake_case and not camelCase?

packages/rules-engine/src/lib/custom-rule-engine/types/custom-rule/index.ts
```diff
@@ -0,0 +1,10 @@
+import { CustomRuleInput } from '@/lib/custom-rule-engine/types/custom-rule-input';
+import { DefaultContextPaths } from '@/lib/custom-rule-engine/consts/default-context-paths';
+
+export interface CustomRule {
+  (inputArguments: {
+    rule: CustomRuleInput;
+    context: Record<string, any>;
```
MatanYadaev, line 7: Best practices 🤷‍♂️
```suggestion
    context: Record<string, unknown>;
```

packages/rules-engine/src/lib/custom-rule-engine/types/custom-rule-input/index.ts
```diff
@@ -0,0 +1,9 @@
+export type CustomRuleInput = {
+  type: 'evaluate' | 'uboHasWarning';
+  options: {
+    fieldPath?: string;
+    cast?: 'String' | 'Number' | 'Boolean' | 'Date';
+    operations?: 'greater_than' | 'less_than' | 'equal' | 'not_equal' | 'contains' | 'not_contains';
```
MatanYadaev: snake_case -> camelCase?

packages/rules-engine/src/lib/custom-rule-engine/helpers/cast-compare-value.ts
```diff
@@ -0,0 +1,10 @@
+import { CustomRuleInput } from '@/lib/custom-rule-engine/types/custom-rule-input';
+import { castValue } from '@/lib/custom-rule-engine/helpers/cast-value';
+
+export const castCompareWithEval = (options: CustomRuleInput['options'], value: string) => {
+  if (value.startsWith('`') && value.endsWith('`')) {
+    return castValue(options, eval(value));
```
MatanYadaev: Please use safe-eval

packages/rules-engine/src/lib/custom-rule-engine/helpers/cast-compare-value.ts
```diff
@@ -0,0 +1,10 @@
+import { CustomRuleInput } from '@/lib/custom-rule-engine/types/custom-rule-input';
+import { castValue } from '@/lib/custom-rule-engine/helpers/cast-value';
+
+export const castCompareWithEval = (options: CustomRuleInput['options'], value: string) => {
+  if (value.startsWith('`') && value.endsWith('`')) {
```
MatanYadaev: I would prefer a different indicator for a JS code. `options.isJavascript` or something.

packages/rules-engine/src/lib/custom-rule-engine/helpers/cast-compare-value.ts
```diff
@@ -0,0 +1,10 @@
+import { CustomRuleInput } from '@/lib/custom-rule-engine/types/custom-rule-input';
+import { castValue } from '@/lib/custom-rule-engine/helpers/cast-value';
+
+export const castCompareWithEval = (options: CustomRuleInput['options'], value: string) => {
```
MatanYadaev, line 12: Consider using `try/catch`

packages/rules-engine/src/lib/custom-rule-engine/helpers/cast-value.ts
```diff
@@ -0,0 +1,16 @@
...
+export const castValue = (options: CustomRuleInput['options'], value: string) => {
+  switch (options.cast) {
+    case 'String':
+      return String(value);
+    case 'Number':
+      return Number(value);
+    case 'Boolean':
+      return Boolean(value);
+    case 'Date':
+      return new Date(value);
```
MatanYadaev: Need to validate that the date (for example, `new Date('invalid').getTime()` will output `NaN`)

packages/rules-engine/src/lib/custom-rule-engine/index.ts
```diff
@@ -0,0 +1,29 @@
...
+  context: Record<string, unknown>;
+  ruleset: CustomRuleSet;
+}) => {
+  const operationMethod = ruleset.operator === 'or' ? 'some' : 'every';
+
+  return ruleset.rules
+    .map(rule => {
+      return invokeRule(rule, context);
+    })
+    [operationMethod](Boolean);
```
MatanYadaev: This might make it a little more readable IMO:

```suggestion
  const invokedRules = ruleset.rules
    .map(rule => {
      return invokeRule(rule, context);
    });
    
  if (ruleset.operator === 'or') {
    return invokedRules.some(Boolean);
  }
  
  return invokedRules.every(Boolean);
```

packages/rules-engine/src/lib/custom-rule-engine/index.ts
```diff
@@ -0,0 +1,29 @@
...
+};
+
+const invokeRule = (rule: CustomRuleInput, context: Record<string, unknown>) => {
```
MatanYadaev, line 24: Add try/catch somewhere in this file?

packages/rules-engine/src/lib/custom-rule-engine/operations/evaluate.ts
```diff
@@ -0,0 +1,64 @@
+import { logger } from 'nx/src/utils/logger';
+import { CustomRule } from '@/lib/custom-rule-engine/types/custom-rule';
+import { castValue } from '@/lib/custom-rule-engine/helpers/cast-value';
+import { castCompareWithEval } from '../helpers/cast-compare-value';
+
+export const evaluate: CustomRule = ({ rule, context, fieldPath }) => {
+  const { options } = rule;
+  const { operations, compareValue: comparableValue } = options;
+  if (!fieldPath) {
+    logger.warn('Field path is not provided');
```
MatanYadaev: Why not throw an error?

packages/rules-engine/src/lib/custom-rule-engine/operations/evaluate.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    return false;
+  }
+  if (!comparableValue) {
```
MatanYadaev: What if `comparableValue` is `0`/`''`/`[]`?

packages/rules-engine/src/lib/custom-rule-engine/operations/evaluate.ts
MatanYadaev: Everywhere in this file, why do you prefer logs over errors?

packages/rules-engine/src/lib/custom-rule-engine/helpers/cast-compare-value.ts
alonp99: Im not sure about using eval here, is it really necessary - it will be hard to validate, still have security concerns even if you use safe-eval

packages/rules-engine/src/lib/custom-rule-engine/operations/evaluate.ts
```diff
@@ -0,0 +1,64 @@
...
+      if (Array.isArray(compareValue)) {
+        return !compareValue.includes(targetValue);
+      }
+
+      return !compareValue.includes(String(targetValue));
+    case 'greater_than':
+      return targetValue > compareValue;
+    case 'less_than':
+      return compareValue < targetValue;
+    case undefined:
```
MatanYadaev, line 57: Please explain 🙂

packages/rules-engine/src/lib/custom-rule-engine/operations/ubo-has-warning.ts
```diff
@@ -0,0 +1,9 @@
+import { CustomRule } from '@/lib/custom-rule-engine/types/custom-rule';
+import { DefaultContextPaths } from '@/lib/custom-rule-engine/consts/default-context-paths';
+
+export const uboHasWarning: CustomRule = ({
+  context,
+  fieldPath = DefaultContextPaths.ubos_default_path,
+}) => {
+  return !!context[fieldPath].any((ubo: { warnings: Array<unknown> }) => ubo.warnings.length > 0);
```
MatanYadaev: Don't you prefer assuming that `ubo` is `unknown`? You can use `lodash.get` or any other tool to ensure `warnings` really exists and is an array

packages/rules-engine/src/lib/custom-rule-engine/index.ts
Omri-Levy: @Blokh Remember `safeEvery`.

packages/rules-engine/src/lib/custom-rule-engine/operations/evaluate.ts
Omri-Levy: @MatanYadaev `satisfies never` catches that `undefined` is an unhandled case.

packages/rules-engine/src/lib/custom-rule-engine/helpers/cast-compare-value.ts
Blokh: any suggestions?
I am moving the validation to a method e.g.
```

const isJavascript = (value: string) => {
  return value.startsWith('`') && value.endsWith('`');
};

```

is that good enough?

packages/rules-engine/src/lib/custom-rule-engine/types/custom-rule-input/index.ts
Blokh: this one i disagree, as this functionality is arriving from the outside, it's more readable.
but we can discuss

packages/rules-engine/src/lib/custom-rule-engine/operations/evaluate.ts
alonp99: I think the at the lower levels (here) the output should be more informative and expose other metadata, for example where you would want to return "failed" status, or "error" status with a message (instead or including the log).

on status 'success', you will obviously add the results - true/false
Since its a library lets think of optimal conditions for the user

packages/rules-engine/src/lib/custom-rule-engine/types/custom-rule/index.ts
Blokh: I know any is a code smell. but this causes less flexibility and TS errors in usage of the context. every please where you try to extract some kind of information.

this is the only spot where any is used here.

packages/rules-engine/src/lib/custom-rule-engine/operations/evaluate.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    return false;
+  }
+  if (!comparableValue) {
+    logger.warn('Comparable value is not provided');
+
+    return false;
+  }
+
+  const fieldValue = context.entity[fieldPath];
```
alonp99: I don't think it should be aware of the entity, on a higher level it will be aware it is a "defaultWorkflowContext" but it's not the right layer to do so

packages/rules-engine/src/lib/custom-rule-engine/operations/evaluate.ts
```diff
@@ -0,0 +1,64 @@
...
+  const compareValue = castValue(options, fieldValue);
+  const targetValue = castCompareWithEval(options, comparableValue);
+
+  switch (operations) {
+    case 'equal':
```
alonp99, line 25: Idea: use functions from lodash, the usage will be looking at lodash docs and using a function from there - could cover a lot of use cases with well tested and mature codebase

packages/rules-engine/src/lib/custom-rule-engine/index.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  return ruleResults.every(Boolean);
+};
+
+const invokeRule = (rule: CustomRuleInput, context: Record<string, unknown>) => {
+  switch (rule.type) {
+    case 'evaluate':
+      return evaluate({ rule, context });
+    case 'uboHasWarning':
+      return uboHasWarning({ rule, context, ...rule.options });
```
Omri-Levy, line 29: Missing `default` - consider using `satisfies never` for the switch/case not using it now.

packages/rules-engine/src/lib/custom-rule-engine/operations/evaluate.ts
```diff
@@ -0,0 +1,64 @@
...
+
+        return false;
+      }
+
+      if (Array.isArray(compareValue)) {
+        return !compareValue.includes(targetValue);
+      }
+
+      return !compareValue.includes(String(targetValue));
+    case 'greater_than':
```
Omri-Levy, line 53: What about `greater_than_equals` and `less_than_equals`?

packages/rules-engine/src/lib/custom-rule-engine/helpers/cast-value.ts
```diff
@@ -0,0 +1,25 @@
...
+      return Boolean(value);
+    case 'Date':
+      const date = new Date(value);
+      if (isNaN(date.getTime())) {
+        throw new Error(`Invalid date value: ${value}`);
+      }
+
+      return date;
+    default:
+      return value;
```
Omri-Levy, line 23: `satisfies never`?

packages/rules-engine/src/lib/custom-rule-engine/index.ts
```diff
@@ -0,0 +1,33 @@
...
+    default:
+      rule.type satisfies never;
```
Omri-Levy, line 31: Remember to return something or throw an error.

packages/rules-engine/src/lib/custom-rule-engine/operations/ubo-has-warning.ts
```diff
@@ -0,0 +1,9 @@
+import { CustomRule } from '@/lib/custom-rule-engine/types/custom-rule';
+import { DefaultContextPath } from '@/lib/custom-rule-engine/consts/default-context-paths';
+import get from 'lodash.get';
+
+export const uboHasWarning: CustomRule = ({ context, fieldPath = DefaultContextPath.ubosPath }) => {
+  return !!get(context, fieldPath).some(
```
Omri-Levy, line 6: `Array.prototype.some` already return boolean without the `!!` no?

packages/rules-engine/src/lib/custom-rule-engine/types/custom-rule/index.ts
MatanYadaev: In this case, flexible means bugs :)
I think ensuring the context is `unknown` will require you to write a more robust code, using `lodash.get` or validating the data schema, etc.

packages/rules-engine/src/lib/custom-rule-engine/types/custom-rule-input/index.ts
MatanYadaev: The whole rule set definition is camelCase, why just the operation is snake_case? Why is it preferred?

packages/rules-engine/src/lib/custom-rule-engine/helpers/cast-compare-value.ts
MatanYadaev: No IMO.
But it's up to you to decide.

packages/rules-engine/src/lib/custom-rule-engine/index.ts
MatanYadaev: @Omri-Levy Why `safeEvery`?

packages/rules-engine/src/lib/custom-rule-engine/operations/evaluate.ts
MatanYadaev: That's not what I meant. I meant why `operation` can be `undefined`?

packages/rules-engine/package.json
```diff
@@ -26,23 +26,27 @@
     "node": ">=12"
   },
   "dependencies": {
+    "@types/lodash.get": "^4.4.9",
```
liorzam, line 29: move to dev dependency

packages/rules-engine/src/lib/custom-rule-engine/helpers/cast-value.ts
```diff
@@ -0,0 +1,25 @@
+import { CustomRuleInput } from '@/lib/custom-rule-engine/types/custom-rule-input';
+
+export const castValue = (options: CustomRuleInput['options'], value: string) => {
+  if (value === 'undefined') {
+    throw new Error(`Invalid value: ${value}`);
+  }
+
+  switch (options.cast) {
```
liorzam, line 8: please check typeof of value to ensure we have a concrete value:

```
> String(null)
'null'
```

```
> Number(null)
0

> Number([])
0
```

packages/rules-engine/src/lib/index.ts
```diff
@@ -0,0 +1,14 @@
+import { TRuleEngine } from '@/lib/types';
+import { CustomRuleEngine } from '@/lib/custom-rule-engine';
+import { JMESPathRuleEngine } from '@/lib/jmespath-rule-engine';
+
+export const createRuleEngine: TRuleEngine = ruleWithProvider => {
+  switch (ruleWithProvider.provider) {
+    case 'custom':
+      return CustomRuleEngine(ruleWithProvider);
+    case 'jmespath':
```
Omri-Levy, line 9: Should we have an enum for providers?

packages/rules-engine/src/lib/index.ts
```diff
@@ -0,0 +1,14 @@
...
+      return JMESPathRuleEngine(ruleWithProvider);
+    default:
+      throw new Error('Invalid provider');
```
Omri-Levy, line 12: `satisfies never`

packages/rules-engine/src/lib/jmespath-rule-engine/index.ts
```diff
@@ -0,0 +1,11 @@
+import jmespath from 'jmespath';
+
+export const JMESPathRuleEngine = ({
+  context,
+  rule,
+}: {
+  context: Record<string, unknown>;
+  rule: string;
+}) => {
+  return !!jmespath.search(context, rule);
```
Omri-Levy, line 10: This can still return empty arrays, empty objects, and so on.

packages/rules-engine/src/lib/types.ts
```diff
@@ -1,3 +1,11 @@
...
 
-export type TCreateRuleEngine = (options: RuleEngineOptions) => RuleEngine;
+export type TCustomerRuleEngineParams = { provider: 'custom' } & Parameters<
+  typeof CustomRuleEngine
+>[0];
+export type TJEMSPathRuleEngineParams = { provider: 'jmespath' } & Parameters<
+  typeof JMESPathRuleEngine
+>[0];
+
+export type TRuleEngine = (input: TCustomerRuleEngineParams | TJEMSPathRuleEngineParams) => boolean;
```
Omri-Levy, line 11: Externally consumed types should not have a prefix i.e. `I` or `T`.

packages/workflow-core/package.json
```diff
@@ -41,6 +41,7 @@
   "devDependencies": {
     "@babel/core": "7.17.9",
     "@babel/preset-env": "7.16.11",
+    "@ballerine/rules-engine-lib": "0.5.2",
```
Omri-Levy, line 44: @alonp99 I think we should remove the `-lib` bit from the package.

liorzam: agree

PR #2468, opened 2024-06-20
RiskRuleService with Notion

MatanYadaev:
## Summary by CodeRabbit

- **New Features**
  - Integrated Notion API with a new `NotionService` for retrieving and transforming database records.
  - Added `RiskRuleService` for handling and retrieving risk rule records.

- **Configuration**
  - Introduced `NOTION_API_KEY` to environment configuration files.

- **Dependencies**
  - Updated dependencies to include `@notionhq/client` and `string-ts`.
  - Removed `@nestjs/serve-static` dependency.


services/workflows-service/src/risk-rule.service.ts
```diff
@@ -0,0 +1,69 @@
...
+    'Additional risk score': z.number().min(0).max(100),
+    'Min risk score': z.number().min(0).max(100),
+    'Max risk score': z.number().min(0).max(100),
+  })
+  .transform(deepCamelKeys)
+  .array();
+
+type RiskRuleRecord = {
+  ruleSet: {
+    operator: 'and' | 'or';
```
liorzam: can be optional and set operator default value to 'and'

services/workflows-service/src/notion.service.ts
```diff
@@ -0,0 +1,113 @@
...
+
+@Injectable()
+export class NotionService {
+  private readonly client: Client;
+
+  constructor(
+    private readonly configService: ConfigService,
+    private readonly logger: AppLoggerService,
+  ) {
+    if (!env.NOTION_API_KEY) {
```
liorzam: Is it make sense to allow this variable as optional and once the service is initialized you will throw an exception?

liorzam: My solution is to create a provider that dynamically selects the appropriate service based on the configuration

something like:
```
export const dynamicServiceProvider: Provider = {
  provide: NotionService,
  useFactory: (configService: ConfigService) => {
    const key = configService.get<string>('NOTION_KEY');
    if (key) {
      return new NotionService();
    }
    return new MockNotionService();
  },
  inject: [ConfigService],
};
```

MockNotionService can throw error when a concrete service api is called

MatanYadaev: I prefer it to be required so in production if you will forget to set the env var it won't provision the pod. The alternative you suggested will notify only when a user tries to use the feature.

PR #2469, opened 2024-06-22
Bal 2254 rule engine

liorzam:


services/workflows-service/src/rule-engine/rule-store.service.ts
```diff
@@ -0,0 +1,14 @@
+import { Injectable } from '@nestjs/common';
+import { RuleSet } from './core/types';
+
+export type RuleStoreServiceFindAllOptions = {
+  databaseId: string;
+  source: 'notion';
+};
+
+@Injectable()
+export class RuleStoreService {
```
MatanYadaev, line 10: Maybe `RuleRepository`? Repository doesn't necessarily have to access your Postgres DB, it can access any DB/store, like Notion.

services/workflows-service/src/rule-engine/core/logger.ts
MatanYadaev, line 1: It can access the logger directly. Isn't it a premature optimization?

services/workflows-service/src/rule-engine/core/index.ts
MatanYadaev, line 1: I think this file isn't in use.

services/workflows-service/src/rule-engine/core/condition/helpers.ts
```diff
@@ -0,0 +1,69 @@
...
+): boolean => {
+  return GTE(dataValue, conditionValue.min) && LTE(dataValue, conditionValue.max);
+};
+
+export const LAST_YEAR: ConditionFn<LastYearsParams> = (
+  dataValue: unknown,
+  conditionValue: LastYearsParams,
+): boolean => {
+  if (typeof dataValue === 'string' || dataValue instanceof Date) {
+    const date = new Date(dataValue);
```
MatanYadaev, line 62: Can you please add a validation to the date? https://www.freecodecamp.org/news/how-to-validate-a-date-in-javascript/

services/workflows-service/src/rule-engine/rule-store.service.ts
liorzam: Oh that just for having additional 'store'/'repository' its doesnt matter at all...
We might remove this class after merge to notion in order to fetch all relevant rules

services/workflows-service/src/rule-engine/core/condition/helpers.ts
```diff
@@ -0,0 +1,69 @@
...
+    const oneYearAgo = new Date();
+    oneYearAgo.setFullYear(oneYearAgo.getFullYear() - conditionValue.years);
+    return date >= oneYearAgo;
```
MatanYadaev, line 65: Do you want to filter only things from last year until now?
```suggestion
    return date >= oneYearAgo && date <= new Date()
```

services/workflows-service/src/rule-engine/core/logger.ts
liorzam: i tried to be persist as we did on workflow core in order to inject logger from outside

services/workflows-service/src/rule-engine/core/index.ts
liorzam: because i havnt called to setLogger yet...

services/workflows-service/src/rule-engine/core/condition/helpers.ts
liorzam: as part of my open items we need to add validation for every operator

services/workflows-service/src/rule-engine/core/condition/enums.ts
```diff
@@ -0,0 +1,14 @@
...
+  LT = 'LT',
+  GTE = 'GTE',
+  LTE = 'LTE',
+  LAST_YEAR = 'LAST_YEAR',
+}
+
+export enum ConditionOperator {
+  AND = 'and',
+  OR = 'or',
+}
```
MatanYadaev, line 14: Enums in TS are considered a bad practice. You can use a simple object instead. https://www.youtube.com/watch?v=jjMbPt_H3RQ
You can check how we implemented `CaseState` as a good example.

services/workflows-service/src/rule-engine/core/condition/constants.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export const conditionHelpers = {
+  [ConditionType.EQUALS]: EQUALS,
+  [ConditionType.BETWEEN]: BETWEEN,
+  [ConditionType.GT]: GT,
+  [ConditionType.GTE]: GTE,
+  [ConditionType.LT]: LT,
+  [ConditionType.LTE]: LTE,
+  [ConditionType.LAST_YEAR]: LAST_YEAR,
+} as const;
```
MatanYadaev, line 13: I'm not sure why the enum is necessary here. You can just do this:

```suggestion
import type { ValueOf } from 'type-fest';

export const conditionHelpers = {
  EQUALS,
  BETWEEN,
  GT,
  GTE,
  LT,
  LTE,
  LAST_YEAR,
} as const;

export type ConditionHelper = ValueOf<typeof conditionHelpers>;
```

services/workflows-service/src/rule-engine/core/condition/helpers.ts
liorzam: The rule checks if this date is within the last 2 years (or another number of years, depending on the parameter you set).

If you also need to validate dates in the future its another use case because we can allow to use this rule for future dates as well

services/workflows-service/src/rule-engine/core/engine.intg.test.ts
```diff
@@ -0,0 +1,226 @@
+import { Test, TestingModule } from '@nestjs/testing';
+import { RuleStoreService, RuleStoreServiceFindAllOptions } from '../rule-store.service';
+import { RuleEngineService } from '../rule-engine.service';
+import { RuleEngineModule } from '../rule-engine.module';
+import { ConditionOperator, ConditionType } from './condition/enums';
+import { RuleSet } from './types';
+
+const rules: RuleSet = [
```
MatanYadaev, line 8: Looks great 🔥

services/workflows-service/src/rule-engine/core/condition/constants.ts
liorzam: Module '"type-fest"' has no exported member 'ValueOf'

> I'm not sure why the enum is necessary here. You can just do this:

MatanYadaev: We use an old version of `type-fest`...

You can use this:
```ts
export type ValueOf<ObjectType, ValueType extends keyof ObjectType = keyof ObjectType> = ObjectType[ValueType];
```

PR #2476, opened 2024-06-25
feat: use run time variables in dist in wf-dashboard

pratapalakshmi:
## Release Notes

- **New Features**
	- Introduced a new `entrypoint.sh` script to configure environment variables for the web application.
	- Added a global configuration object `env` for centralized management of environment settings.
	- Included a script tag in the HTML to load the new configuration file.

- **Improvements**
	- Enhanced the production container initialization process for better custom entrypoint behavior.
	- Updated the build process to include a cleanup step before building.

- **Bug Fixes**
	- Adjusted the Axios instance configuration to utilize the new global environment variables. 

- **Documentation**
	- Added new environment variable declarations to the `.env` file and Docker Compose configuration.


apps/workflows-dashboard/entrypoint.sh
```diff
@@ -0,0 +1,35 @@
...
+fi
+
+
+cat << EOF > /usr/share/nginx/html/config.js
+globalThis.env = {
+  VITE_API_URL: "http://$VITE_DOMAIN/api/v1/",
+  VITE_ENVIRONMENT_NAME: "$VITE_ENVIRONMENT_NAME",
+  MODE: "$MODE",
+  VITE_IMAGE_LOGO_URL: "$VITE_IMAGE_LOGO_URL",
+}
```
liorzam: Why do you need the file `config.js` if you are adding the same file content inline in here?

liorzam: i just say that you can delete config.js file

apps/workflows-dashboard/src/main.tsx
```diff
@@ -4,6 +4,7 @@ import * as ReactDOM from 'react-dom/client';
 import { RouterProvider } from 'react-router-dom';
 import './index.css';
 import { initializeMonitoring } from '@/initialize-monitoring/initialize-monitoring';
+import '../public/config.js?url';
```
liorzam: if you add it in here it would take in place just if you build the app again

apps/workflows-dashboard/index.html
```diff
@@ -6,6 +6,7 @@
     <link rel="icon" type="image/png" sizes="16x16" href="/public/favicon-16x16.png" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <title>Workflow Dashboard</title>
+    <script type="text/javascript" src="/config.js"></script>
```
liorzam, line 9: the path should be relative im not sure its actually working

liorzam (dismissed):

apps/workflows-dashboard/public/config.js
```diff
@@ -0,0 +1,6 @@
+globalThis.env = {
```
pratapalakshmi, line 1: please check how this operates how this works if there is a change in .env

apps/workflows-dashboard/src/main.tsx
pratapalakshmi, line 7: why `?url`. what is the significance

apps/workflows-dashboard/public/config.js
pratapalakshmi: We will deal only with env variables and therefore .env willnot be used for consideration

apps/workflows-dashboard/index.html
pratapalakshmi: This works because this will be present along with index.html in the dist
```
ls dist
assets            avatar.png        config.js         favicon-16x16.png favicon-32x32.png index.html
```

apps/workflows-dashboard/src/main.tsx
pratapalakshmi: Yes it will

pratapalakshmi:
Need help in migrating [these](https://github.com/ballerine-io/ballerine/blob/dev/apps/workflows-dashboard/src/common/env/schema.ts#L4)

alonp99 (approved):

alonp99 (approved):

alonp99 (approved):

PR #2484, opened 2024-06-27
Fixed home page redirect

Omri-Levy:
## Summary by CodeRabbit

- **New Features**
  - Introduced a new custom hook `useRedirectToRootUrl` for more flexible URL redirections based on user locale and customer configurations.
  
- **Enhancements**
  - Updated redirection logic in the `Logo`, `NotFoundRedirect`, and `AuthProvider` components to use the new `useRedirectToRootUrl` hook.
  - Simplified configuration options for `AuthProvider`.

- **User Experience**
  - Improved URL redirection behavior for authenticated and unauthenticated users.
  - Added a `FullScreenLoader` component to the Home page for better loading state handling.

### DIFF
--- a/apps/backoffice-v2/src/common/components/templates/Providers/Providers.tsx
+++ b/apps/backoffice-v2/src/common/components/templates/Providers/Providers.tsx
@@ -2,32 +2,11 @@ import { QueryClientProvider } from '@tanstack/react-query';
 import { FunctionComponent, PropsWithChildren } from 'react';
 import { queryClient } from '../../../../lib/react-query/query-client';
 import { AuthProvider } from '../../../../domains/auth/context/AuthProvider/AuthProvider';
-import { env } from '../../../env/env';
-import { useLocation } from 'react-router-dom';
-import { useLocale } from '@/common/hooks/useLocale/useLocale';
 
 export const Providers: FunctionComponent<PropsWithChildren> = ({ children }) => {
-  const { state } = useLocation();
-  const locale = useLocale();
-
   return (
     <QueryClientProvider client={queryClient}>
-      <AuthProvider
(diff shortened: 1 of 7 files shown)


apps/backoffice-v2/src/common/components/templates/Providers/Providers.tsx
```diff
@@ -2,32 +2,11 @@ import { QueryClientProvider } from '@tanstack/react-query';
...
-            : `/${locale}/home/statistics`,
-        }}
-        signOutOptions={{
-          redirect: env.VITE_AUTH_ENABLED,
-          callbackUrl: `/${locale}/auth/sign-in`,
-        }}
-      >
-        {children}
-      </AuthProvider>
+      <AuthProvider>{children}</AuthProvider>
```
Omri-Levy, line 9: These configs are no longer needed and `useRedirectToRootUrl` has to be wrapped by the query client provider or it would error.

apps/backoffice-v2/src/domains/auth/context/AuthProvider/AuthProvider.tsx
```diff
@@ -17,33 +20,25 @@ export const AuthContext = createContext<{
...
-  redirectUnauthenticatedTo,
-  signInOptions,
-  signOutOptions,
-}) => {
+export const AuthProvider: FunctionComponentWithChildren = ({ children }) => {
   const { isLoading } = useAuthenticatedUserQuery();
+  const urlToRoot = useRedirectToRootUrl();
+  const { state } = useLocation();
+  const locale = useLocale();
   const contextValues = useMemo(
```
Omri-Levy, line 28: Same config that was passed from outside as before.

alonp99 (approved):

PR #2487, opened 2024-06-30
fix: docker compose server dependency wait for completion

liorzam:
## Summary by CodeRabbit

- **Chores**
  - Updated service dependencies in the `docker-compose.yml` to ensure the `app` service waits for the successful completion of the `migrate` service.


Omri-Levy:
Did you manage to reproduce the issue and confirm it no longer reproduces with the fix?

PR #2489, opened 2024-06-30
fix: docker-compose-build with .env

pratapalakshmi:
## Summary by CodeRabbit

- **Chores**
  - Enhanced environment configuration by adding new variables for improved functionality and integrations.
  - Removed outdated environment variables to streamline the configuration setup.
  - Updated deployment scripts to build Docker images directly from the specified context for greater flexibility.
  - Revised Docker Compose command to utilize the updated build configuration for application deployment.


Omri-Levy (approved):

alonp99 (approved):

deploy/.env
```diff
@@ -17,4 +17,4 @@ WORKFLOW_DASHBOARD_PORT=5200
 WEBSOCKET_SVC_PORT=3500
 KYB_APP_PORT=5201
 DOMAIN_NAME=""
-HASHING_KEY_SECRET="secret"
+HASHING_KEY_SECRET_BASE64=JDJiJDEwJDNFeWtwWEs4QkdiczlRaWFwLkM4Vk8=
```
liorzam: When initializing the monorepo there is a script that generate new salt and transform it to base64 than update .env with the value in order to prevent similar envs with same secrets.

would we like also to replace the hashing key in here ?

pratapalakshmi: Added those changes

liorzam (approved):

deploy/docker-compose-build.yml
```diff
@@ -28,7 +28,8 @@ services:
   ballerine-workflow-service:
     container_name: workflow-service
     platform: linux/amd64
-    image: ghcr.io/ballerine-io/workflows-service:latest
+    build:
+      context: ../services/workflows-service/
```
Omri-Levy, line 32: @alonp99 Does this change align with our goals?

pratapalakshmi: We had to do this because, the latest image is one month old and we have seen this as an issue in one off our OSS customer and so we are building image rather than pulling the docker image

alonp99 (approved):

PR #2512, opened 2024-07-09
Added NOT EQUALS, IN and NOT_IN operations to rule engine

tomer-shvadron:
## Description
* Added the following operations to the rule engine:
  * NOT_EQUALS
  * IN
  * NOT_IN


services/workflows-service/src/rule-engine/core/test/rule-engine.unit.test.ts
```diff
@@ -380,4 +381,180 @@ describe('Rule Engine', () => {
...
+      const engine = RuleEngine(ruleSetExample);
+      let result = engine.run(context);
+
+      expect(result).toBeDefined();
+      expect(result).toHaveLength(1);
+      expect(result[0]).toMatchInlineSnapshot(`
+        {
+          "error": undefined,
+          "rule": {
+            "key": "pluginsOutput.companySanctions.data.length",
```
liorzam, line 407: copy context and push new item to data than check that rule get failed status

liorzam (approved):

PR #2514, opened 2024-07-09
Feature/risk rule plugin lior merge dev

liorzam:


packages/common/src/rule-engine/operators/helpers.ts
```diff
@@ -1,11 +1,11 @@
-import { primitive, ConditionFn, BetweenParams, LastYearsParams, ExistsParams } from './types';
+import { ConditionFn, BetweenParams, LastYearsParams, ExistsParams, Primitive } from './types';
 import { TOperation } from '../types';
 import { ZodSchema } from 'zod';
 import { BetweenSchema, LastYearsSchema, PrimitiveSchema } from './schemas';
 import { ValidationFailedError } from '../errors';
-import { isEmpty } from 'lodash';
+import isEmpty from 'lodash.isempty';
```
tomer-shvadron, line 6: Why do we need this change?

services/workflows-service/src/workflow/workflow.service.ts
```diff
@@ -1939,6 +1943,32 @@ export class WorkflowService {
         },
         // @ts-expect-error - error from Prisma types fix
         extensions: workflowDefinition.extensions,
+        invokeRiskRulesAction: async (
+          context: object,
```
tomer-shvadron, line 1947:
```suggestion
          context: Record<string, unknown>,
```
`object` can be anything, including 'kaki' or 2.

packages/workflow-core/src/lib/plugins/common-plugin/risk-rules-plugin.ts
```diff
@@ -0,0 +1,89 @@
...
+    this.stateNames = pluginParams.stateNames;
+    this.rulesSource = pluginParams.rulesSource;
+    this.action = pluginParams.action;
+    this.successAction = pluginParams.successAction;
+    this.errorAction = pluginParams.errorAction;
+  }
+
+  async invoke(context: TContext) {
+    try {
+      const rulesetResult = await this.action(context, this.rulesSource);
```
tomer-shvadron, line 26: Learning from past experience, I think we should add logging to the `try` and not just the `catch` so we'd have better understanding what happens in the plugin

tomer-shvadron (approved):

apps/backoffice-v2/src/common/components/organisms/Header/hooks/useNavbarLogic/useNavbarLogic.tsx
```diff
@@ -22,6 +22,16 @@ export const useNavbarLogic = () => {
...
+    ...(customer?.config?.isExample || customer?.config?.isDemo
+      ? [
+          {
+            text: 'Home',
+            icon: <Home size={20} />,
+            href: `/${locale}/home`,
+            key: 'nav-item-Home',
+          },
+        ]
+      : []),
```
liorzam, line 34: @Omri-Levy @Blokh Do we need it?

packages/common/src/rule-engine/operators/helpers.ts
liorzam: lodash was exists on workflow service, since its move to 'common' we dont have the full lodash library

packages/common/src/rule-engine/types.ts
```diff
@@ -31,3 +31,8 @@ export type FailedRuleResult = {
 export type RuleResult = PassedRuleResult | FailedRuleResult;
 
 export type RuleResultSet = RuleResult[];
+
+export interface TFindAllRulesOptions {
+  databaseId: string;
+  source: 'notion';
+}
```
liorzam, line 38: @Blokh Is it related to the engine? or should be placed in workflow core?

packages/workflow-core/src/lib/plugins/common-plugin/risk-rules-plugin.ts
liorzam: totally agree!,
we also should inject sentry and notify us for critical issues 
already seen when notion api wasnt respond

apps/backoffice-v2/src/common/components/organisms/Header/hooks/useNavbarLogic/useNavbarLogic.tsx
Omri-Levy: Just the one without the condition.

alonp99: Its a bad merge there are two routes @liorzam

services/workflows-service/src/workflow/workflow.service.ts
liorzam: @Blokh

apps/backoffice-v2/src/domains/workflows/fetchers.ts
```diff
@@ -143,63 +143,6 @@ export const fetchWorkflowById = async ({
             ...data.context?.pluginsOutput?.website_monitoring,
             data: deepCamelKeys(data.context?.pluginsOutput?.website_monitoring?.data ?? {}),
           },
-          risk_evaluation: {
```
liorzam, line 146: @alonp99 Here

PR #2523, opened 2024-07-13
bal 2264 aml rules

liorzam:
- feat: add aml rule
- chore: remove comment

## Summary by CodeRabbit

- **New Features**
  - Added new dependencies for improved type safety and functionality related to Lodash's `get` method.
  
- **Bug Fixes**
  - Enhanced type safety in the `CollectionFlowService` method for better handling of optional properties.

- **Chores**
  - Updated subproject commit reference in data migrations to incorporate recent enhancements or fixes.


packages/common/src/rule-engine/operators/helpers.ts
```diff
@@ -1,9 +1,22 @@
-import { ConditionFn, BetweenParams, LastYearsParams, ExistsParams, Primitive } from './types';
-import { TOperation } from '../types';
+import get from 'lodash.get';
```
MatanYadaev, line 2: Use Remeda instead of Lodash. It's considered more modern, still maintained, and has great TypeScript support built-in.

packages/common/src/rule-engine/operators/helpers.ts
```diff
@@ -24,6 +37,16 @@ export abstract class BaseOperator<T = Primitive> {
 
   abstract evaluate(dataValue: Primitive, conditionValue: T): boolean;
 
+  extractValue(data: any, rule: Rule) {
```
MatanYadaev: If there's no clear reason to use `any`, consider using `unknown` instead. It's safer to assume you don't know the type of the data in this case. It might require you to make some important checks on the data, for example, checking whether it's an array or an object.

packages/common/src/rule-engine/rules/schemas.ts
```diff
@@ -0,0 +1,23 @@
...
+}
+
+export const RuleSchema = z
+  .object({
+    key: z.string(),
+    operator: z.enum(getValues(OPERATION)),
+  })
+  .and(OperationsValueSchema);
+
+// @ts-ignore - cycle zod types are not correct
```
MatanYadaev, line 89: There's a way to use Zod's lazy schemas, you didn't use it well. Please refer to the docs and fix it.

packages/common/src/rule-engine/operators/helpers.ts
liorzam: i havnt seen that they support install parial operations, it means that we have to install the whole dependency in our ballerine common which isnt good idea

MatanYadaev: They support tree-shaking, only the functions you use will get into the bundle.

packages/common/src/rule-engine/rules/schemas.ts
liorzam: couldnt find a reference for that, its looks find as they mentioned in their docs

packages/common/src/rule-engine/errors.ts
```diff
@@ -14,35 +14,26 @@ export class DataValueNotFoundError extends Error {
...
       path: zodIssue.path.join('.'),
     }));
+
+    super(
+      `Validation failed for '${key}', message: ${message}, ${
+        errors ? `error: ${JSON.stringify(error, null, 2)}` : '  '
+      }`,
+    );
+    this.name = 'ValidationFailedError';
+    this.errors = errors;
```
MatanYadaev, line 31: For better logging:
```ts
this.cause = error
```
or:
```ts
super(youMessage, { cause: error })
```

packages/common/src/rule-engine/operators/helpers.ts
liorzam: In the future we will add support to change the property access provider, i think we can get to it later

packages/common/src/rule-engine/rules/schemas.ts
MatanYadaev: If you had used it as they mentioned in the docs, you wouldn't have to use `@ts-ignore`: https://zod.dev/?id=recursive-types

packages/common/src/rule-engine/errors.ts
liorzam: its array of messaged from zodErrors, are you sure cause it should be an array 
`Enumerable | no`

image

alonp99 (approved):

MatanYadaev (approved):

alonp99 (approved):

PR #2538, opened 2024-07-22
feature: added new change in helm chart yaml update action to add normal wf-service image in prisma migrate job

codechirag123:
In this PR, we're adding:

- New input in the YAML update action to update image tag in prismaMigrate job

## Summary by CodeRabbit

- **New Features**
  - Improved deployment workflow by ensuring all necessary image tags are accounted for before executing the Helm chart update process.
- **Enhancements**
  - The `update-helm-chart` job now requires the successful completion of an additional job, enhancing the robustness of the deployment process.


.github/workflows/publish-workflows-service.yml
```diff
@@ -220,7 +220,8 @@ jobs:
           changes: |
             {
               "kubernetes/helm/wf-service/${{steps.update_helm_check.outputs.file_name}}": {
-                "image.tag": "${{ needs.build-and-push-ee-image.outputs.docker_tag }}"
+                "image.tag": "${{ needs.build-and-push-ee-image.outputs.docker_tag }}",
+                "prismaMigrate.image.tag": "${{ needs.build-and-push-image.outputs.docker_tag }}"
```
liorzam, line 224: prisma migrate should be the same image as our current deployment
and what about data-migration?

codechirag123: The current deployment in wf-service uses ee image. In the requirement analysis we discussed that the prismaMigrate will use image without prefix ee.
Regarding the data-migration and db-sync, since both of these jobs use image with prefix ee it will be pulled from the "image.tag"

PR #2560, opened 2024-07-24
UI fixes to the Merchant Monitoring table

tomer-shvadron:
## Summary by CodeRabbit

- **New Features**
	- Introduced search functionality in the Merchant Monitoring page for better report filtering.
	- Enhanced the Clipboard copy functionality by renaming it to `CopyToClipboardButton` and refining the event handling logic.
	- Added a new count method for more accurate counting of business reports.

- **Bug Fixes**
	- Improved error handling in the updated `CopyToClipboardButton`.

- **Style**
	- Updated components for improved UI consistency, including adjustments to the MerchantMonitoringTable for enhanced user interaction.

- **Chores**
	- Refactored import paths for better clarity and maintainability of components, such as relocating the Search component.


apps/backoffice-v2/src/common/components/atoms/CopyToClipboardButton/CopyToClipboardButton.tsx
```diff
@@ -18,7 +18,10 @@ export const CopyToClipboard: FunctionComponent<ICopyToClipboardProps> = ({
     <Button
       variant={'ghost'}
       size={'icon'}
-      onClick={copyToClipboard(textToCopy)}
+      onClick={async e => {
+        e.preventDefault();
```
Omri-Levy:
```suggestion
      onClick={async event => {
        event.preventDefault();
```

apps/backoffice-v2/src/common/hooks/useFilter/useFilter.tsx
```diff
@@ -7,6 +7,8 @@ export const useFilter = () => {
   const onFilter = useCallback(
     (accessor: string) => {
       return (values: Array<string | null>) => {
+        console.log('in useFilter()/onFilter()');
```
Omri-Levy: 👀

apps/backoffice-v2/src/common/hooks/useSearch/useSearch.tsx
```diff
@@ -25,6 +25,8 @@ export const useSearch = (
       return;
     }
 
+    console.log('in useSearch()/useEffect()');
```
Omri-Levy: 👀

apps/backoffice-v2/src/common/hooks/useSearchParamsByEntity/validation-schemas.ts
```diff
@@ -4,7 +4,7 @@ import { CaseStatus, CaseStatuses } from '../../enums';
 export const BaseSearchSchema = z.object({
   sortDir: z.enum(['asc', 'desc']).catch('desc'),
   pageSize: z.coerce.number().int().positive().catch(50),
-  page: z.coerce.number().int().positive().catch(1),
+  page: z.coerce.number().int().positive().catch(10),
```
Omri-Levy: Why page 10 for a catch? What if there are no 10 pages?

apps/backoffice-v2/src/common/utils/get-entity-type-by-filter-id/get-entity-type-by-filter-id.ts
```diff
@@ -1,6 +1,6 @@
 import { getFiltersFromQuery } from '../get-filters-from-query/get-filters-from-query';
-import { TEntityType } from 'src/domains/entities/types';
+import { TEntityType } from '@/common/hooks/useEntityType/useEntityType';
 
-export function getEntityTypeByFilterId(filterId: string): null | TEntityType {
+export const getEntityTypeByFilterId = (filterId: string): null | TEntityType => {
```
Omri-Levy, line 4: Appreciated. 😘

apps/backoffice-v2/src/common/utils/get-severity-from-risk-score.ts
```diff
@@ -1,6 +1,12 @@
 import { Severity, TSeverity } from '@/common/types';
 
-export const getSeverityFromRiskScore = (riskScore: number): TSeverity | undefined => {
+export const getSeverityFromRiskScore = (
+  riskScore: number | null | undefined,
+): TSeverity | undefined => {
+  if (!riskScore && riskScore !== 0) {
+    return;
+  }
```
Omri-Levy, line 8: Its nice just for getting this flow out of the way soon and explicitly but was this needed for some reason if <= 39 catches 0 and if no condition was hit the result is the same?

apps/backoffice-v2/src/domains/business-reports/query-keys.ts
```diff
@@ -9,22 +9,23 @@ import { TBusinessReportType } from '@/domains/business-reports/types';
 
 export const businessReportsQueryKey = createQueryKeys('business-reports', {
   list: ({
-    reportType,
```
Omri-Levy, line 12: Why was report type removed?

apps/backoffice-v2/src/pages/Entities/hooks/useEntities/useEntities.tsx
```diff
@@ -46,6 +46,7 @@ export const useEntities = () => {
   const onFilterChange = useCallback(
     (key: string) => {
       return (values: string[]) => {
+        console.log('in useEntities()/onFilterChange()');
```
Omri-Levy: 👀

apps/backoffice-v2/src/common/hooks/useFilter/useFilter.tsx
tomer-shvadron: Good catch

apps/backoffice-v2/src/pages/MerchantMonitoring/MerchantMonitoring.page.tsx
```diff
@@ -38,14 +42,16 @@ export const MerchantMonitoring: FunctionComponent = () => {
...
       <div className="flex flex-1 flex-col gap-6 overflow-auto">
-        {isNonEmptyArray(businessReports) && (
-          <MerchantMonitoringTable data={businessReports ?? []} />
-        )}
+        {isNonEmptyArray(businessReports) && <MerchantMonitoringTable data={businessReports} />}
         {Array.isArray(businessReports) && !businessReports.length && !isLoadingBusinessReports && (
           <NoBusinessReports />
         )}
         <div className={`mt-auto flex items-center gap-x-2`}>
+          <div className={`flex w-[12ch] text-sm`}>{`Page ${page} of ${totalPages}`}</div>
```
Omri-Levy: Is the loading state covered or when the values are null/undefined? Could be worth adding `?? 0` or similar. Skeleton would be ideal.

apps/backoffice-v2/src/common/hooks/useSearch/useSearch.tsx
tomer-shvadron: Good catch

apps/backoffice-v2/src/common/hooks/useSearchParamsByEntity/validation-schemas.ts
tomer-shvadron: I forgot to remove that. Good catch

apps/backoffice-v2/src/domains/business-reports/query-keys.ts
tomer-shvadron: Together with `search`, they don't get "special treatment" and will be used when spreading `...params` just a couple of lines later

apps/backoffice-v2/src/pages/Entities/hooks/useEntities/useEntities.tsx
tomer-shvadron: Good catch

apps/backoffice-v2/src/common/utils/get-severity-from-risk-score.ts
tomer-shvadron: For typing reasons, I added the `if (!risk)` line, but that would fail if the risk score is 0, so I added it as well to make sure we return the correct value.

apps/backoffice-v2/src/pages/MerchantMonitoring/hooks/useMerchantMonitoringLogic/useMerchantMonitoringLogic.tsx
```diff
@@ -1,31 +1,52 @@
...
+
+  const [{ page, pageSize, sortBy, sortDir, search: searchParamValue }] = useZodSearchParams(
+    MerchantMonitoringSearchSchema,
+  );
+
+  const { search: searchTerm, onSearch } = useSearch({
+    initialSearch: searchParamValue,
+  });
+
+  const search = searchTerm as string;
```
Omri-Levy, line 23: Can searchTerm be undefined?

apps/backoffice-v2/src/pages/MerchantMonitoring/MerchantMonitoring.page.tsx
tomer-shvadron: You're right. I thought it would be better to not present the entire pagination area before we get results (otherwise they are just a component that the user could click and fail), so I wrapped the entire `div` with a `!!totalPages` condition. Tell me what do you think about it

apps/backoffice-v2/src/pages/MerchantMonitoring/hooks/useMerchantMonitoringLogic/useMerchantMonitoringLogic.tsx
tomer-shvadron: It defaults to an empty string, so no.

services/workflows-service/src/business-report/business-report.controller.internal.ts
```diff
@@ -225,37 +225,61 @@ export class BusinessReportControllerInternal {
   @swagger.ApiOkResponse({ type: [String] })
   @swagger.ApiForbiddenResponse({ type: errors.ForbiddenException })
   @common.UsePipes(new ZodValidationPipe(ListBusinessReportsSchema, 'query'))
-  async listBusinessReports(
+  async findMany(
```
Omri-Levy: We use the terminology of `list` in many places. `findMany` has a very strong database or repository connotation for a restful endpoint that responds with an array of a resource.

services/workflows-service/src/business-report/business-report.repository.ts
```diff
@@ -65,4 +65,13 @@ export class BusinessReportRepository {
       ),
     );
   }
+
+  async count<T extends Omit<Prisma.BusinessReportFindManyArgs, 'where'>>(
```
Omri-Levy: Not sure if the omit/where thing needs changing.
```suggestion
  async count<T extends Omit<Prisma.BusinessReportCountArgs, 'where'>>(
```

apps/backoffice-v2/src/domains/business-reports/query-keys.ts
Omri-Levy: No need to revert your change since we don't know when it would happen, but this query is supposed to be used for different views i.e. an ongoing reports view. Didn't understand why it was actively removed.

apps/backoffice-v2/src/pages/MerchantMonitoring/MerchantMonitoring.page.tsx
Omri-Levy: I think these kind of layout shifts are jarring, it could change widths and heights, and I don't follow on why not having `totalPages` or them loading would fail the pagination? For the fail part a disable seems nicer.

Omri-Levy (approved):

apps/backoffice-v2/src/domains/business-reports/query-keys.ts
tomer-shvadron: It was removed from the destructure of `params` which is later on spread (`...params`) into the same place, i.e., nothing changed in the implementation. `reportType` is still being sent

apps/backoffice-v2/src/pages/MerchantMonitoring/MerchantMonitoring.page.tsx
tomer-shvadron: This is not a final design. I talked with Nitzan about him properly designing (UI + UX) the pagination components. This was a quick addition so that users could see that page they're in without reading the URL. The entire pagination `div` is not shown if we didn't get a response from the backend about the business reports. So, for a second while loading, it's not shown (which I think is better than "Page 0 of 0", it think it's a bit dumb as we don't even present the table or the empty view yet)

services/workflows-service/src/business-report/business-report.controller.internal.ts
tomer-shvadron: I don't even know why it was changed... Probably while renaming something else in the service / repo...

PR #2561, opened 2024-07-24
Connect statistics to DB

MatanYadaev:
## Summary by CodeRabbit

- **New Features**
  - Enhanced the Statistics component with loading state management and dynamic data fetching for improved interactivity.
  - Updated the PortfolioRiskStatistics component to display detailed risk statistics based on the newly fetched data.
  - Introduced a new API endpoint for retrieving home metrics, enhancing data-fetching capabilities.

- **Improvements**
  - Improved type safety and data handling in various components through structured data schemas.
  - Enhanced user experience by providing clearer and more informative content across UI elements.


apps/backoffice-v2/src/pages/Statistics/Statistics.page.tsx
```diff
@@ -1,15 +1,23 @@
...
 import { UserStatistics } from '@/pages/Statistics/components/UserStatistics/UserStatistics';
 import { PortfolioRiskStatistics } from '@/pages/Statistics/components/PortfolioRiskStatistics/PortfolioRiskStatistics';
 import { WorkflowStatistics } from '@/pages/Statistics/components/WorkflowStatistics/WorkflowStatistics';
+import { useStatisticsQuery } from '@/pages/Statistics/statistics.query';
+import { Loader2 } from 'lucide-react';
 
 export const Statistics: FunctionComponent = () => {
+  const { data } = useStatisticsQuery();
+
+  if (!data) {
```
Omri-Levy: Why not use the `isLoading` flag?

apps/backoffice-v2/src/pages/Statistics/components/PortfolioRiskStatistics/hooks/usePortfolioRiskStatisticsLogic/usePortfolioRiskStatisticsLogic.tsx
```diff
@@ -5,82 +5,74 @@ import {
...
+  const totalRiskIndicators = data.riskIndicators.reduce((acc, curr) => acc + curr.count, 0);
+  const riskIndicators = useMemo(() => {
+    return data.riskIndicators
+      .sort((a, b) => {
+        if (riskIndicatorsSorting === 'asc') {
+          return a.count - b.count;
+        }
+        return b.count - a.count;
+      })
+      .slice(0, 5);
```
Omri-Levy, line 36: Slice was intentionally placed before sort since sort mutates and slice returns a new array. Last time I checked I could not use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toSorted in the backoffice.

Omri-Levy, line 36: Slice was intentionally placed before sort since sort mutates and slice returns a new array. Last time I checked I could not use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toSorted in the backoffice.

apps/backoffice-v2/src/pages/Statistics/Statistics.page.tsx
```diff
@@ -1,15 +1,23 @@
...
+    return <Loader2 className={'w-4 animate-spin'} />;
+  }
+
   return (
     <div>
       <h1 className={'pb-5 text-2xl font-bold'}>Statistics</h1>
       <div className={'flex flex-col space-y-8'}>
         <UserStatistics fullName={'John Doe'} />
-        <PortfolioRiskStatistics />
+        <PortfolioRiskStatistics data={data} />
```
Omri-Levy: I think we should explicitly pass the specific properties from data that are in use. Like this its not too clear what the component needs and since props live on a props object there's an added nesting level.

apps/backoffice-v2/src/pages/Statistics/components/PortfolioRiskStatistics/hooks/usePortfolioRiskStatisticsLogic/usePortfolioRiskStatisticsLogic.tsx
```diff
@@ -5,82 +5,74 @@ import {
...
-        .sort((a, b) => {
-          if (sorting === 'asc') {
-            return a.amount - b.amount;
-          }
-
-          return b.amount - a.amount;
-        }),
-    [sorting],
-  );
+  const totalRiskIndicators = data.riskIndicators.reduce((acc, curr) => acc + curr.count, 0);
```
Omri-Levy: Do we have a `sum` util?

apps/backoffice-v2/src/pages/Statistics/statistics.query.ts
Omri-Levy, line 1: Not a blocker. If you do refactor to using the metrics module it would make sense to keep this under `domains/metrics/hooks` and use the same structure as the other queries.

services/workflows-service/src/app.module.ts
```diff
@@ -124,6 +125,7 @@ export const validate = async (config: Record<string, unknown>) => {
     initHttpMoudle(),
     RuleEngineModule,
     NotionModule,
+    StatisticsModule,
```
Omri-Levy: Would be confusing to have both a metrics module and a statistics module. Just a naming thing.

services/workflows-service/src/statistics/statistics.controller.internal.ts
```diff
@@ -0,0 +1,22 @@
...
+import { StatisticsOutputDto } from '@/statistics/statistics-output.dto';
+import { StatisticsInputDto } from '@/statistics/statistics-input.dto';
+
+@swagger.ApiExcludeController()
+@common.Controller('internal/statistics')
+export class StatisticsControllerInternal {
+  constructor(protected readonly service: StatisticsService) {}
+
+  @common.Post('')
+  @swagger.ApiOkResponse({ type: StatisticsOutputDto })
```
Omri-Levy: New endpoints should use `@ApiResponse` and `@Validate` with Typebox like its done in the event endpoint and the new ones for editing workflow definitions. 🙂

services/workflows-service/src/statistics/statistics-output.dto.ts
```diff
@@ -0,0 +1,45 @@
...
+  @Type(() => ReportsByRiskLevel)
+  approved!: ReportsByRiskLevel;
+}
+
+export class StatisticsOutputDto {
+  @ValidateNested({ each: true })
+  @Type(() => RiskIndicatorsStatisticsDto)
+  riskIndicators!: RiskIndicatorsStatisticsDto[];
+
+  @ValidateNested()
```
Omri-Levy: How does `ValidateNested` work with only `Type` decorators? Doesn't it only work with `@IsNumber` and so on?

services/workflows-service/src/statistics/statistics.service.ts
```diff
@@ -0,0 +1,146 @@
...
+const HIGH_LTE_RISK_SCORE = 84;
+const CRITICAL_GTE_RISK_SCORE = 85;
+
+@Injectable()
+export class StatisticsService {
+  constructor(private readonly prismaService: PrismaService) {}
+
+  async getStatistics(
+    input: StatisticsInputDto,
+    projectIds: TProjectIds,
```
Omri-Levy: Thoughts?
```suggestion
    currentProjectId: TProjectId,
```

services/workflows-service/src/statistics/statistics.service.ts
```diff
@@ -0,0 +1,146 @@
...
+          "riskIndicator" ->> 'name' ASC;`
+    ).map(({ name, count }) => ({
+      name,
+      count: Number(count),
+    }));
+  }
+
+  async getReportsByRiskLevel(projectId: TProjectId) {
+    const results = await this.prismaService.$queryRaw<
+      Array<{ riskLevel: 'low' | 'medium' | 'high' | 'critical'; count: number }>
```
Omri-Levy: We have a type in `@ballerine/common` called `TSeverity` and a `Severity` enum.

apps/backoffice-v2/src/pages/Statistics/components/PortfolioRiskStatistics/hooks/usePortfolioRiskStatisticsLogic/usePortfolioRiskStatisticsLogic.tsx
MatanYadaev: 10% of users use browsers that don't support `toSorted`. For now I'll use `structuredClone` to not mutate the original value.

apps/backoffice-v2/src/pages/Statistics/components/PortfolioRiskStatistics/hooks/usePortfolioRiskStatisticsLogic/usePortfolioRiskStatisticsLogic.tsx
MatanYadaev: Remeda has, but we didn't install it in the backoffice.

apps/backoffice-v2/src/pages/Statistics/components/PortfolioRiskStatistics/hooks/usePortfolioRiskStatisticsLogic/usePortfolioRiskStatisticsLogic.tsx
```diff
@@ -5,82 +5,74 @@ import {
...
-    return sortedData.map(item =>
-      item.amount === 0 ? 0 : Math.max((item.amount / maxValue) * 100, 2),
+    return riskIndicators.map(item =>
+      item.count === 0 ? 0 : Math.max((item.count / maxValue) * 100, 2),
     );
-  }, [sortedData]);
+  }, [riskIndicators]);
   const filters = [
     {
       name: 'Merchant Monitoring',
```
Omri-Levy, line 47: The merchant monitoring navigation link is not from a filter by the way. Since `entity` right now represents `individuals` or `businesses` we might want to reconsider the `entityPlural` approach. Also, are you familiar with this package? https://www.npmjs.com/package/pluralize

MatanYadaev: I am familiar with `pluralize`, but I don't really need it here. The term is always plural, so when we add it to the JSON, it should already be pluralized. Why add another package to the runtime when I can just type the right entity?

services/workflows-service/src/statistics/statistics.service.ts
MatanYadaev: It's in the backoffice, not the `@ballerine/common`.

Omri-Levy: The PR that moves the report UI components to shared is the one that moves the types and enum to common, its not merged yet, my bad.

apps/backoffice-v2/src/pages/Statistics/components/PortfolioRiskStatistics/hooks/usePortfolioRiskStatisticsLogic/usePortfolioRiskStatisticsLogic.tsx
Omri-Levy: Adding a new column to filters for a very specific use case not directly related to how filters are used is something I don't want to do if I don't have to. `entity` and `entityPlural` can be different words entirely for better and worse.

services/workflows-service/src/metrics/metrics.controller.ts
```diff
@@ -110,4 +119,19 @@ export class MetricsController {
...
+  @ApiResponse({
+    status: 403,
+    description: 'Forbidden',
+    schema: Type.Record(Type.String(), Type.Unknown()),
+  })
+  @Validate({
+    response: HomeMetricsSchema,
+  })
+  async getHomeMetrics(
+    @ProjectIds() projectIds: TProjectIds,
```
Omri-Levy: Usually we use this decorator.
```suggestion
    @CurrentProject() currentProjectId: TProjectId,
```

Omri-Levy (approved):

apps/backoffice-v2/src/pages/Statistics/components/PortfolioRiskStatistics/hooks/usePortfolioRiskStatisticsLogic/usePortfolioRiskStatisticsLogic.tsx
MatanYadaev: Sorry, I have to move on. I'll merge the PR for now but we will revisit this topic later if you want.

PR #2591, opened 2024-07-31
List customers and set customer config

MatanYadaev:
## Summary by CodeRabbit

- **New Features**
  - Introduced a new method to list customers, accessible to admins only.
  - Added a configuration option in the customer creation process, allowing for more detailed customer setup.
  - Enhanced configuration schema to include an optional property for merchant monitoring.

- **Bug Fixes**
  - Improved validation for the new configuration property during customer creation to ensure compliance with the defined schema.

- **Documentation**
  - Updated API documentation to reflect the addition of the new `config` property in customer creation.


alonp99 (approved):

services/workflows-service/src/customer/customer.controller.external.ts
```diff
@@ -25,12 +26,20 @@ export class CustomerControllerExternal {
     protected readonly prisma: PrismaService,
   ) {}
 
+  @common.Get()
```
alonp99: It should probably be internal

PR #2592, opened 2024-07-31
Added the ability to define default value for document fields

Omri-Levy:
## Summary by CodeRabbit

- **New Features**
	- Introduced a new parameter for handling default values in the document blocks hook.
	- Enhanced logic for determining field values, improving reliability and flexibility.

- **Refactor**
	- Improved control flow by encapsulating field value determination within a dedicated function for better clarity.


apps/backoffice-v2/src/lib/blocks/hooks/useDocumentBlocks/useDocumentBlocks.tsx
```diff
@@ -388,9 +388,32 @@ export const useDocumentBlocks = ({
...
+                        if (
+                          typeof properties?.[title] === 'undefined' &&
+                          typeof defaultValue !== 'undefined'
+                        ) {
+                          return defaultValue;
+                        }
+
+                        if (typeof properties?.[title] === 'undefined' && type === 'boolean') {
+                          return false;
+                        }
```
Omri-Levy, line 408: Can be worthwhile to map other types as well such as returning 0 for type number and so on.

apps/backoffice-v2/src/lib/blocks/hooks/useDocumentBlocks/useDocumentBlocks.tsx
```diff
@@ -388,9 +388,32 @@ export const useDocumentBlocks = ({
...
+
+                        if (typeof properties?.[title] === 'undefined') {
+                          return '';
+                        }
```
Omri-Levy, line 412: Made all checks in a way that won't fail on things like `0`, `false`, empty strings and so on. `null` is actually a valid value especially since some fields could be nullable.

alonp99 (approved):

PR #2597, opened 2024-08-01
Feeature/resolve bal2455 conflicts

Blokh:
resolved bal2455 with dev conflicts


services/workflows-service/src/workflow/workflow.service.ts
```diff
@@ -2078,6 +2087,7 @@ export class WorkflowService {
             customerNormalizedName: customer.name,
           };
         },
+        secretsManager: { getAll: secretsManager.getAll.bind(secretsManager) },
```
liorzam, line 2090: also in here missing this part of workflow creations become static

liorzam (changes requested):

PR #2617, opened 2024-08-05
Bal 2552 lior transactions improvements

liorzam:


services/workflows-service/src/transaction/transaction.controller.external.ts
```diff
@@ -333,42 +336,82 @@ export class TransactionControllerExternal {
     @Query() getTransactionsByAlertParameters: GetTransactionsByAlertDto,
     @CurrentProject() projectId: types.TProjectId,
   ) {
-    const alert = await this.prisma.alert.findUnique({
-      where: {
-        id: getTransactionsByAlertParameters.alertId,
+    const queryArgs = this.scopeService.scopeFindOne(
```
alonp99, line 339: You should use "scope service" in the repository, not in the service
the service should use the repository and it handles scoping etc..

services/workflows-service/src/transaction/transaction.controller.external.ts
```diff
@@ -333,42 +336,82 @@ export class TransactionControllerExternal {
...
+        },
+        where: {
+          id: getTransactionsByAlertParameters.alertId,
+        },
       },
-    });
+      [projectId],
+    );
+
+    const alert = await this.prisma.alert.findUnique(queryArgs);
```
alonp99, line 351: user the service..

services/workflows-service/src/transaction/transaction.controller.external.ts
```diff
@@ -333,42 +336,82 @@ export class TransactionControllerExternal {
...
-    const inlineRule = alertDefinition.inlineRule as InlineRule;
-    // @ts-expect-error
-    const { timeValue, timeUnit } = inlineRule.options;
+    if (!inlineRule || !inlineRule.options) {
+      return filters;
+    }
+    if (!filters.endDate) {
+      filters.endDate = alert.updatedAt;
+    } else if (
+      !filters.startDate &&
```
alonp99, line 374: we can remove start date, this endpoint is get transaction by alert - so we can remove almost all filters and let the backend handle what he fetches by the alert definition, no?

we have the other endpoint if the user want to fetch by other filters like end date etc..

PR #2625, opened 2024-08-07
WIP - Move report components to common packages

Omri-Levy:
- feat(*): checkpoint
- feat(*): checkpoint


apps/backoffice-v2/src/lib/blocks/components/DirectorsCallToAction/interfaces.ts
```diff
@@ -0,0 +1,10 @@
+export interface ICallToActionDocumentOption {
+  name: string;
+  value: string;
+}
+
+export interface ICallToActionDocumentSelection {
+  options: ICallToActionDocumentOption[];
+  value?: ICallToActionDocumentOption['value'];
+  onSelect: (value: ICallToActionDocumentOption['value']) => void;
+}
```
Omri-Levy, line 10: Previously this type was imported but it did not exist. We need to properly deprecate directors call to action.

tomer-shvadron (approved):

PR #2635, opened 2024-08-09
fix: wait for DB container to stabilize

riturajgc:
Updates the setup script to add a sleep command before starting `db:reset` and data seeding. Without this, the application fails to start due to the following error, as the DB containers are still not completely ready to accept connections.

image

## Summary by CodeRabbit

- **Chores**
	- Updated the setup script to include a 10-second pause between database shutdown and reset to ensure a stable state before reinitialization.
- **New Features**
	- Expanded document type schema to include new categories: `eida` and `trade_license`.
	- Introduced a new optional property `isNameAsTradeLicense` to enhance document validation.
- **Bug Fixes**
	- Adjusted workflow event handling with a delay to manage timing issues during customer data fetching.
- **Improvement**
	- Enhanced session management by removing session protection from various routes.


services/workflows-service/package.json
```diff
@@ -5,7 +5,7 @@
   "description": "workflow-service",
   "scripts": {
     "spellcheck": "cspell \"*\"",
-    "setup": "npm run docker:db:down && npm run docker:db && npm run db:reset && npm run seed",
+    "setup": "npm run docker:db:down && npm run docker:db && sleep 10 && npm run db:reset && npm run seed",
```
Omri-Levy, line 8: Hey @riturajgc , thanks for the contribution! Ideally the solution would wait for the database to be ready by using something like a lifecycle hook or similar over a sleep. Is this something you're willing to explore?

Omri-Levy: I can see we use a package called `wait-on`, could you check if by waiting on the db port is solves the issue? 🙏🏼

Omri-Levy:
A fix was already merged.

PR #2642, opened 2024-08-14
Replaced pagination component

Omri-Levy:
## Summary by CodeRabbit

- **New Features**
	- Enhanced pagination controls, allowing users to navigate directly to the last page.
	- Introduced visual loading states using Skeleton components for improved user feedback during data loading.
	- Added conditional rendering for pagination and no results components to streamline user experience.

- **Bug Fixes**
	- Improved accessibility for disabled pagination links, ensuring they are visually distinct and non-interactive.

- **Chores**
	- Cleaned up import statements and optimized variable naming for better clarity.


apps/backoffice-v2/src/pages/Entities/hooks/useEntities/useEntities.tsx
```diff
@@ -86,20 +78,27 @@ export const useEntities = () => {
 
   const { workflowDefinition } = useCaseCreationWorkflowDefinition();
 
+  const isNoCases = !isLoading && Array.isArray(cases) && !cases.length;
```
tomer-shvadron, line 81: Do you think we should use `useMemo` here?

apps/backoffice-v2/src/pages/Entities/hooks/useEntities/useEntities.tsx
```diff
@@ -20,11 +21,8 @@ export const useEntities = () => {
     pageSize,
     search,
   });
-
-  const {
-    meta: { totalPages },
-    data: workflows,
-  } = data || { meta: { totalPages: 0 }, data: [] };
+  const cases = data?.data;
+  const totalPages = data?.meta?.totalPages ?? 0;
```
tomer-shvadron, line 25: Do you think we should use `useMemo` here?

tomer-shvadron (approved):

apps/backoffice-v2/src/pages/Entities/hooks/useEntities/useEntities.tsx
Omri-Levy: No. There is no computation here, no multiple references made (i.e. an object inside an array), and these values are not used in many `useEffect` calls or similar.

apps/backoffice-v2/src/pages/Entities/hooks/useEntities/useEntities.tsx
Omri-Levy: `useMemo` is never used for derived state, maybe something very rare.

PR #2712, opened 2024-09-16
chore(wf): add data to error logs

alonp99:
## Summary by CodeRabbit

- **New Features**
	- Enhanced error handling capabilities with additional context in error responses, including request details such as URL, method, headers, query, body, user, IP, user agent, and referer.
	- Improved observability and debuggability of errors for better issue diagnosis in production environments.
	- Robust handling of request metadata, ensuring only valid and non-sensitive data is included.


Omri-Levy (approved):

services/workflows-service/src/common/filters/AllExceptions.filter.ts
```diff
@@ -77,6 +78,15 @@ export class AllExceptionsFilter extends BaseExceptionFilter {
       error: inspect(errorRes),
       message: error.message,
       responseTime: Date.now() - request.startTime,
+      url: request.url,
+      method: request.method,
+      headers: removeSensitiveHeaders(request.headers),
+      query: request.query,
+      body: request.body,
```
liorzam: super risky should check that the http content is json

liorzam: you can use `getReqMetadataObj` and also add the suggested body manipulation for json only 

```
export const getReqMetadataObj = (req: Request<unknown>) => {
  const cleanHeaders = removeSensitiveHeaders(req.headers);
  const isJsonPayload = req.headers['content-type']?.includes('application/json');
  
  // Check if the body is a valid JSON (object or array, but not null)
  const isValidBody = isJsonPayload && req.body !== null && (typeof req.body === 'object');

  return {
    startTime: req.startTime ? new Date(req.startTime)?.toISOString() : new Date().toISOString(),
    'X-Request-ID': req.id,
    query: req.query,
    params: req.params,
    url: req.originalUrl,
    method: req.method,
    headers: cleanHeaders,
    body: isValidBody ? req.body : undefined, // Log only if valid JSON (object or array)
  };
};
```

alonp99: good point, done

PR #2725, opened 2024-09-23
Added Saola integration

tomer-shvadron:
## Summary by CodeRabbit

- **New Features**
	- Added support for session recording using the Saola AI library.
	- Introduced an optional API key configuration (`VITE_SAOLA_API_KEY`) for enhanced functionality in both applications.
  
- **Chores**
	- Updated project dependencies to include the Saola AI library (`@saola.ai/browser` version `^1.1.7`).

### DIFF
--- a/apps/backoffice-v2/src/initialize-session-recording/initialize-session-recording.ts
+++ b/apps/backoffice-v2/src/initialize-session-recording/initialize-session-recording.ts
@@ -1,0 +1,15 @@
+import { env } from '@/common/env/env';
+import { init as initSaola } from '@saola.ai/browser';
+
+export const initializeSessionRecording = () => {
+  // @TODO: Uncomment when Saola tests are production ready
+  // if (window.location.host.includes('127.0.0.1') || window.location.host.includes('localhost')) {
+  //   return;
+  // }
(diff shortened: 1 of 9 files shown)


apps/backoffice-v2/src/initialize-session-recording/initialize-session-recording.ts
```diff
@@ -0,0 +1,15 @@
+import { env } from '@/common/env/env';
+import { init as initSaola } from '@saola.ai/browser';
+
+export const initializeSessionRecording = () => {
+  // @TODO: Uncomment when Saola tests are production ready
+  // if (window.location.host.includes('127.0.0.1') || window.location.host.includes('localhost')) {
+  //   return;
+  // }
```
liorzam, line 8: we do use domain name for non production environment like 'dev' + 'sb' 
and i suggest to enable it on sb+dev env in order to record sessions

tomer-shvadron: It is enabled in dev already, when we get to sb we'll enable it there too (in Saola's settings)

PR #2793, opened 2024-10-22
Notion schema fix and Cron schedule change

tomer-shvadron:
## Release Notes

- **New Features**
	- Introduced a new operator, `IN_CASE_INSENSITIVE`, enhancing the rule schema's capabilities for case-insensitive operations.
	- Expanded the `riskLevel` categorization to include a new value, `high`, for more nuanced risk assessments.

- **Bug Fixes**
	- Updated the cron job frequency from every minute to every hour for improved resource management.

- **Refactor**
	- Streamlined import statements for better organization in the rule engine package.
	- Refactored function declarations to modern syntax without changing functionality.

- **Tests**
	- Enhanced unit tests to validate the new `IN_CASE_INSENSITIVE` operator and improved error handling for various scenarios in the rule engine.


MatanYadaev (approved):

packages/workflow-core/src/lib/plugins/common-plugin/types.ts
```diff
@@ -75,7 +75,7 @@ export interface RiskRulesPluginParams {
       id: string;
       domain: string;
       indicator: string;
-      riskLevel: 'critical' | 'moderate' | 'positive';
+      riskLevel: 'critical' | 'high' | 'moderate' | 'positive';
```
liorzam, line 78: Can we infer the value from here? looks pretty similar for the type we just generated here 
image

liorzam (approved):

packages/workflow-core/src/lib/plugins/common-plugin/types.ts
tomer-shvadron: In general it's a good idea and can be done pretty easily by using `zod.infer`, the issue is that the schema exists in `workflows-service` but the potential use of it is in `workflow-core`. Since right now we don't share types that are related to the reports in the `common` package (something that in general we should), I don't want to move just this type, as it could lead to a bit of headache for the wide-spread of the types.

PR #2828, opened 2024-11-09
fix: improve transaction monitoring - HVHAI_APM | MGAV_CC and fix flaky tests

liorzam:
## Summary by CodeRabbit

- **New Features**
	- Enhanced indexing strategy for improved query performance on `TransactionRecord` and `Alert` models.
	- Introduced a new optional parameter for dynamic date generation in the `TransactionFactory` class.

- **Bug Fixes**
	- Updated alert generation tests to improve accuracy based on transaction conditions.

- **Refactor**
	- Simplified SQL query structure in the `DataAnalyticsService` for better readability and maintainability.

- **Tests**
	- Expanded test cases for alert generation to include new conditions and refined logic.

### DIFF
--- a/services/workflows-service/src/transaction/test-utils/transaction-factory.ts
+++ b/services/workflows-service/src/transaction/test-utils/transaction-factory.ts
@@ -197,24 +197,29 @@ export class TransactionFactory {
 
   projectId: string;
 
+  _fakeDateByFn: (() => Date) | undefined;
+
   constructor({
     prisma,
     number = 1,
     data = {},
     runBeforeCreate = [],
     projectId,
+    fakeDateByFn,
   }: {
     prisma: PrismaService;
     projectId: string;
     number?: number;
     data?: Partial<TransactionCreateData>;
     runBeforeCreate?: Array<() => Promise<void>>;
+    fakeDateByFn?: () => Date;
   }) {
     this.prisma = prisma;
     this.number = number;
     this.data = data;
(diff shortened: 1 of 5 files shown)


services/workflows-service/src/transaction/test-utils/transaction-factory.ts
```diff
@@ -369,6 +375,14 @@ export class TransactionFactory {
...
   }
 
+  public date(dateFunction: (...args: any[]) => Date, ...args: any[]) {
+    const factory = this.clone();
+
+    factory._fakeDateByFn = () => dateFunction(...args);
+
+    return factory;
+  }
+
```
liorzam, line 385: @MatanYadaev @alonp99  FYI: 
The purpose of this change is to allow each generated transaction to have a unique date based on the provided `dateFunction`.

 *  Previously, all transaction records might have been using the same date generate by the faker, leading to a lack of variation in transaction timestamps. all the timestamps were the same!
 *  By introducing the `date` method with a `dateFunction` (such as `faker.date.past()`, `faker.date.recent()`, etc.), we ensure that each transaction can have its own random date, adding variability to the generated data.
 *  The method works by cloning the factory instance and assigning the custom date logic through `_fakeDateByFn`, which is executed when a date is needed for each record.

alonp99 (approved):

alonp99 (approved):

PR #2837, opened 2024-11-14
Search fix for reports

Omri-Levy:
- fix(backoffice-v2): no longer not rendering search when fetching reports
- fix(backoffice-v2): fixed disabled flag

## Summary by CodeRabbit

- **New Features**
	- Introduced an optional `isDisabled` prop for the `Search` component, allowing it to be disabled based on certain conditions.

- **Improvements**
	- The `Search` component is now always rendered in the `MerchantMonitoring` page, with its interactivity controlled by the `isDisabled` prop depending on the loading state and availability of business reports. 

- **Dependencies**
	- Updated the `tailwindcss` dependency version to `^3.4.14`.

These changes enhance user experience by providing consistent access to the search functionality while managing its usability effectively.

### DIFF
--- a/apps/backoffice-v2/package.json
+++ b/apps/backoffice-v2/package.json
@@ -190,7 +190,7 @@
     "prettier-plugin-tailwindcss": "^0.2.1",
     "storybook": "^7.0.0-rc.10",
     "storybook-addon-react-router-v6": "^1.0.2",
-    "tailwindcss": "^3.2.4",
+    "tailwindcss": "^3.4.14",
     "type-fest": "^4.23.0",
     "typescript": "^5.5.4",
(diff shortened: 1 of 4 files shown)


apps/backoffice-v2/package.json
```diff
@@ -190,7 +190,7 @@
     "prettier-plugin-tailwindcss": "^0.2.1",
     "storybook": "^7.0.0-rc.10",
     "storybook-addon-react-router-v6": "^1.0.2",
-    "tailwindcss": "^3.2.4",
+    "tailwindcss": "^3.4.14",
```
Omri-Levy, line 193: Enables :has CSS class.

PR #2860, opened 2024-11-20
feat(assign-dropdown): add role filtering for assignees

alonp99:
- Introduce filterUsersByRole utility to exclude assignees by roles
- Update AssignDropdown and AlertsAssignDropdown to utilize filtering
- Set default excluded role to 'viewer' in relevant components

(Your role filtering is so selective, it's like a bouncer at a VIP club)

## Summary by CodeRabbit

- **New Features**
  - Introduced an `excludedRoles` property to the `AssignDropdown` and `AlertsAssignDropdown` components, allowing users to filter out specific roles from the dropdown options.
  - Enhanced filtering logic to prioritize authenticated users while excluding specified roles from the assignee list.
  - Added a `roles` property to the `UsersListSchema` validation schema for improved user role validation.

- **Bug Fixes**
  - Improved the handling of user roles in the dropdown components to ensure accurate filtering.

- **Documentation**
  - Updated prop signatures to reflect the addition of the `excludedRoles` property in relevant components.

### DIFF
--- a/apps/backoffice-v2/src/domains/users/utils/filter-users-by-role.ts
+++ b/apps/backoffice-v2/src/domains/users/utils/filter-users-by-role.ts
@@ -1,0 +1,18 @@
+import { TAuthenticatedUser } from '@/domains/auth/types';
+
+export type TUserRole = 'viewer';
+
+export const filterUsersByRole = (
+  users: Array<Partial<TAuthenticatedUser>>,
+  excludedRoles: TUserRole[],
+) => {
+  return users.filter(user => {
+    if (!user) return false;
+
+    // If user has no roles, include them
+    if (!('roles' in user)) return true;
+
+    // Exclude user if they have any of the excluded roles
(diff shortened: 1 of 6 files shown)


apps/backoffice-v2/src/domains/users/utils/filter-users-by-role.ts
```diff
@@ -0,0 +1,18 @@
...
+  excludedRoles: TUserRole[],
+) => {
+  if (!Array.isArray(users)) return [];
+
+  return users.filter(user => {
+    if (!user) return false;
+
+    if (!('roles' in user)) return true;
+
+    return !excludedRoles.some(role => Array.isArray(user.roles) && user.roles.includes(role));
```
tomer-shvadron:
```suggestion
    if (!('roles' in user) || !Array.isArray(user.roles)) {
      return true;
    }

    return !excludedRoles.some(role => user.roles.includes(role));
```

tomer-shvadron (approved):

PR #2877, opened 2024-12-08
style(components): change button state styling to fix disappearing buttons

shanegrouber:
- Update button classes to include hover and enabled states
- Remove unused CSS rules to streamline styles

## Summary by CodeRabbit

- **New Features**
	- Enhanced button styling across various components for improved visual feedback during user interactions.

- **Bug Fixes**
	- Simplified rendering of the Chatbot component by removing unnecessary wrapper, potentially improving performance and layout.

- **Documentation**
	- Updated CSS classes for better theming and responsiveness in dark mode.

- **Chores**
	- Removed unused components and import statements to streamline the codebase.


apps/backoffice-v2/src/domains/notes/Notes.tsx
```diff
@@ -88,7 +88,7 @@ export const Notes = ({
                 size={`sm`}
                 aria-disabled={isLoading}
                 className={
-                  'mt-3 h-5 self-end p-4 text-sm font-medium aria-disabled:pointer-events-none aria-disabled:opacity-50'
+                  'mt-3 h-5 self-end p-4 text-sm font-medium enabled:bg-slate-800 enabled:hover:bg-slate-800/90 aria-disabled:pointer-events-none aria-disabled:opacity-50'
```
Omri-Levy: Let's use bg-primary over slate directly. That's the default color for Button.

apps/backoffice-v2/src/lib/blocks/components/CallToActionLegacy/CallToActionLegacy.tsx
```diff
@@ -189,7 +189,10 @@ export const CallToActionLegacy: FunctionComponent<ICallToActionLegacyProps> = (
             size="wide"
             variant="warning"
             disabled={disabled}
-            className={ctw({ 'flex gap-2': isReuploadResetable })}
+            className={ctw({
+              'flex gap-2': isReuploadResetable,
+              'enabled:bg-warning enabled:hover:bg-warning/90': !isReuploadResetable,
```
Omri-Levy: Can use 'enabled' directly without JS. ctw accepts multiple arguments and it can also receive arrays. If we wanted to keep this line the same passing `true` would be less misleading as this class does not apply when the button is disabled regardless of the condition. https://www.npmjs.com/package/clsx

apps/backoffice-v2/src/pages/Entity/components/Case/actions-variants/DefaultActions/DefaultActions.tsx
```diff
@@ -35,7 +35,11 @@ export const DefaultActions = () => {
             size="md"
             variant="warning"
             disabled={isLoadingActions || !canRevision}
-            className={ctw({ loading: debouncedIsLoadingRejectCase }, 'whitespace-nowrap')}
+            className={ctw(
+              { loading: debouncedIsLoadingRejectCase },
+              'whitespace-nowrap',
+              'enabled:bg-warning enabled:hover:bg-warning/90',
```
Omri-Levy, line 41: Use this as reference for the other places with enabled. First argument is an object and second is string literal. 🙂

Omri-Levy (approved):
Great job. 💪🏼

PR #2899, opened 2024-12-18
Dev 335/preview helm changes

codechirag123:
## Summary by CodeRabbit

- **New Features**
	- Introduced a `.helmignore` file to streamline Helm package building.
	- Added a new Helm chart configuration for the "ballerine" application.
	- Implemented multiple Helm templates for services including backoffice, headless example, kyb-app, workflows dashboard, and workflow service.
	- Added Kubernetes Job resources for database migrations, data synchronization, and Prisma migrations.
	- Defined ClusterRole and ClusterRoleBinding for the workflows service.

- **Documentation**
	- Created a comprehensive README file detailing installation and configuration for the Ballerine project.

- **Chores**
	- Introduced a modular import template for dynamic inclusion of service templates.


deploy/helm-preview/Chart.yaml
```diff
@@ -0,0 +1,30 @@
+apiVersion: v2
+name: helm
```
liorzam, line 2: would we like to use this helm also for other purposes ?
Do we want our OSS users will use it or its will be only for our usage?

codechirag123: This helm is only created for running preview environment.
We wanted a review on this helm to verify the changes that we performed so that it doesn't affect the OSS users.
Although we have verified it on our end.
We propose to replace the oss helm with this helm as it can support both OSS and preview purposes.

pratapalakshmi (commented):
Any reason why we are not using the existing helm-charts present [here](https://github.com/ballerine-io/ballerine/tree/main/deploy/helm)?
Also, I would recommend to add Horizontal pod Auto-scaling?

deploy/helm-preview/Chart.yaml
```diff
@@ -0,0 +1,30 @@
...
+
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application. Versions are not expected to
+# follow Semantic Versioning. They should reflect the version the application is using.
+# It is recommended to use it with quotes.
+appVersion: "0.4.14"
+dependencies:
+  - condition: postgresql.enabled
+    name: postgresql
+    version: 12.0.0
```
pratapalakshmi, line 28: please check the image that is being used,
"sibedge/postgres-plv8:15.3-3.1.7" we use this image. I am not sure the postgres that comesup will have this image. or ballerine is compatible with the postgres image

PR #2911, opened 2024-12-27
feat(business-report): upgrade user-facing part of the social media data UI (BAL-2577)

r4zendev:
- Updates social media tab UI according to the one described in the initial ticket;
- Updates lucide-react to a recent version (was required to pull the Ban icon that is in the UI scratch);
- Removes redundant abstractions where possible and improves the logic in general;
- Adds temporary replacement for ads section to display empty state as it was marked as unnecessary to complete now. I wasn't sure if this section is to be deprecated and removed from the design altogether or if we actually want to have it later, so I did not remove it completely for now.

Examples of the UI:
image
image

In 1920x1080 aspect ratio:
image

With disabled states:
image

Small explanation as to why the customers tab in the backoffice application was throwing 401 for me:
In `services/workflows-service/src/common/middlewares/auth-key.middleware.ts` there is a middleware that assigns the `customer` role to any request going through it. Changing the role to `admin` locally allowed me to make a request and see the customers tab. Just a note for you in case it needs to be fixed.

Note: we are missing decent loading states.
Note 2: I wasn't sure how we maintain our change logs and bump packages, so I just bumped the ui package and updated the `CHANGELOG.md` manually. I saw `changesets` but IIRC they keep the changelog for the entire monorepo and store it in the root. I didn't see any kind of npm script or anything similar in the `ui` package. Let me know if I should do something here.

## Summary by CodeRabbit

- **New Features**
	- Introduced a new `build:watch` script for building the project in watch mode.
	- Added a new button variant, `browserLink`, to enhance visual presentation.
	- New components `FacebookIcon` and `InstagramIcon` added for social media representation.
	- Enhanced `AdsAndSocialMedia` component with improved layout and functionality.
	- Added new functions `toSocialMediaPresence` and `toAdsImages` for structured data extraction.

- **Bug Fixes**
	- Made `width` and `height` properties optional in the `ImageProps` interface.

- **Removals**
	- Removed `AdImageWithLink` component and its associated exports.
	- Deleted exports for `ads-provider-adapter` and `create-report-adapter`, affecting accessibility in the application.
	- Removed export for `report-adapter` and its associated exports.

- **Refactor**
	- Updated `reportAdapter` logic for better data processing and structure.
	- Restructured `useReportTabs` for clearer component organization.


packages/ui/src/components/templates/report/adapters/report-adapter.ts
```diff
@@ -23,6 +22,47 @@ export const toRiskLabels = (riskIndicators: Array<{ name: string; riskLevel: st
   }));
 };
 
+export const toSocialMediaPresence = (data: Record<string, any>) => {
```
MatanYadaev, line 25: WDYT about using `unknown` instead of `any`? I think we should try to use `any` less often because it weakens type safety, which can cause runtime errors and make the code harder to maintain.

rzssh: sgtm, i'd say that the best approach would be to type the responses/normalized data in the end, so any temporary loose type here works and unknown is prolly better indeed.

rzssh: image

however, using unknown basically makes it a bit more strict and sort of forces dev to type it properly. unsure if we want to prioritize that right now since from my observations data coming from agents gets transformed a lot in the middle, i'd start with simplifying the pipelines that the data goes through first tbh along with properly typing incoming data. we have quite a lot of anys in the codebase anyway. it might be a good candidate for a tech debt epic.

MatanYadaev (approved):
Great work Sasha! Two comments:
1. A new version should be published. You can do it with 2 commands: `pnpm changeset` and then `pnpm version-packages`
2. Please hide the ads section. Currently it doesn't work well, we prefer to not show this section to our clients at the moments.

packages/ui/src/components/templates/report/components/AdsAndSocialMedia/AdsAndSocialMedia.tsx
```diff
@@ -1,104 +1,236 @@
...
+      <h3 className="mb-2 text-base font-bold">Ads</h3>
+      <Card
+        className={ctw(
+          'flex w-full justify-between p-4 shadow-lg',
+          // relatedAdsImages && relatedAdsImages.length > 0 ? 'opacity-100' : 'opacity-60',
+          false ? 'opacity-100' : 'opacity-60',
+        )}
+      >
+        {/* relatedAdsImages */}
+        {false ? (
```
alonp99: Can you explain this part?

packages/ui/src/components/templates/report/components/AdsAndSocialMedia/AdsAndSocialMedia.tsx
```diff
@@ -1,104 +1,236 @@
...
+    </div>
+
+    {/* TODO: Shows failing state by default for now */}
+    <div>
+      <h3 className="mb-2 text-base font-bold">Ads</h3>
+      <Card
+        className={ctw(
+          'flex w-full justify-between p-4 shadow-lg',
+          // relatedAdsImages && relatedAdsImages.length > 0 ? 'opacity-100' : 'opacity-60',
+          false ? 'opacity-100' : 'opacity-60',
```
alonp99: 🤔

packages/ui/src/components/templates/report/components/AdsAndSocialMedia/AdsAndSocialMedia.tsx
rzssh: i left it hidden by default using hardcoded condition, since there was no logic for ads. as Matan mentioned it can be removed, I replaced harcoded parts with the actual logic and commented out the entire ads section.

PR #2923, opened 2024-12-31
Allow sending child workflows in webhooks

Omri-Levy:
## Summary by CodeRabbit

- **Dependency Updates**
  - Updated `@ballerine/common` from version 0.9.61 to 0.9.63 across multiple packages
  - Updated `@ballerine/workflow-browser-sdk` from version 0.6.81 to 0.6.82
  - Updated `@ballerine/workflow-node-sdk` from version 0.6.81 to 0.6.82
  - Updated `@ballerine/workflow-core` from version 0.6.81 to 0.6.82
  - Updated `@ballerine/backoffice-v2` from version 0.7.86 to 0.7.87
  - Updated `@ballerine/kyb-app` from version 0.3.101 to 0.3.102
  - Updated `@ballerine/headless-example` from version 0.3.80 to 0.3.81
  - Updated `@ballerine/workflows-service` from version 0.7.85 to 0.7.86
  - Updated `@ballerine/web-ui-sdk` from version 1.5.62 to 1.5.64

- **Risk Evaluation Schema**
  - Modified the type of `value` property in risk evaluation plugin schema to be more flexible

- **Workflow Enhancements**
  - Added support for including child workflow data in webhook events
  - Introduced optional configuration for child workflow handling


services/workflows-service/src/case-management/case-management.service.ts
```diff
@@ -111,6 +111,16 @@ export class CaseManagementService {
             email: ubo.email,
             nationalId: ubo.nationalId,
             percentageOfOwnership: ubo.ownershipPercentage ?? ubo.percentageOfOwnership,
+            role: ubo.role,
```
Omri-Levy, line 114: Child workflows have no context schema, hence why validation does not fail on these fields.

services/workflows-service/src/common/types.ts
```diff
@@ -14,6 +14,11 @@ export const SubscriptionSchema = z.discriminatedUnion('type', [
       type: z.literal('webhook'),
       url: z.string().url(),
       events: z.array(z.string()),
+      config: z
+        .object({
+          withChildWorkflows: z.boolean().optional(),
+        })
+        .optional(),
```
Omri-Levy, line 21: Config was added on make's side and would be added on Postman's /run request.

alonp99 (approved):

PR #2927, opened 2025-01-01
fix: browser back button functionality in merchant report page (BAL-3296)

r4zendev:
## Changes

- Makes use of recently added optional `replace` property in the `useZodSearchParams` function to be passed to the corresponding `react-router-dom` function, which eliminates non-user-friendly history rewrites.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed an issue with the browser back button functionality on the merchant report page

- **Version Update**
  - Updated application version to 0.7.90

### DIFF
--- a/apps/backoffice-v2/package.json
+++ b/apps/backoffice-v2/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@ballerine/backoffice-v2",
-  "version": "0.7.89",
+  "version": "0.7.90",
   "description": "Ballerine - Backoffice",
   "homepage": "https://github.com/ballerine-io/ballerine",
   "type": "module",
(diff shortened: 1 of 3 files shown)


apps/backoffice-v2/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@ballerine/backoffice-v2",
-  "version": "0.7.89",
+  "version": "0.7.90",
```
MatanYadaev, line 3: The backoffice has CD regardless the version. We don't maintain versions for apps or services (like backoffice or worfklow-service), we only maintain versioning for packages. So this is redundant.

MatanYadaev (approved):

apps/backoffice-v2/package.json
rzssh: no harm though, right? can i still merge without adding another commit?

MatanYadaev: Sure

PR #2948, opened 2025-01-08
Bal 3307 - Improve Monthly Visitors Chart

shanegrouber:
- Add ChartContainer component for chart rendering
- Implement ChartTooltip and ChartTooltipContent components
- Enhance the WebsiteCredibility component to use the new Chart components

(your JSX structure is more nested than a Russian doll collection)

### DIFF
--- a/packages/ui/src/components/templates/report/components/WebsiteCredibility/WebsiteCredibility.tsx
+++ b/packages/ui/src/components/templates/report/components/WebsiteCredibility/WebsiteCredibility.tsx
@@ -85,11 +85,14 @@ export const WebsiteCredibility: FunctionComponent<{
   );
 
   const calculateTrend = (data: Array<{ label: string; value: string }>) => {
-    if (data.length < 2) return { direction: 'No trend data', percentage: 0 };
+    if (data.length < 2) {
+      return { direction: 'No trend data', percentage: 0 };
+    }
     const lastMonthValue = parseInt(data[data.length - 1]?.value ?? '0');
     const previousMonthValue = parseInt(data[data.length - 2]?.value ?? '0');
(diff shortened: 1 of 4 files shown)


packages/ui/src/components/templates/report/components/WebsiteCredibility/WebsiteCredibility.tsx
```diff
@@ -80,6 +84,19 @@ export const WebsiteCredibility: FunctionComponent<{
...
   );
 
+  const calculateTrend = (data: Array<{ label: string; value: string }>) => {
+    if (data.length < 2) return { direction: 'No trend data', percentage: 0 };
+    const lastMonthValue = parseInt(data[data.length - 1]?.value ?? '0');
+    const previousMonthValue = parseInt(data[data.length - 2]?.value ?? '0');
+    const percentageChange = ((lastMonthValue - previousMonthValue) / previousMonthValue) * 100;
+    const direction = lastMonthValue > previousMonthValue ? 'up' : 'down';
+    return { direction, percentage: Math.abs(percentageChange) };
+  };
```
tomer-shvadron:
```suggestion
  const calculateTrend = (data: Array<{ label: string; value: string }>) => {
    if (data.length < 2) {
      return { direction: 'No trend data', percentage: 0 };
    }
    
    const lastMonthValue = parseInt(data[data.length - 1]?.value ?? '0');
    const previousMonthValue = parseInt(data[data.length - 2]?.value ?? '0');
    const percentageChange = ((lastMonthValue - previousMonthValue) / previousMonthValue) * 100;
    const direction = lastMonthValue > previousMonthValue ? 'up' : 'down';
    
    return { direction, percentage: Math.abs(percentageChange) };
  };
```

tomer-shvadron (approved):

PR #2955, opened 2025-01-13
fix(logger): improve error logging structure

shanegrouber:
- Enhance logging of error properties
- Capture name and stack for better debugging

(Your error handling was so vague, it could moonlight as a fortune cookie)

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved error logging mechanism to capture more detailed error information.
	- Enhanced error object logging with comprehensive error properties.

### DIFF
--- a/services/workflows-service/src/common/utils/winston-logger/winston-logger.ts
+++ b/services/workflows-service/src/common/utils/winston-logger/winston-logger.ts
@@ -64,19 +64,16 @@ export class WinstonLogger implements IAppLogger {
     if (typeof error === 'string') {
       this.logger.error({ message: error, ...payload });
     } else {
-      const errorProperties = Object.getOwnPropertyNames(error).reduce(
-        (acc, key) => ({
-          ...acc,
-          [key]: (error as any)[key],
-        }),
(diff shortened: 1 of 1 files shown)


shanegrouber:
By default, Error objects in JavaScript are not enumerable, so their properties (message, name, stack) are excluded when logged via JSON.stringify.

e.g.
`> JSON.stringify(new Error('bad error'))`
`'{}'`

Even with **format.errors({ stack: true })** in Winston, passing an Error object directly
(e.g., `{ error: new Error('bad error'), payload: { test: 1 } }`) will log only the other properties, dropping the Error. 
This is because our logger statement encompasses the error with an additional payload `abstract error(error: unknown, payload: LogPayload): void;` (see [medium article](https://codewithdaniel.medium.com/how-to-fix-winston-hiding-error-objects-caecafa2ef99))

Also, found an [open winston issue](https://github.com/winstonjs/winston/issues/2527) where error stacks get omitted when used with splat() 

Solution is to manually spread the error object within the logger.

MatanYadaev (approved):

PR #2956, opened 2025-01-13
fix: replaced source of country codes

chesterkmr:
- replaced country codes source to `ballerine/common` in order to ensure that single source of country codes is used

## Summary by CodeRabbit

- **Refactor**
	- Updated country data retrieval method in `getCountries` function
	- Improved country code mapping and title generation
	- Streamlined import statements for country-related utilities

The changes primarily focus on optimizing how country information is processed and retrieved, with minimal impact on the overall functionality of the existing methods.

### DIFF
--- a/apps/kyb-app/src/helpers/countries-data.ts
+++ b/apps/kyb-app/src/helpers/countries-data.ts
@@ -1,9 +1,10 @@
+import { countryCodes } from '@ballerine/common';
+import { State } from 'country-state-city';
 import isoCountries from 'i18n-iso-countries';
 import enCountries from 'i18n-iso-countries/langs/en.json';
 import cnCountries from 'i18n-iso-countries/langs/zh.json';
 import nationalities from 'i18n-nationality';
 import enNationalities from 'i18n-nationality/langs/en.json';
-import { State } from 'country-state-city';
 import { TFunction } from 'i18next';
 
 isoCountries.registerLocale(enCountries);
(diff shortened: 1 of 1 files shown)


apps/kyb-app/src/helpers/countries-data.ts
```diff
@@ -21,11 +22,10 @@ const unsupportedNationalities = {
 
 export const getCountries = (lang = 'en') => {
   const language = languageConversionMap[lang as keyof typeof languageConversionMap] ?? lang;
-  const countries = isoCountries.getNames(language, { select: 'official' });
 
-  return Object.entries(countries).map(([isoCode, title]) => ({
+  return countryCodes.map(isoCode => ({
```
tomer-shvadron, line 26: Could you explain what was the reason for the change? Also, why do we have so many different sources for country codes?

tomer-shvadron (approved):

apps/kyb-app/src/helpers/countries-data.ts
chesterkmr: > Could you explain what was the reason for the change?

During the investigation of the failing workflow submissions, I discovered that the issue was caused by unsupported country codes in one of our services. For example, our service expected the country code GB for the United Kingdom, while KYB sent UK. This mismatch resulted in exceptions and ultimately led to the failure of the workflow.

> Also, why do we have so many different sources for country codes?

Having multiple sources for country codes is a clear oversight. We should standardize and use a single, unified source for country codes across all services. Ideally, this source should be provided by the common package to ensure consistency and prevent similar issues in the future.

tomer-shvadron: I agree with your message, can you add that change to this PR?

PR #2960, opened 2025-01-14
feat(monitoring): adds loading state for a single merchant record (BAL-3359)

r4zendev:
Screenshot 2025-01-14 at 21 06 13

## Summary by CodeRabbit

- **New Features**
	- Added loading skeletons to the Merchant Monitoring Business Report page
	- Improved user experience by providing visual feedback during data fetching

- **Chores**
	- Updated subproject commit in workflows service data migrations


alonp99:
@r4zendev can you do something with the black tag near the status? 🙏

apps/backoffice-v2/src/pages/MerchantMonitoringBusinessReport/hooks/useMerchantMonitoringBusinessReportLogic/useMerchantMonitoringBusinessReportLogic.tsx
```diff
@@ -24,9 +24,9 @@ const statusToBadgeData = {
 
 export const useMerchantMonitoringBusinessReportLogic = () => {
   const { businessReportId } = useParams();
-  const { data: businessReport } = useBusinessReportByIdQuery({
-    id: businessReportId ?? '',
-  });
+  const { data: businessReport, isFetching: isFetchingBusinessReport } = useBusinessReportByIdQuery(
```
Omri-Levy, line 27: Are we sure we prefer isFetching over isLoading for this task?

rzssh: We could use `isLoading` to prevent the loading state from appearing if a refetch will happen or if the query results become invalidated, but that's a UX question. I think showing the loading state in all occasions when the fetch is happening is better than hotswapping data on a user's screen without letting him know that the load is happening. We could do both.

Omri-Levy: We can change it post merge if needed.

Omri-Levy (approved):

PR #2962, opened 2025-01-14
feat(monitoring): preserves scroll position on a data table (BAL-3248)

r4zendev:
https://github.com/user-attachments/assets/3f627658-4419-4940-8488-b9f7a94201c7

This PR includes:
- fix: tailwind loading twice / overriding styles
- fix: dynamic height of the main element based on viewport size to prevent user from scrolling out of the main shell (with back button etc.)
- fix: navbar now remembers last state of a merchant monitoring tab to route the user correctly
- feat: scroll state is persisted for tables

## Summary by CodeRabbit

- **New Features**
	- Introduced a new `usePersistentScroll` hook to maintain scroll position across page reloads.
	- Enhanced `DataTable` and `ScrollArea` components with improved scroll event handling and ref management.
	- Added a new constant for dynamic navigation parameters in the "Merchant Monitoring" section.

- **Refactor**
	- Updated `ScrollArea` component type parameters and props structure.
	- Simplified `DataTable` component with forward ref implementation.
	- Modified `MerchantMonitoringSearchSchema` to ensure default date values for search parameters.

- **Bug Fixes**
	- Adjusted the height of the `ScrollArea` component in the `MerchantMonitoringBusinessReport` for better layout.


apps/backoffice-v2/src/common/hooks/useRadixScrollBoundaries/useRadixScrollBoundaries.tsx
```diff
@@ -0,0 +1,32 @@
+import { useEffect, useRef } from 'react';
+
+export const useRadixScrollBoundaries = () => {
```
Omri-Levy: Consider a name that does not reference Radix, it seems like something along the lines of useScrollTo or useScrollRestoration etc would be more descriptive.

apps/backoffice-v2/src/common/hooks/useRadixScrollBoundaries/useRadixScrollBoundaries.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+  const resetScrollPosition = () => {
+    localStorage.removeItem('scrollPosition');
+  };
+
+  const restoreScrollPosition = () => {
+    const savedPosition = localStorage.getItem('scrollPosition');
+
+    if (savedPosition && scrollAreaRef.current) {
+      if (scrollAreaRef.current) {
```
Omri-Levy: This if condition is redundant.

packages/ui/src/components/organisms/DataTable/DataTable.tsx
```diff
@@ -327,3 +335,8 @@ export const DataTable = <TData extends RowData, TValue = any>({
     </div>
   );
 };
+
+const forward = React.forwardRef as <T, P = NonNullable<unknown>>(
+  render: (props: P, ref: React.Ref<T>) => React.ReactNode | null,
+) => (props: P & React.RefAttributes<T>) => React.ReactNode | null;
```
Omri-Levy: ReactNode includes null already 🙂

apps/backoffice-v2/src/common/hooks/useRadixScrollBoundaries/useRadixScrollBoundaries.tsx
```diff
@@ -0,0 +1,32 @@
...
+        scrollAreaRef.current.scrollTop = parseInt(savedPosition, 10);
```
Omri-Levy: Not a must, consider using the scroll method over mutating scrollTop. It's subjective unless we want to use scroll behavior.

apps/backoffice-v2/src/common/hooks/useRadixScrollBoundaries/useRadixScrollBoundaries.tsx
rzssh: It used to only work with Radix until I made it more general. Forgot to update the name.

apps/backoffice-v2/src/pages/MerchantMonitoring/schemas.ts
```diff
@@ -125,6 +126,6 @@ export const MerchantMonitoringSearchSchema = BaseSearchSchema.extend({
       ],
     )
     .catch('All'),
-  from: z.string().date().optional(),
-  to: z.string().date().optional(),
+  from: z.string().date().default(dayjs().subtract(30, 'day').format('YYYY-MM-DD')),
+  to: z.string().date().default(dayjs().format('YYYY-MM-DD')),
```
Omri-Levy: Prefer catch to default for search query params. This way it covers both invalid input and the default state.

Omri-Levy (approved):

PR #2964, opened 2025-01-14
feat(monitoring): adds exhaustive check for action before deboarding a merchant (BAL-3343)

r4zendev:
Repro steps:
- Click on the `Turn Monitoring Off` option from the options list in the top-right corner of the page
- Notice a confirmation modal popping up letting user to select the "deboarding reason"

Screenshot 2025-01-14 at 20 47 11

See in action:

https://github.com/user-attachments/assets/2607bcef-1858-468b-a0c9-11d9994e65f7

Includes changes to store user-entered data on the backend.

## Summary by CodeRabbit

- **New Features**
  - Added a dialog component for managing merchant monitoring status
  - Introduced more detailed monitoring state changes with reasons
  - Enhanced form validation for monitoring state updates

- **Improvements**
  - Streamlined monitoring toggle functionality
  - Improved error handling for monitoring state changes
  - Added user feedback through toast notifications

- **Bug Fixes**
  - Implemented more robust error checking for merchant monitoring actions


apps/backoffice-v2/src/pages/MerchantMonitoringBusinessReport/MerchantMonitoringBusinessReport.page.tsx
```diff
@@ -68,20 +124,113 @@ export const MerchantMonitoringBusinessReport: FunctionComponent = () => {
...
-                      businessReport?.monitoringStatus
-                        ? turnOngoingMonitoringOff(businessReport?.merchantId)
-                        : turnOngoingMonitoringOn(businessReport?.merchantId);
+                      if (!businessReport?.merchantId) {
+                        throw new Error('Merchant ID is missing');
+                      }
+
+                      turnOngoingMonitoringOn({ merchantId: businessReport.merchantId });
+                      setIsDeboardModalOpen(false);
+                      setIsDropdownOpen(false);
```
Omri-Levy: Should move to the onSuccess of the turnOngoingMonitoringOn mutation

services/workflows-service/src/business/business.controller.external.ts
```diff
@@ -152,26 +155,15 @@ export class BusinessControllerExternal {
...
-              enabled,
-            },
-          },
-        };
+    const updatedMetadata = _.merge({}, metadata, {
+      featureConfig: {
+        [FEATURE_LIST.ONGOING_MERCHANT_REPORT]: {
+          enabled: data.state === 'on',
+          reason: data.reason ?? null,
+          userReason: data.userReason ?? null,
```
tomer-shvadron, line 163: Why do we need both `reason` and `userReason`?

tomer-shvadron (approved):
LGTM, please make sure Nitzan and Adi and Robbie take a look at the design and texts

apps/backoffice-v2/src/pages/MerchantMonitoringBusinessReport/hooks/useMerchantMonitoringBusinessReportLogic/useMerchantMonitoringBusinessReportLogic.tsx
```diff
@@ -33,8 +50,75 @@ export const useMerchantMonitoringBusinessReportLogic = () => {
     noteableType: 'Report',
   });
 
-  const turnMonitoringOnMutation = useTurnMonitoringOnMutation();
-  const turnMonitoringOffMutation = useTurnMonitoringOffMutation();
+  const [isDeboardModalOpen, setIsDeboardModalOpen] = useState(false);
+  const [isDropdownOpen, setIsDropdownOpen] = useState(false);
```
Omri-Levy: Consider the useToggle hook

apps/backoffice-v2/src/pages/MerchantMonitoringBusinessReport/hooks/useMerchantMonitoringBusinessReportLogic/useMerchantMonitoringBusinessReportLogic.tsx
```diff
@@ -33,8 +50,75 @@ export const useMerchantMonitoringBusinessReportLogic = () => {
...
+      (data.reason === 'other' && !data.userReason) ||
+      (data.userReason && data.userReason?.length < 5)
+    ) {
+      form.setError('userReason', {
+        type: 'minLength',
+        message: 'Please provide a reason of at least 5 characters',
+      });
+
+      return;
+    }
```
Omri-Levy: Can't this logic be covered by the form's zod schema?

services/workflows-service/src/business/dtos/business-monitoring.patch.dto.ts
```diff
@@ -0,0 +1,18 @@
...
+import { IsOptional, IsString } from 'class-validator';
+
+export class BusinessMonitoringPatchDto {
+  @ApiProperty({ type: String, required: true })
+  @IsString()
+  state!: 'on' | 'off';
+
+  @ApiProperty({ type: String, required: false })
+  @IsOptional()
+  @IsString()
```
Omri-Levy, line 12: Consider @IsIn from class validator

apps/backoffice-v2/src/pages/MerchantMonitoringBusinessReport/fetchers.ts
```diff
@@ -4,17 +4,32 @@ import { Method } from '@/common/enums';
...
 import { handleZodError } from '@/common/utils/handle-zod-error/handle-zod-error';
 
+export type TurnOngoingMonitoringBody = z.infer<typeof TurnOngoingMonitoringBodySchema>;
+export const TurnOngoingMonitoringBodySchema = z
+  .object({
+    state: z.string().optional(),
+    reason: z.string().optional(),
+    userReason: z.string().optional(),
+  })
+  .optional();
```
Omri-Levy: In what state body is optional or an empty object? (all properties optional)

services/workflows-service/src/business/business.controller.external.ts
rzssh: reason is an option from the select input in the form. userReason is a reason provided by the user in text form (additional details input)

apps/backoffice-v2/src/pages/MerchantMonitoringBusinessReport/hooks/useMerchantMonitoringBusinessReportLogic/useMerchantMonitoringBusinessReportLogic.tsx
rzssh: It's not easy to implement with zod literals (assigned to object properties directly) without using `custom`/`refine` etc. because the `userReason` property validator relies on the actual runtime value present for `reason` also zod resolver here is supposed to include all the possible values for fields. making reason required would disallow setting it to undefined and showing placeholder in the select field. zod schema should be loose here.

Omri-Levy: Using refine like in the password confirmation example in the documentation seems like the same use-case, I think it's important to keep the validation in one place as well. (in the schema) https://zod.dev/?id=customize-error-path

Can you please clarify why we'd need to make reason required?

rzssh: Don't we want to store the reason that the user have chosen to be able to categorize them? If not the case, we can allow just simply clicking the button without any additional validation at all, since that's just for the users of the backoffice app. Do we want to make the properties optional and allow to turn off without specifying the reason?

rzssh: TIL `z.object().refine()` is actually really nice for handling conditional forms. Please resolve if that form is what we're looking for or if we should have `reason` as a required param when turning off the monitoring.

PR #2968, opened 2025-01-16
fix: corrected home page merchants metrics source of truth (BAL-3396, BAL-3397)

r4zendev:
## Summary by CodeRabbit

- **New Features**
	- Enhanced business report metrics to include count of unmonitored merchants.
	- Introduced a method to count business records based on specific criteria.
	- Added a new method for calculating merchant monitoring metrics with detailed insights.

- **Improvements**
	- Improved ongoing monitoring feature state management with clearer logic and timestamp tracking.
	- Streamlined code readability in business controller.

### DIFF
--- a/services/workflows-service/src/business-report/business-report.controller.external.ts
+++ b/services/workflows-service/src/business-report/business-report.controller.external.ts
@@ -44,6 +44,7 @@ import {
 } from '@/business-report/dtos/business-report-metrics.dto';
 import { BusinessReportMetricsDto } from './dtos/business-report-metrics-dto';
 import { FEATURE_LIST, TCustomerWithFeatures } from '@/customer/types';
+import dayjs from 'dayjs';
 
 @ApiBearerAuth()
 @swagger.ApiTags('Business Reports')
@@ -168,7 +169,28 @@ export class BusinessReportControllerExternal {
   ) {
     const { id: customerId } = await this.customerService.getByProjectId(currentProjectId);
 
-    return await this.merchantMonitoringClient.getMetrics({ customerId, from, to });
+    const unmonitoredMerchants = await this.prismaService.business.count({
+      where: {
+        projectId: currentProjectId,
+        metadata: {
+          path: ['featureConfig', FEATURE_LIST.ONGOING_MERCHANT_REPORT, 'disabledAt'],
+          not: 'null',
+          ...(from && { gt: dayjs(from).toDate().getTime() }),
+          ...(to && { lte: dayjs(to).toDate().getTime() }),
+        },
+      },
+    });
+
+    const merchantMonitoringMetrics = await this.merchantMonitoringClient.getMetrics({
+      customerId,
+      from,
+      to,
+    });
+
+    return {
+      ...merchantMonitoringMetrics,
+      removedMerchantsCount: unmonitoredMerchants,
+    };
   }
 
   @common.Post()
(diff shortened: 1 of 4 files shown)


services/workflows-service/src/business-report/business-report.controller.external.ts
```diff
@@ -168,7 +169,28 @@ export class BusinessReportControllerExternal {
   ) {
     const { id: customerId } = await this.customerService.getByProjectId(currentProjectId);
 
-    return await this.merchantMonitoringClient.getMetrics({ customerId, from, to });
+    const unmonitoredMerchants = await this.prismaService.business.count({
+      where: {
+        projectId: currentProjectId,
+        metadata: {
+          path: ['featureConfig', FEATURE_LIST.ONGOING_MERCHANT_REPORT, 'disabledAt'],
+          not: 'null',
```
Omri-Levy: This is not supposed to be null literal?

Omri-Levy (approved):

services/workflows-service/src/business-report/business-report.controller.external.ts
rzssh: putting anything other than string there is a type error + it works, so I assume it's some Prisma's internal logic to transform it to stringified values

MatanYadaev (approved):

services/workflows-service/src/business/business.service.ts
```diff
@@ -66,6 +68,105 @@ export class BusinessService {
...
+    unmonitoredMerchants: number;
+  }> {
+    // Metrics are currently mostly requested by month
+    if (!from) {
+      from = dayjs().startOf('month').toISOString();
+    }
+
+    if (!to) {
+      to = dayjs(from).add(1, 'month').toISOString();
+    }
```
tomer-shvadron: This initializations can be moved to lines 74-75

services/workflows-service/src/business/business.service.ts
```diff
@@ -66,6 +68,105 @@ export class BusinessService {
...
+                TCustomerFeaturesConfig & { disabledAt: number | null | undefined }
+              >;
+            }
+          )?.featureConfig?.[FEATURE_LIST.ONGOING_MERCHANT_REPORT]?.disabledAt,
+        );
+
+      return (
+        disabledAt === null ||
+        (b.metadata === null && features?.ONGOING_MERCHANT_REPORT?.options?.runByDefault)
+      );
```
tomer-shvadron: I'm pretty sure this would work as well, I think it's a bit more readable, WDYT?
```suggestion
      const isEnabled = b?.metadata?.featureConfig?.[FEATURE_LIST.ONGOING_MERCHANT_REPORT]?.enabled,
        );

      return (
        isEnabled === false ||
        (isEnabled === undefined && features?.ONGOING_MERCHANT_REPORT?.options?.runByDefault)
      );
```

services/workflows-service/src/business/business.service.ts
```diff
@@ -66,6 +68,105 @@ export class BusinessService {
...
+    }).length;
+
+    const addedMerchantsCount = await this.repository.count(
+      {
+        where: {
+          OR: [
+            {
+              metadata: {
+                path: ['featureConfig', FEATURE_LIST.ONGOING_MERCHANT_REPORT, 'disabledAt'],
```
tomer-shvadron: similar suggestion, change the use of disabledAt to enabled === false or enabled === undefined and runByDefault is true

services/workflows-service/src/business/business.service.ts
```diff
@@ -66,6 +68,105 @@ export class BusinessService {
...
+      projectIds,
+    );
+
+    const unmonitoredMerchants = await this.repository.count(
+      {
+        where: {
+          OR: [
+            {
+              metadata: {
+                path: ['featureConfig', FEATURE_LIST.ONGOING_MERCHANT_REPORT, 'disabledAt'],
```
tomer-shvadron, line 132: same :)

tomer-shvadron (approved):

services/workflows-service/src/business/business.service.ts
rzssh: unapplicable, since we need to filter by date depending on the value of `disabledAt` field in this case.

PR #2978, opened 2025-01-20
fix: fixed logo refetch on re-render

chesterkmr:
## Summary by CodeRabbit

- **Refactor**
	- Updated internal logic for handling image loading in the Logo component to improve reference management.


apps/kyb-app/src/components/layouts/AppShell/Logo.tsx
```diff
@@ -16,14 +17,16 @@ const prefetchImage = (url: string) =>
...
   useEffect(() => {
     if (!onLoad) {
       return;
     }
 
     // Using race here in case if image is corrupted or load takes to long we don't want to lock stepper breadcrumbs forever.
-    Promise.race([prefetchImage(logoSrc), fallback(3000)]).then(onLoad);
-  }, [logoSrc, onLoad]);
+    Promise.race([prefetchImage(logoSrc), fallback(3000)]).then(onLoadRef.current);
+  }, [logoSrc, onLoadRef]);
```
tomer-shvadron, line 29: Should this line be:
```suggestion
  }, [logoSrc, onLoadRef.current]);
```

tomer-shvadron (approved):

apps/kyb-app/src/components/layouts/AppShell/Logo.tsx
ArtemLavrentii, line 29: @tomer-shvadron Wouldn't that bring back the bug ? I think best way to handle would be something like

```suggestion
    Promise.race([prefetchImage(logoSrc), fallback(3000)]).then((value) => { if (onLoadRef.current) onLoadRef.current(value)) });
  }, [logoSrc, onLoadRef]);
```
(cleanup is needed, just an idea suggestion)

chesterkmr: Nope, it will trigger useEffect each time prop updates, that was the case of bug. So only using link to callback and not callback itself as dependency.

chesterkmr: Nope. It solves bug. Ref object is same each rerender whenever current is being updated. What you suggesting is pretty much safe check but its not needed in case with`then` since it wont do anything if it will recieve `undefined`

ArtemLavrentii: I was referring to proposed above changes in the first half :)

This isn't just a safe-check - onLoadRef will be updated with latest onLoad callback, and this will make sure that we always call most-recent version of the callback ( and thats something that I would expect from the callbacks ).

PR #2994, opened 2025-01-28
fix: fixed final submission failure without directors and ubos

chesterkmr:
Fixes bug when final-submission fails on flows that doesnt have directors or ubos.

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved handling of empty director and UBO arrays during workflow submission
	- Prevented sending empty arrays in payload, setting them to `undefined` when appropriate

### DIFF
--- a/services/workflows-service/src/collection-flow/controllers/collection-flow.controller.ts
+++ b/services/workflows-service/src/collection-flow/controllers/collection-flow.controller.ts
@@ -167,7 +167,7 @@ export class CollectionFlowController {
               ...director,
             };
           },
-        ),
+        ) || [],
       );
 
       const ubos = await Promise.all(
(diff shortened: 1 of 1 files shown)


tomer-shvadron (approved):

services/workflows-service/src/collection-flow/controllers/collection-flow.controller.ts
```diff
@@ -199,8 +199,8 @@ export class CollectionFlowController {
               entity: {
                 data: {
                   additionalInfo: {
-                    directors,
-                    ubos,
+                    directors: directors?.length ? directors : undefined,
+                    ubos: ubos?.length ? ubos : undefined,
```
Omri-Levy, line 203: Does a length of zero behave like we expect?

Omri-Levy, line 203: Don't have to do it if there's no need, just might be safer.
```suggestion
                    ...(directors && { directors }),
                    ...(ubos && { ubos }),
```

PR #3025, opened 2025-02-04
feat: get full access popup & customer PUT endpoint (BAL-3495)

r4zendev:
image

## Summary by CodeRabbit

- **New Features**
  - Introduced an interactive prompt card on primary pages that encourages users to unlock additional features via a scheduling link. The card includes a clear title, brief description, a clickable button, and an easy-to-use close option.
  - Enhanced customer management capabilities to support more efficient updates, contributing to a smoother user experience.
  - Added a new property to customer configurations to control the visibility of the full access popup.


apps/backoffice-v2/src/common/components/molecules/GetFullAccessCard/GetFullAccessCard.tsx
```diff
@@ -0,0 +1,67 @@
...
+import { DataListImage } from './DataListImage';
+
+export const GetFullAccessCard = () => {
+  const [isHidden, setIsHidden] = useState(false);
+
+  const { data: customer, isLoading } = useCustomerQuery();
+
+  if (env.VITE_ENVIRONMENT_NAME === 'production' || isLoading || customer?.config?.hasFullAccess) {
+    return null;
+  }
```
MatanYadaev: Why did you prefer making this condition in the component, instead of doing it outside? I'm just curious to understand pros/cons, I don't have a preference.

apps/backoffice-v2/src/domains/customer/fetchers.ts
```diff
@@ -38,11 +38,13 @@ const CustomerSchema = z.object({
       isMerchantMonitoringEnabled: z.boolean().default(false),
       isExample: z.boolean().default(false),
       isDemo: z.boolean().default(false),
+      hasFullAccess: z.boolean().default(false),
```
MatanYadaev: As mentioned on Slack:
* Make it true by false. We don't want to display the popup by default.
* Rename to something like `showGetFullAccessPopup`/`showTalkToUsPopup`

MatanYadaev (approved):

apps/backoffice-v2/src/common/components/molecules/GetFullAccessCard/GetFullAccessCard.tsx
rzssh: in my opinion, the logic should belong to the component that requires it. doing it outside would mean repeating the same logic in places where this component will be reused (and the logic will obviously be identical). so i made the component a bit smarter than a regular view component should be to avoid unnecessary repetitions and complexity.

PR #3042, opened 2025-02-11
fix(CaseOverview): UBO risk indicator title

shanegrouber:
- Ensure 'ubos' case returns the correct tab name

The notion rules had 'Domain' as 'UBOs'
``` ts
const domain = 'UBOs'
console.log(camelCase(domain))
> 'ubOs'
```

This does not match the mapping in `TabToLabel` because of
```ts 
title: TabToLabel[tab as keyof typeof TabToLabel] ?? titleCase(domain ?? '')
```
so it would default back to use:
``` ts
console.log(titleCase(camelCase(domain)))
> 'UB Os'
```

## Summary by CodeRabbit

- **New Features**
  - Improved the consistency of tab names in the Case Overview, ensuring certain domain labels display as intended.
  - Enhanced date handling logic in merchant monitoring to ensure end dates are inclusive.

- **Chores**
  - Streamlined internal commit processing by excluding unnecessary lock file changes from version control diffs.


scripts/auto-commit.js
```diff
@@ -75,7 +75,7 @@ async function generateCommitMessage(diff) {
 async function main() {
   try {
     // Get git diff
-    const diff = execSync('git diff --cached').toString();
+    const diff = execSync('git diff --cached -- . ":(exclude)pnpm-lock.yaml"').toString();
```
alonp99, line 78: 👑

apps/backoffice-v2/src/pages/Entity/components/Case/components/CaseOverview/CaseOverview.tsx
```diff
@@ -28,7 +28,14 @@ export const CaseOverview = ({ processes }: { processes: string[] }) => {
       workflow?.context?.pluginsOutput?.risk_evaluation?.riskIndicatorsByDomain ??
       {},
   )?.map(([domain, riskIndicators]) => {
-    const tab = camelCase(domain);
+    const getTab = (domain: string) => {
```
alonp99: Why do we camel case the domains? 🤔

shanegrouber: In order to be consistent with our mapping - so when we do lookups to get the title, it will be a consistent lookup.
Made a change, I think we can just directly make all domains lowercase.

```ts
const TabToLabel: {
    readonly summary: "Summary";
    readonly companyInformation: "Company";
    readonly storeInformation: "Store";
    readonly documents: "Documents";
    readonly ubos: "UBOs";
    readonly associatedCompanies: "Associated Companies";
    readonly directors: "Directors";
    readonly monitoringReports: "Monitoring Reports";
    readonly customData: "Custom Data";
}
```

tomer-shvadron (approved):

PR #3043, opened 2025-02-11
feat: document tracker (BAL-3459, BAL-3435)

r4zendev:
image
image
image
image


apps/backoffice-v2/src/common/components/molecules/DocumentTracker/DocumentTracker.tsx
```diff
@@ -0,0 +1,143 @@
...
+  // Feature flag
+  // const { data: customer, isLoading: isLoadingCustomer } = useCustomerQuery();
+
+  // if (isLoadingCustomer) {
+  //   return <Loader2 />;
+  // }
+
+  // if (!isLoadingCustomer && !customer?.features?.isDocumentOcrEnabled) {
+  //   return null
+  // }
```
rzssh: can we use `isDocumentOcrEnabled` for this matter? or should we create a new feature flag?

Omri-Levy: We need to create a new feature flag and ensure its part of the case overview component and the other feature flag related to the case overview.

apps/backoffice-v2/src/common/components/molecules/DocumentTracker/hooks/useDocumentTracker.tsx
```diff
@@ -0,0 +1,84 @@
...
+      return {
+        leftIcon: null,
+        text: 'Document ID is missing',
+        itemClassName: 'p-1',
+      };
+    }
+
+    return {
+      leftIcon: (
+        <div
```
Omri-Levy: Please use our button component or the native HTML button element for clickable UI at least of similar nature. 🙏🏼

apps/backoffice-v2/src/common/components/molecules/DocumentTracker/hooks/useDocumentTracker.tsx
```diff
@@ -0,0 +1,84 @@
...
+          onClick={() => {
+            if (doc.status === 'unprovided') {
+              setSelectedIdsToRequest(prev => [...prev, documentId]);
+            }
+
+            if (selectedIdsToRequest.includes(documentId)) {
+              setSelectedIdsToRequest(prev => prev.filter(id => id !== documentId));
+            }
```
Omri-Levy: I think it would be safer to switch the two statements around.

apps/backoffice-v2/src/common/components/molecules/DocumentTracker/hooks/useDocumentTracker.tsx
```diff
@@ -0,0 +1,94 @@
...
+          text: 'Document ID is missing',
+          itemClassName: 'p-1',
+        };
+      }
+
+      return {
+        leftIcon: (
+          <button
+            className={ctw('cursor-default', status === 'unprovided' && 'cursor-pointer')}
+            role="button"
```
Omri-Levy: Did you mean `type` button?

apps/backoffice-v2/src/domains/documents/hooks/schemas/document.ts
```diff
@@ -0,0 +1,32 @@
...
+    version: z.string(),
+  }),
+  entity: z.object({
+    entityType: z.string(),
+    id: z.string(),
+    companyName: z.string().optional(),
+    firstName: z.string().optional(),
+    lastName: z.string().optional(),
+  }),
+});
```
Omri-Levy, line 23: Appreciate it. 🙏🏼

apps/backoffice-v2/src/domains/workflow-definitions/fetchers.ts
```diff
@@ -27,6 +27,7 @@ export const WorkflowDefinitionConfigSchema = z
     isAssociatedCompanyKybEnabled: z.boolean().default(false),
     isCaseOverviewEnabled: z.boolean().default(false),
     isCaseRiskOverviewEnabled: z.boolean().default(false),
+    isDocumentTrackingEnabled: z.boolean().default(false),
```
Omri-Levy: I think it's for the best that the feature flag would match the UI component it controls.
```suggestion
    isDocumentTrackerEnabled: z.boolean().default(false),
```

packages/ui/src/components/molecules/AccordionCard/AccordionCard.Item.tsx
```diff
@@ -50,17 +50,19 @@ export const AccordionCardItem: FunctionComponent<AccordionCardItemProps> = ({
...
-                  {text}
-                  {rightIcon}
-                </li>
-              ))}
+              subitems.map(({ leftIcon, text, rightIcon, itemClassName }, index) => {
+                return (
+                  <li
+                    {...liProps}
+                    className={ctw(`flex items-center gap-x-2`, itemClassName, liProps?.className)}
+                    key={typeof text === 'string' ? `${text}-${index}` : index}
```
Omri-Levy: Do we expect text to repeat itself?

services/workflows-service/src/document/document.controller.external.ts
```diff
@@ -108,6 +108,60 @@ export class DocumentControllerExternal {
...
+    response: Type.Any(),
+  })
+  async getDocumentsByWorkflowId(
+    @Param('workflowId') workflowId: string,
+    @CurrentProject() projectId: TProjectId,
+  ) {
+    return await this.documentService.getDocumentTrackerByWorkflowId(projectId, workflowId);
+  }
+
+  @Post('request')
```
Omri-Levy: Thoughts? Or some similar name? 🙂 
```suggestion
  @Post('request-upload')
```

services/workflows-service/src/document/document.service.ts
```diff
@@ -439,6 +435,17 @@ export class DocumentService {
     return result;
   }
 
+  async requestDocumentsByIds(projectId: TProjectId, documentIds: string[]) {
+    // TODO call email flow for given documents
```
Omri-Levy, line 439: This is where I'll continue from?

services/workflows-service/src/document/document.service.ts
```diff
@@ -269,13 +269,9 @@ export class DocumentService {
...
-    workflowRuntimeDataId: string,
-  ) {
-    const uiDefinition = await this.uiDefinitionService.getByWorkflowDefinitionId(
-      workflowDefinitionId,
+  async getDocumentTrackerByWorkflowId(projectId: TProjectId, workflowId: string) {
+    const uiDefinition = await this.uiDefinitionService.getByRuntimeId(
+      workflowId,
       'collection_flow',
       [projectId],
     );
```
Omri-Levy, line 277: Not a blocker but couldn't we get both the workflow and the ui definition in a single query at the end?

apps/backoffice-v2/src/common/components/molecules/DocumentTracker/interfaces.ts
```diff
@@ -0,0 +1,13 @@
...
+export interface IDocumentTrackerProps {
+  workflow: TWorkflowById;
+  plugins: Array<
+    | NonNullable<NonNullable<TWorkflowDefinitionById['extensions']>['apiPlugins']>[number]
+    | NonNullable<
+        NonNullable<TWorkflowDefinitionById['extensions']>['childWorkflowPlugins']
+      >[number]
+    | NonNullable<NonNullable<TWorkflowDefinitionById['extensions']>['commonPlugins']>[number]
+  >;
+}
```
Omri-Levy: These are not needed right? It seems at least workflow is not in use and if yes maybe we can pass just the properties in use and not the full workflow.

apps/backoffice-v2/src/common/components/molecules/DocumentTracker/hooks/useDocumentTracker.tsx
```diff
@@ -0,0 +1,94 @@
...
+        return {
+          leftIcon: null,
+          text: 'Document ID is missing',
+          itemClassName: 'p-1',
+        };
+      }
+
+      return {
+        leftIcon: (
+          <button
```
Omri-Levy, line 45: Was the idea that the icon should serve as the button instead of a dropdown because the upload feature was pushed to next sprint?

rzssh: not necessarily because of that, but rather because it's more intuitive to the user, small detail. i can either add a dropdown now to have it prepared for the upload feature that is coming next, or we can add it later. no matter when we add it, it should also include the option to add the document to the requested state.

apps/backoffice-v2/src/common/components/molecules/DocumentTracker/hooks/useDocumentTracker.tsx
rzssh: yep

services/workflows-service/src/document/document.service.ts
rzssh: i guess so, not sure who's responsible for that part of work.

Omri-Levy (approved):

PR #3046, opened 2025-02-11
feat(bal-3616): added demo account config option

ArtemLavrentii:
## Summary by CodeRabbit

- **New Features**
  - Introduced a demo account option within customer and workflow settings, allowing enhanced configuration flexibility.

- **Chores**
  - Updated development dependencies and improved package management settings for a more reliable build process.

### DIFF
--- a/apps/backoffice-v2/src/domains/customer/fetchers.ts
+++ b/apps/backoffice-v2/src/domains/customer/fetchers.ts
@@ -39,12 +39,14 @@ const CustomerSchema = z.object({
       isExample: z.boolean().default(false),
       isDemo: z.boolean().default(false),
       showFullAccessPopup: z.boolean().default(false),
+      demoAccount: z.boolean().default(false),
     })
     .nullable()
     .default({
       isMerchantMonitoringEnabled: false,
       isExample: false,
       showFullAccessPopup: false,
+      demoAccount: false,
     }),
 });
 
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -80,19 +80,42 @@
...
-    "openai": "^4.70.3"
+    "openai": "^4.70.3",
+    "prettier": "^2.8.7"
   },
   "dependencies": {
     "concurrently": "^7.6.0",
     "cross-env": "^7.0.3",
     "wait-on": "^7.0.1"
+  },
+  "pnpm": {
```
ArtemLavrentii, line 98: This is a new feature of pnpm@10, without this modules like bcrypt won't install properly.

tomer-shvadron (approved):

apps/backoffice-v2/src/domains/customer/fetchers.ts
```diff
@@ -39,12 +39,14 @@ const CustomerSchema = z.object({
       isExample: z.boolean().default(false),
       isDemo: z.boolean().default(false),
       showFullAccessPopup: z.boolean().default(false),
+      demoAccount: z.boolean().default(false),
```
MatanYadaev: `is` prefix will make the flag more consistent with the other flags we have

```suggestion
      isDemoAccount: z.boolean().default(false),
```

package.json
MatanYadaev: Can you explain? Maybe share in Slack

MatanYadaev (approved):

package.json
alonp99: shouldn't it be in the project package json?

ArtemLavrentii: Because workspaces are used this options affects every sub-project of workspace.

PR #3052, opened 2025-02-13
feat: create report modal (BAL-3607)

r4zendev:
## Changes
- Introduces popup for creating reports
- Adjusts customer config that is being sent to the client to include demo customer data.
- Small refactoring initiatives, such as creating merchant monitoring module to share the client for customer module and allow gathering data required for demo accounts in one request (customers/by-currect-project-id).

## Examples

image
image

image
image

## Summary by CodeRabbit

- **New Features**
  - Introduced a business report card that clearly indicates remaining reports, demo days, and account status.
  - Launched an interactive report creation dialog with form validations, loading indicators, and success feedback.
  - Enhanced the monitoring page with an intuitive button for initiating report submissions.
  - Added demo access details to enhance customer information visibility.

- **Refactor**
  - Updated component rendering conditions to leverage refined demo access details.
  - Reorganized dependencies and modules to streamline Merchant Monitoring functionality.
  - Improved error handling and response structure in customer-related methods.


services/workflows-service/src/global.d.ts
```diff
@@ -10,6 +10,8 @@ declare module '@prisma/client' {
   import type { TCustomerConfig, TCustomerSubscription } from '@/customer/schemas/zod-schemas';
   export * from '@prisma/client/index';
 
+  // FIXME: this is a problem
```
MatanYadaev, line 13: What's the problem? :)

services/workflows-service/src/customer/customer.service.ts
```diff
@@ -1,20 +1,46 @@
...
+
+    const businessReportsCount = await this.merchantMonitoringClient.count({
+      customerId,
+    });
+
+    return {
+      totalReports: businessReportsCount,
+      maxBusinessReports: config.maxBusinessReports ?? 10,
+      showFullAccessPopup: config.showFullAccessPopup ?? false,
+      // FIXME: should count from the first login date / should be passed on creation ?
```
MatanYadaev, line 39: I recommend you to discuss it with @ArtemLavrentii, as he's working on the customer's email onboarding and signup.
We wanted it to start counting since the user was signed it first time. But we can also compromise on the customer creation date, as you did. It depends on the effort. I think Artem can help make a decision here.

services/workflows-service/src/customer/customer.service.ts
```diff
@@ -1,20 +1,46 @@
...
+    if (!config || !config.isDemoAccount || !createdAt) {
+      return null;
+    }
+
+    const businessReportsCount = await this.merchantMonitoringClient.count({
+      customerId,
+    });
+
+    return {
+      totalReports: businessReportsCount,
```
MatanYadaev, line 36: Do we have a mechanism in the front-end the ensure this property gets updated when reports are added?

MatanYadaev: And just a quick comment: Ideally, this shouldn’t be under customers and should have its own endpoint, like `GET /api/business-reports/count`. But since it’s just for a demo, it's not a big deal

services/workflows-service/src/customer/customer.service.ts
```diff
@@ -1,20 +1,46 @@
...
+      maxBusinessReports: config.maxBusinessReports ?? 10,
+      showFullAccessPopup: config.showFullAccessPopup ?? false,
+      // FIXME: should count from the first login date / should be passed on creation ?
+      expiresAt: dayjs(createdAt).add(14, 'days').unix(),
```
MatanYadaev, line 40: Let's add a new `expiresAt` property in the config. This way we could easily change the expiration date manually. I also suggest moving the number `14` to the `env.ts` file.

services/workflows-service/src/customer/customer.service.ts
ArtemLavrentii: Do we expect for people to not sign in for multiple days? Creation date of a user wouldn't be when they send a request, but rather when they get the email / its get approved by the risk team.

We can also add first sign in date to a user, shouldn't be too hard to implement. I'm just unsure how useful it would be outside of demo. Maybe we could include that info in the customer config if isDemoAccount is set, wdyt ? ( we could even do it as "expiresAt" - eg just set it when user first logs in, if its not present

services/workflows-service/src/customer/customer.controller.external.ts
```diff
@@ -49,14 +55,17 @@ export class CustomerControllerExternal {
   @common.Get('/by-current-project-id')
   @swagger.ApiOkResponse({ type: [CustomerModel] })
   @swagger.ApiForbiddenResponse()
-  async getByCurrentProjectId(
-    @CurrentProject() currentProjectId: TProjectId,
-  ): Promise<TCustomerWithFeatures | null> {
+  async getByCurrentProjectId(@CurrentProject() currentProjectId: TProjectId): Promise<
```
ArtemLavrentii: Wouldn't something like this also work ?
```suggestion
  async getByCurrentProjectId(@CurrentProject() currentProjectId: TProjectId): Promise<(TCustomerWithFeatures & { config: { demoAccessDetails: TDemoAccessDetails } } ) | null>
```

In either case I would create separate type for TCustomerWithFeatures with demo access details so the return type isn't as messy.

services/workflows-service/src/global.d.ts
rzssh: image
image

maybe i'm missing something, but these types do not resolve to anything for me, i don't see them being imported. so i wonder how is this supposed to work? our types depend on these but they are literally `any`?

apps/backoffice-v2/src/common/components/molecules/GetFullAccessCard/GetFullAccessCard.tsx
```diff
@@ -15,7 +15,7 @@ export const GetFullAccessCard = () => {
   if (
     env.VITE_ENVIRONMENT_NAME === 'production' ||
```
MatanYadaev, line 16: We are going to create the demo accounts in production, and not sandbox this time, as this environment is more stable. Can you please remove this condition and make sure it works well on a production environment?

Context: https://teamballerine.slack.com/archives/C06MQM4F17D/p1739886859740529

services/workflows-service/src/customer/customer.service.ts
MatanYadaev, line 40: We usually use ISO format, please don't introduce new conventions (Unix timestamp) as it makes the source code confusing and inconsistent.

apps/backoffice-v2/src/pages/MerchantMonitoring/schemas.ts
```diff
@@ -127,3 +127,25 @@ export const MerchantMonitoringSearchSchema = BaseSearchSchema.extend({
   from: z.string().date().optional(),
   to: z.string().date().optional(),
 });
+
+const URL_REGEX =
+  /((https?):\/\/)?([a-zA-Z0-9-_]+\.)+[a-zA-Z0-9]+(\.[a-z]{2})?(\/[a-zA-Z0-9_#-]+)*(\/)?(\?[a-zA-Z0-9_-]+=[a-zA-Z0-9_-]+(&[a-zA-Z0-9_-]+=[a-zA-Z0-9_-]+)*)?(#[a-zA-Z0-9_-]+)?/;
```
MatanYadaev, line 132: Move it to `@ballerine/common` and reuse it in the WFS?

MatanYadaev: We also use it in other places as well.

PR #3069, opened 2025-02-19
BAL-3498: enhance rule-engine evaluation logic with path comparison

shanegrouber:
## Summary by CodeRabbit

- **New Features**
  - Improved rule processing to deliver more robust value extraction and advanced comparison handling.
  - Added a new property for indicating path-based comparisons in specific rule operators.
  - Enhanced rule engine tests with support for path comparisons.
- **Refactor**
  - Enhanced internal rule validation logic for clearer, more consistent evaluation.
- **Chores**
  - Updated internal migration settings to support ongoing system maintenance.


packages/common/src/rule-engine/operators/helpers.ts
```diff
@@ -39,12 +39,36 @@ export abstract class BaseOperator<TDataValue = Primitive, TConditionValue = Pri
 
   extractValue(data: unknown, rule: Rule) {
     const value = get(data, rule.key);
+    const OPERATORS_WITHOUT_PATH_COMPARISON = [
+      OPERATION.AML_CHECK,
+      OPERATION.BETWEEN,
+      OPERATION.LAST_YEAR,
+    ] as const;
```
tomer-shvadron: There's no need to re-declare this array every time the `extractValue` function is triggered. Extract it outside the class

services/workflows-service/src/rule-engine/core/rule-engine.ts
```diff
@@ -23,11 +23,19 @@ export const validateRule = (rule: Rule, data: any): RuleResult => {
     throw new OperatorNotFoundError(rule.operator);
   }
 
-  const value = operator.extractValue(data, rule);
+  const extractedValue = operator.extractValue(data, rule);
+
+  const isPathComparison =
+    typeof extractedValue === 'object' &&
```
tomer-shvadron: use the `isObject` function from `@ballerine/common` instead, see if you can use the `?.` notation after that

tomer-shvadron (approved):

PR #3071, opened 2025-02-19
BAL-3460 Documents context sync & bugfixes

chesterkmr:
- Implemented collection flow state transition when document is requested
- Implemented mapping of document statuses to context
- Added new documents to `common` package
- Implemented `$this` scope for DocumentField schema definition
- Various bugfixes


packages/common/src/schemas/documents/workflow/documents/schemas/ZZ.ts
```diff
@@ -86,6 +86,63 @@ export const getUniversalDocuments = (): TDocument[] => {
         issueDate: OptionalTypePastDate,
       }),
     },
+    {
+      category: 'proof_of_ownership',
+      type: 'business_utility_bill',
+      issuer: { country: 'ZZ' },
+      issuingVersion: 1,
+      version: 1,
+      propertiesSchema: Type.Object({}),
```
Omri-Levy: Are you sure using Type.Object({}) won't fail non-empty properties objects?
```suggestion
      propertiesSchema: Type.Object({}),
```

services/workflows-service/src/document/helpers/add-requested-document-to-entity-documents.ts
```diff
@@ -0,0 +1,122 @@
...
+  let result: IUIDefinitionElement | undefined = undefined;
+  const pages = (uiDefinition.uiSchema as unknown as { elements: IUIDefinitionPage[] }).elements;
+
+  const run = (elements: IUIDefinitionElement[]) => {
+    for (const element of elements) {
+      if (
+        element.element === 'documentfield' &&
+        element?.params?.template?.type === type &&
+        element?.params?.template?.category === category
+      ) {
```
Omri-Levy, line 53: This logic needs to be the same as the one in getDocumentId and in the backoffice. Where we also take into consideration fields like version and country.

packages/common/src/schemas/documents/workflow/documents/schemas/ZZ.ts
chesterkmr: Fixed

Omri-Levy (approved):

PR #3085, opened 2025-02-22
BAL-3436 Requested documents for individuals

chesterkmr:
- Implemented mapping of documents for individuals
- Implemented requested document creation for individials
- Fixed mapping
- Reworked entityfieldgroup document field to handle requested documents of individuals


services/workflows-service/src/document/document.service.ts
```diff
@@ -588,34 +601,46 @@ export class DocumentService {
...
-            ),
-          ),
-        ),
+        ubos: parsedUIDocuments.individuals.ubos.map(parsedDocument => {
+          const { ballerineEntityId } = parsedDocument;
+          const ubo = entitiesWithDocuments.ubos.find(ubo => ubo.id === ballerineEntityId);
+
+          if (!ubo) {
+            throw new Error('Ubo not found');
+          }
```
Omri-Levy, line 610: Don't think we should fail all the other ubos if one ubo was not found.

services/workflows-service/prisma/migrations/20250222165148_added_unprovided_status_to_document/migration.sql
Omri-Levy, line 1: "unprovided" is an impossible state. we might want to remove document status all together.

services/workflows-service/src/document/document.service.ts
```diff
@@ -673,7 +699,15 @@ export class DocumentService {
...
 
     const createdDocuments = await Promise.all(
-      documentsToCreate.map(doc => this.repository.create(doc)),
+      documentsToCreate.map(async ({ entityType, ...doc }) => {
+        const createdDocument = await this.repository.create(doc);
+
+        return {
+          ...createdDocument,
+          entityType,
+          entityId: entityType === 'business' ? undefined : createdDocument.endUserId,
```
Omri-Levy, line 708: Why not pass a business ID?

services/workflows-service/src/document/document.service.ts
```diff
@@ -683,21 +717,31 @@ export class DocumentService {
...
+            document.entityType as 'business' | 'ubo' | 'director',
+            uiDefinition,
+            documentToInsert,
+          )
+        : addRequestedDocumentToIndividualDocuments(
+            context,
+            document.entityType as 'ubo' | 'director',
+            uiDefinition,
+            documentToInsert,
+          );
```
Omri-Levy, line 744: Let's throw an error if the entity type is unsupported. There are other places in the system that use a ternary we're trying to remove over time.

services/workflows-service/src/document/helpers/add-requested-document-to-individuals-documents.ts
```diff
@@ -0,0 +1,125 @@
...
+    issuingCountry: string;
+    issuingVersion: string;
+    version: string;
+    status: Document['status'];
+    decision: Document['decision'];
+    entityId: string | undefined;
+  },
+) => {
+  if (entityType !== 'ubo' && entityType !== 'director') {
+    throw new Error('Requested documents are not supported for UBOs or Directors.');
```
Omri-Levy, line 32: Wrong message?

services/workflows-service/src/document/helpers/parse-document-definition.ts
```diff
@@ -0,0 +1,39 @@
...
+    .object({
+      type: z.string(),
+      id: z.string(),
+      category: z.string(),
+      issuer: z.object({
+        country: z.string(),
+      }),
+      issuingVersion: z.number(),
+      version: z.string(),
+      entityType: z.enum(['business', 'ubo', 'director']).default('business'),
```
Omri-Levy, line 21: No entity type seems more like an error than something to use a default for.

PR #3124, opened 2025-03-02
feat(bal-3613): added magic link authentication

ArtemLavrentii:
## Summary by CodeRabbit

- **New Features**
  - Introduced Magic Link Authentication for a seamless and secure login experience—magic links now auto-remove tokens from the URL after successful sign-in.
  - Added new environment variable for Magic Link authentication configuration.
- **Refactor**
  - Streamlined sign-in and sign-out workflows to improve reliability and user experience.
  - Updated import paths for clarity and maintainability.
- **Chores**
  - Upgraded dependencies and enhanced monitoring integrations to boost overall system stability.


services/workflows-service/src/env.ts
```diff
@@ -95,6 +95,7 @@ export const serverEnvSchema = {
   IN_MEMORIES_SECRET_CONSUMER_KEY: z.string().optional(),
   SYNC_UNIFIED_API: z.string().optional().default('true'),
   DEFAULT_DEMO_DURATION_DAYS: z.number().optional().default(14),
+  MAGIC_LINK_JWT: z.string(),
```
alonp99: Should be called USER_AUTH_JWT_SECRET probably

services/workflows-service/src/auth/auth.service.ts
```diff
@@ -11,20 +12,46 @@ export class AuthService {
     private readonly passwordService: PasswordService,
   ) {}
 
+  async validateUserById(userId: string): Promise<UserInfo | null> {
```
alonp99: This function dosent actually validate anything, it takes fetches user data and returns return some of the data

services/workflows-service/src/auth/auth.service.ts
```diff
@@ -11,20 +12,46 @@ export class AuthService {
...
     }
 
     return null;
   }
+
+  private createUserInfo = (
+    user: { status: UserStatus; id: string; firstName: string; lastName: string; roles: JsonValue },
+    email: string,
+  ) => {
+    if (user?.status !== UserStatus.Active) {
```
alonp99, line 48: checking if the user active dosent seem to be the job of "createUserInfo" there is a mix up with the other function validateUserById (probably)

apps/backoffice-v2/src/domains/auth/components/AuthenticatedLayout/AuthenticatedLayout.loader.ts
```diff
@@ -1,12 +1,27 @@
...
-import { LoaderFunction } from 'react-router-dom';
 
-export const authenticatedLayoutLoader: LoaderFunction = async () => {
+export const authenticatedLayoutLoader: LoaderFunction = async ({ request }) => {
   if (!env.VITE_AUTH_ENABLED) return null;
 
+  const url = new URL(request.url);
+  const token = url.searchParams.get('token');
+
+  if (token) {
```
alonp99, line 14: Did you test user+pass worked properly

apps/backoffice-v2/src/domains/auth/fetchers.ts
```diff
@@ -51,6 +51,22 @@ export const fetchSignIn = async ({ callbackUrl, body }: ISignInProps) => {
   return handleZodError(error, session);
 };
 
+export const fetchMagicLinkSignIn = async ({ token }: { token: string }) => {
+  const [session, error] = await apiClient({
+    endpoint: 'auth/magic-link-login',
+    method: Method.POST,
```
tomer-shvadron: in the function name, fetch tells the developer that it's a "get" type of function, so they could expect `method: Method.GET`. Because it is a POST request, my guess is that `createMagicLinkSignIn` would be a better name

apps/backoffice-v2/src/domains/auth/components/AuthenticatedLayout/AuthenticatedLayout.loader.ts
```diff
@@ -1,12 +1,27 @@
...
+    try {
+      await queryClient.fetchQuery(['magic-link-auth', token], ({ queryKey }) =>
+        fetchMagicLinkSignIn({ token: queryKey[1]! }),
+      );
+    } catch (e) {
+      console.error('Error using magic link', e);
```
tomer-shvadron, line 20: I believe that `console.error` and then redirect won't work very well together. I think we should report to Sentry instead

tomer-shvadron (approved):

services/workflows-service/src/env.ts
ArtemLavrentii: But this is only used only for magic links, or should we just give the magic link tokens a its own scope ? ( agree with _SECRET part, through )

apps/backoffice-v2/src/domains/auth/components/AuthenticatedLayout/AuthenticatedLayout.loader.ts
ArtemLavrentii: redirect is here so people won't just have infinite loader. Right now we don't have a page for "magic link expired", and people still need to login somehow after magic token auth has failed.

Or you are trying to say that redirect in RR wouldn't use browser history ? I think it still does use it, and there isn't an actual page reload. In SSR it would log to backend, so its still OK in that case too.

apps/backoffice-v2/src/domains/auth/fetchers.ts
ArtemLavrentii: Its in the spirit of this file :) Lets then rename normal sign in too.

I think create is also super-horrible name for POST ( PUT is for creation operations ), and its especially for login. 

What about `doSignInViaMagicLink` ( and `doSignIn` for the other function ) ?

tomer-shvadron: I'm all for renaming badly named functions :) It could make discussions like this one not needed 😂 
Usually POST is for creation and PUT is for updating complete resources (as partial-update with PATCH).
I don't really like "do" as I think it works the same without it, so `signIn` and `signInWithMagicLink` look fine to me.

apps/backoffice-v2/src/domains/auth/components/AuthenticatedLayout/AuthenticatedLayout.loader.ts
tomer-shvadron: I mean that if we console.error, it shows in the client's console but we have no record about it. You can disregard the "won't work very well together comment", I think I was thinking about something else... Sorry about that

ArtemLavrentii: I think its better to add sentry integration for console integration, would make more sense.

apps/backoffice-v2/src/domains/auth/components/AuthenticatedLayout/AuthenticatedLayout.loader.ts
ArtemLavrentii: Yes, it also works.

Issue #2721, opened 2024-09-18
Bug - Installation Failure - SQL Error


sarshad01:
### Description
`pnpm run kyb-manual-review-example` results in failure, more specifically database connection error  when `@ballerine/workflows-service@0.7.37 db:reset` is invoked. Running  `npm db:reset` executes successfully if ran manually within services/workflows-service directory. However, running `npm run seed` throws Foreign key constraint error (Alert_businessId_fkey (index))

### Expected Behaviour
BackOffice, Collection and Workflows services should start listening

### Actual Behaviour
Command Fails with Exit code 1

### Environment
* Node version: v20.17.0
* Device, operating system and its version: Ubuntu 22.04.4 LTS
  *
* Browser and its version: Firefox 130.0
  *
* Version of the affected apps and packages: Dev Branch @ c39a8df
  *

### Steps to Reproduce
1. Follow the steps here : https://docs.ballerine.com/en/getting_started/installation/
2. `pnpm run kyb-manual-review-example` would exit the process with the database connection error
3. Run `npm run db:reset` while the db container is running and the command will succeed
4. Run `npm run seed` this will fail showing foreign key error

### Any possible solutions
1- My assumption was the database connection error occurs due to localhost being used, I switched it to 127.0.0.1 in the /services/workflows-service/.env however it did not solve the issue.

2- I then tried to run `npm run db:reset` manually and it worked

3- I then manually ran `npm run seed` which failed throwing foreign key constraint error

### Examples and references

Output of `pnpm run kyb-manual-review-example` :

![Screenshot from 2024-09-18 19-28-50](https://github.com/user-attachments/assets/36a2f3e4-a764-4212-bb76-3cbeb2e49af6)

 

Output of  `npm run seed` : 

![Screenshot from 2024-09-18 19-32-58](https://github.com/user-attachments/assets/d0564f21-3853-4806-bfa1-943a41f5c1c9)

WFA-juliver:
is this first run?

sarshad01:
@WFA-juliver  I have tried multiple times and unsuccessful before trying to install it via docker

WFA-juliver:
@sarshad01 when running the example, did you verify if theres a db container with postgres running?

try

```
rm -rf node_modules \
&& pnpm install \
&& pnpm monorepo:init \
&& pnpm run kyb-manual-review-example

```
and check if theres database running `docker ps -a`

sarshad01:
[...]

PR #3129, opened 2025-03-04
Give DB more time to come online

EternalBlueScreen:
Added sleep statement for database to fully come online. Without this the migration fails.

## Summary by CodeRabbit

- **Chores**
  - Updated the database initialization process to use a fixed 2-second delay before resetting and seeding, ensuring a smoother startup experience.


EternalBlueScreen:
The script provided at: https://github.com/ballerine-io/ballerine/issues/2721#issuecomment-2368817872 Might be a more elegant solution which also provides a dynamic wait.

PR #3130, opened 2025-03-04
fix(MerchantMonitoringTable): remove duplicate risk indicators

shanegrouber:
- Implement logic to filter out duplicate violation indicators
- Ensure each indicator is unique based on its ID

## Summary by CodeRabbit

- **Bug Fixes**
	- Enhanced the monitoring view by filtering out duplicate violations, ensuring only unique entries are displayed for a clearer overview.
	- Updated filtering logic to exclude specific risk levels from the violations displayed.

### DIFF
--- a/apps/backoffice-v2/src/pages/MerchantMonitoring/components/MerchantMonitoringTable/columns.tsx
+++ b/apps/backoffice-v2/src/pages/MerchantMonitoring/components/MerchantMonitoringTable/columns.tsx
@@ -162,6 +162,9 @@ export const useColumns = ({ isDemoAccount = false }) => {
             .filter(
               el => el.id !== NO_VIOLATION_DETECTED_RISK_INDICATOR_ID && el.name && el.riskLevel,
             )
+            .filter(
+              (violation, index, self) => index === self.findIndex(t => t.id === violation.id),
+            )
             .sort((a, b) => {
               return a.riskLevel === b.riskLevel
                 ? (a.name ?? '').localeCompare(b.name ?? '')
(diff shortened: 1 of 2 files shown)


ArtemLavrentii (approved):

apps/backoffice-v2/src/pages/MerchantMonitoring/components/MerchantMonitoringTable/columns.tsx
```diff
@@ -160,7 +163,12 @@ export const useColumns = ({ isDemoAccount = false }) => {
         cell: ({ row }) => {
           const violations = (row.original.data?.allViolations ?? [])
             .filter(
-              el => el.id !== NO_VIOLATION_DETECTED_RISK_INDICATOR_ID && el.name && el.riskLevel,
+              (violation, index, self) =>
+                violation.id !== NO_VIOLATION_DETECTED_RISK_INDICATOR_ID &&
+                violation.name &&
+                violation.riskLevel &&
+                violation.riskLevel !== POSITIVE_RISK_LEVEL_ID &&
+                index === self.findIndex(t => t.id === violation.id),
```
ArtemLavrentii, line 171: I feel like this is overcomplicated a bit, couldn't this do the same job ?
```suggestion
                self[index]?.id === violation.id,
```

shanegrouber: I don't think this works. You are just comparing the value at the current index to the violation ID, so all elements will always pass the filter.

If you want to remove the inner loop to lower the time complexity + make it more readable, we can maybe implement a Set to track seen violation ids?

ArtemLavrentii: Ahh, so you are checking if it was seen before or not.

Then yes, I think it would be better to have a set to track such cases, or even have a filter that would pre-filter for unique values like this:
```
.filter((violation) => {
  if (seenViolations.has(violation.id)) return false;
  seenViolations.add(violation.id);
}
```

Currently for me personally its not clear that you are trying to find if this violation was displayed before, so at least creating variable like `const seenBefore = ...` would help here for code self-documentation.

But this is just a nit, maybe sometimes in the future ;-)

PR #3174, opened 2025-03-19
feat(export reports): add loading indicator, remove unncessary toasts

jonathan-shemer:
## Summary by CodeRabbit

- **New Features**
  - Improved report export interface now displays a dynamic loading indicator and enhanced tooltip for clearer user feedback during export operations.

- **Refactor**
  - Streamlined export and report-fetching workflows by removing redundant notifications, resulting in a cleaner, more direct user interaction.
  - Updated internal state management for export actions ensures consistent behavior during the export process.
  
- **Chores**
  - Updated the `string-ts` dependency to a specific version for improved stability.


apps/backoffice-v2/src/common/utils/export-to-csv/export-to-csv.unit.test.ts
```diff
@@ -143,12 +134,10 @@ describe('CSV Export Utils', () => {
       vi.clearAllMocks();
     });
 
-    test('should show error toast for empty data', () => {
+    test('returns false for empty data', () => {
```
Omri-Levy: Consider keeping the previous format.
```suggestion
    test('should return false for empty data', () => {
```

Omri-Levy (approved):

PR #3186, opened 2025-03-26
Company info editing

Omri-Levy:


apps/backoffice-v2/src/lib/blocks/hooks/useEntityInfoBlock/useEntityInfoBlock.tsx
```diff
@@ -0,0 +1,136 @@
...
+}: {
+  entity: TWorkflowById['context']['entity'];
+  workflow: TWorkflowById;
+  entityDataAdditionalInfo: TWorkflowById['context']['entity']['data']['additionalInfo'];
+}) => {
+  const { mutate: mutateEvent } = useEventMutation();
+  const onMutateEvent = useCallback(() => {
+    mutateEvent({
+      workflowId: workflow?.id,
+      event: 're_run_checks',
```
alonp99: a bit more explicit, tomorrow there will be another block that will run different checks

alonp99 (approved):

apps/backoffice-v2/src/common/components/organisms/EditableDetailsV2/components/EditableDetailsV2Options.tsx
```diff
@@ -41,6 +45,16 @@ export const EditableDetailsV2Options: FunctionComponent<{
             <Edit size={16} className="me-2" /> Edit
           </Button>
         </DropdownMenuItem>
+        <DropdownMenuItem className={`h-6 w-full`} asChild>
+          <Button
+            variant={'ghost'}
+            className="justify-start text-xs leading-tight aria-disabled:pointer-events-none aria-disabled:opacity-50"
+            aria-disabled={actions.reRunChecks.disabled}
```
chesterkmr, line 52: `aria-disabled` is present but not actual `disabled` flag. Is that intentional ?

apps/backoffice-v2/src/common/components/organisms/EditableDetailsV2/EditableDetailsV2.tsx
```diff
@@ -91,27 +102,25 @@ export const EditableDetailsV2: FunctionComponent<IEditableDetailsV2Props> = ({
...
-              filteredFields?.some(({ props }) => props.isEditable) && (
-                <Button
-                  type="submit"
-                  className={`aria-disabled:pointer-events-none aria-disabled:opacity-50`}
-                  aria-disabled={config.actions.save.disabled}
-                >
-                  Save
-                </Button>
-              )}
+            {isEditable && filteredFields?.some(({ props }) => props.isEditable) && (
```
chesterkmr: Consider memoize this check.

apps/backoffice-v2/src/lib/blocks/variants/ManualReviewBlocks/hooks/useManualReviewBlocksLogic/useManualReviewBlocksLogic.tsx
```diff
@@ -58,7 +58,7 @@ export const useManualReviewBlocksLogic = () => {
     [mutateRevisionTaskById],
   );
 
-  const businessInformation = useCaseInfoBlock({
+  const businessInformation = useEntityInfoBlock({
     entity: workflow?.context?.entity ?? {},
```
chesterkmr, line 62: Consider memoize this otherwise each rerender logic within hook related to entity will re-run.

apps/backoffice-v2/src/common/components/organisms/EditableDetailsV2/components/EditableDetailsV2Options.tsx
Omri-Levy: Yes. This is considered an accessibility best practice. Disabled removes information from screen readers.

apps/backoffice-v2/src/lib/blocks/variants/ManualReviewBlocks/hooks/useManualReviewBlocksLogic/useManualReviewBlocksLogic.tsx
Omri-Levy: I really don't think I should. This would just mean doing it for every single hook in the backoffice,
I think our focus should be to decouple markup from logic and data and then memoize data properly.

In these cases these objects are composed of multiple nested references with no computation done,
I don't see how useMemo here would help unless we could use React compiler and hope
these objects are handled by the compiler.

PR #3190, opened 2025-03-27
Country select input

Omri-Levy:


packages/common/rollup.config.ts
```diff
@@ -102,6 +102,7 @@ function esm({ input, packageDir, external, banner }: Options): RollupOptions {
       dir: `${packageDir}/dist/esm`,
       banner,
       preserveModules: true,
+      preserveModulesRoot: path.resolve(process.cwd(), 'src'),
```
Omri-Levy, line 105: Without this change dist includes the full path to common like "packages/common/src" happens once the dependencies introduces in this PR are exported.

chesterkmr (approved):

PR #3195, opened 2025-03-31
feat(BAL-3355): assert UBOs match between collection flow and the com…

jonathan-shemer:
…pany registry

## Summary by CodeRabbit

- **New Features**
  - Introduced a new operator for UBO matching that enables enhanced comparisons of Ultimate Beneficial Owner data.
  - Expanded rule configurations to support specific UBO matching criteria with improved validation.
  - Added a new schema for UBO match operations to facilitate validation.
  - Introduced a new constant and class for handling UBO mismatch operations.

- **Tests**
  - Added comprehensive test scenarios to ensure robust and reliable behavior under various UBO matching conditions.


packages/common/src/rule-engine/rules/schemas.ts
```diff
@@ -106,6 +106,11 @@ export const RuleSchema = z.discriminatedUnion('operator', [
     isPathComparison: z.boolean().default(false),
     threshold: z.number().min(0).max(100).default(80),
   }),
+  z.object({
+    key: z.string().optional(),
+    operator: z.literal(OPERATION.UBO_MATCH),
+    value: PrimitiveSchema.optional(),
+  }),
```
tomer-shvadron:
```suggestion
  z.object({
    key: z.string().optional(),
    operator: z.literal(OPERATION.UBO_MATCH),
    value: PrimitiveSchema.optional(),
    isPathComparison: z.boolean().default(false),
  }),
```

tomer-shvadron (approved):

packages/common/src/rule-engine/operators/helpers.ts
```diff
@@ -455,6 +456,74 @@ class FuzzyMatchScoreLt extends BaseOperator<Primitive, Primitive, Promise<boole
...
+    }
+  }
+
+  evaluate = (data: { collectionUbos: string[]; registryUbos: string[] }): boolean => {
+    const { collectionUbos, registryUbos } = data;
+    const exactMatch =
+      collectionUbos.length === registryUbos.length &&
+      collectionUbos.every((name, index) => name === registryUbos[index]);
+
+    return !exactMatch;
```
alonp99, line 523: It is a bit confusing that the rule is called UboMatch, and the return value is the opposite

packages/common/src/rule-engine/operators/helpers.ts
```diff
@@ -455,6 +456,74 @@ class FuzzyMatchScoreLt extends BaseOperator<Primitive, Primitive, Promise<boole
...
+  constructor() {
+    super({
+      operator: 'UBO_MATCH',
+    });
+  }
+
+  extractValue(data: unknown): { collectionUbos: string[]; registryUbos: string[] } {
+    try {
+      const normalizedString = z.string().transform(name => name.toUpperCase().trim());
+      const result = z
```
alonp99, line 469: It might be better to extract the schema to the outside of the class
- it can be reused (for example to the fuzzy matching)
- extract value will be more readable
- we won't create it every time the rule runs

alonp99: oh you can also use the typescript type of the schema and replace the "any" in baseOperator

services/workflows-service/src/rule-engine/core/test/rule-engine.unit.test.ts
```diff
@@ -1139,4 +1112,129 @@ describe('Rule Engine', () => {
       `);
     });
   });
+
+  describe('UBO match operator', () => {
```
alonp99, line 1116: ❤️

packages/common/src/rule-engine/operators/helpers.ts
jonathan-shemer: got the same comment from https://github.com/ballerine-io/ballerine/pull/3195#pullrequestreview-2729102897

and applied

packages/common/src/rule-engine/operators/helpers.ts
jonathan-shemer: so basically the `any` type in `BaseOperator` refers to the argument received by `evaluate`, rather than `extractValue` in which i have declared the schema.

definitely can be reused, i'll leave that for future shemer to do when implementing `UBOMismatchFuzzey`, or more probably the same rule with a parameter that mentioned if it should be exactly equal (`threshold: 100`) or some fuzzy threshold.

PR #3199, opened 2025-04-01
feat(workflows): enhance workflow authentication and request handling

tomer-shvadron:
- Implement combined authentication guard for improved security
- Add utility to extract workflow ID from URL for request enrichment
- Refactor request handling to include workflow ID in requests when present

## Summary by CodeRabbit

- **New Features**
  - Added plugins to support updating end user information and fetching company information within collection flows.
  - Introduced new utility functions for building collection flow URLs and handling workflow IDs.
  - Enhanced authentication with a new workflow-based guard for improved security and flexibility.
  - Expanded data models to track actor information in workflow runtime data.

- **Improvements**
  - Updated multiple API endpoints and URL patterns to include workflow IDs for better traceability.
  - Enhanced integration with external plugins and improved request handling by including credentials in cross-origin requests.
  - Refined query logic and context management in the collection flow application for more robust workflow handling.
  - Improved test infrastructure with additional context and actor services.

- **Bug Fixes**
  - Adjusted authentication decorators and guards to ensure correct application of default authentication logic.
  - Fixed various minor code style and readability issues.

- **Chores**
  - Removed deprecated token-based authentication logic and related environment variables.
  - Updated and reorganized type definitions and interfaces for clarity and maintainability.
  - Cleaned up redundant or outdated code and configuration files.


services/workflows-service/src/common/guards/combined-auth.guard.ts
```diff
@@ -0,0 +1,81 @@
...
+  ) {}
+
+  async canActivate(context: ExecutionContext) {
+    const req = context.switchToHttp().getRequest<Request>();
+    const workflowId = req.params['wf-id'] || req.query['wf-id'];
+
+    const authenticatedEntity = req.user as AuthenticatedEntity & { projectIds?: TProjectIds };
+
+    if (
+      req.isAuthenticated() ||
```
jonathan-shemer: if you have an expired session, this would fallback to `token` - we don't want that i think. can we think of another solution?

or can we assert that going through the edit flow would never include the `token` for the `endUser`?

scenario:
1. i'm an agent, i want to edit a specific workflow.
2. i've spent 3 hours editing it (and having lunch).
3. my session-cookie is now expired (ttl set to 2 hours).
4. i click submit, and the backend now thinks i'm the `endUser` rather than myself.

tomer-shvadron: Done

apps/backoffice-v2/src/common/components/molecules/CaseVideoGuide/CaseVideoGuide.tsx
```diff
@@ -35,7 +35,7 @@ export const CaseVideoGuide: FunctionComponent<CaseVideoGuideProps> = ({
                   height: 0,
                 }}
               >
-                <Skeleton className="absolute inset-0 size-full" />
+                <Skeleton className="size-full absolute inset-0" />
```
jonathan-shemer, line 38: can we not commit these css changes?

apps/kyb-app/src/common/providers/AccessTokenProvider/AccessTokenProvider.tsx
```diff
@@ -10,34 +9,46 @@ interface IAccessTokenProviderProps {
...
-    [accessToken, setAccessToken],
+    [wfIdToken, accessToken, setWfIdToken, setAccessToken],
   );
 
   useEffect(() => {
+    if (wfIdToken) {
+      const previousToken = searchParams.get('wf-id');
+
+      if (previousToken !== wfIdToken) {
+        setSearchParams({ 'wf-id': wfIdToken });
```
jonathan-shemer: `wfIdToken` is a token? there's something i don't understand

apps/kyb-app/src/common/utils/request.ts
```diff
@@ -1,9 +1,17 @@
...
-import ky, { HTTPError } from 'ky';
+import ky, { HTTPError, Options } from 'ky';
 import { isExceptionWillBeHandled } from './helpers';
 
-export const request = ky.create({
+// Helper function to get workflow ID from URL
+const getWorkflowIdFromUrl = () => {
+  const urlParams = new URLSearchParams(window.location.search);
+
+  return urlParams.get('wf-id');
```
jonathan-shemer: why did you choose to have this as a query-param rather than the path?
i.e. `collection-flow/cm8r65m1y00b011aiwo8upuz9?token=a8b84fd0-3cae-46fa-88b8-e2f64e34573c`

services/workflows-service/src/common/guards/combined-auth.guard.ts
```diff
@@ -0,0 +1,84 @@
+import type { Request } from 'express';
+import { ClsService } from 'nestjs-cls';
+import { CanActivate, ExecutionContext, Injectable, UnauthorizedException } from '@nestjs/common';
+
+import { AuthenticatedEntity, TProjectIds } from '@/types';
+import { WorkflowService } from '@/workflow/workflow.service';
+import { WorkflowTokenService } from '@/auth/workflow-token/workflow-token.service';
+
+@Injectable()
+export class CombinedAuthGuard implements CanActivate {
```
jonathan-shemer: what do you think about
```suggestion
export class WorkflowAuthGuard implements CanActivate {
```
as this seems very coupled to authenticating access to workflows.

services/workflows-service/src/common/guards/combined-auth.guard.ts
```diff
@@ -0,0 +1,84 @@
...
+  constructor(
+    private readonly cls: ClsService,
+    private readonly workflowService: WorkflowService,
+    private readonly workflowTokenService: WorkflowTokenService,
+  ) {}
+
+  async canActivate(context: ExecutionContext) {
+    const req = context.switchToHttp().getRequest<Request>();
+    // const workflowId = typeof req.query['wf-id'] === 'string' ? req.query['wf-id'] : null;
```
jonathan-shemer: i removed this comment in my brach

apps/kyb-app/src/common/utils/request.ts
```diff
@@ -13,11 +21,15 @@ export const request = ky.create({
...
   },
+  credentials: 'include',
   timeout: 30_000,
   hooks: {
     beforeRequest: [
       request => {
-        request.headers.set('Authorization', `Bearer ${getAccessToken()}`);
+        request.headers.set(
+          'Authorization',
+          `Bearer ${getAccessToken('wf-id') ?? getAccessToken()}`,
```
jonathan-shemer: why do you use the `wf-id` as if it's a `Bearer` token? shouldn't this just be an ID?
Authorization should be handled solely in the session cookie and this header should not be set.

jonathan-shemer (changes requested):

apps/kyb-app/src/common/providers/AccessTokenProvider/AccessTokenProvider.tsx
tomer-shvadron: Improved naming - all instances of `wfIdToken` were changed to `workflowId`

apps/kyb-app/src/common/utils/request.ts
tomer-shvadron: I mainly followed existing structure with `token`, I didn't think it's necessary to use it as part of the path (because you anyway need to add it to the query-param upon creating the server request, otherwise we would also need to update ALL endpoints... WDYT?

apps/kyb-app/src/common/utils/request.ts
tomer-shvadron: We have to do that to support other existing middlewares (`auth-key.middleware.ts` and `admin-key.middleware.ts`). Without it they fail.

apps/backoffice-v2/src/common/components/molecules/CaseVideoGuide/CaseVideoGuide.tsx
tomer-shvadron: That shit snuck in and when I try to fix it locally, it auto-fixes it back. It is annoying but I don't have a normal solution for that, I think we should try and treat that as tech-debt or something like that

apps/kyb-app/src/common/utils/request.ts
jonathan-shemer: let's fix those middlewares to handle these cases

jonathan-shemer:
Let's make sure to address `collection-flow.controller.ts`, it still uses `TokenAuthGuard` exclusively (and any others).

apps/kyb-app/src/common/utils/request.ts
```diff
@@ -67,3 +81,33 @@ export const request = ky.create({
...
+  return {
+    ...options,
+    searchParams,
+  };
+};
+
+// Export wrapped instance
+export const request = {
+  ...instance,
+  get: (url: string, options?: Options) => instance.get(url, addWorkflowId(options)),
```
chesterkmr: Instead of wrapping each request type wont it be possible do add this logic to beforeRequest interceptor?

chesterkmr (approved):

apps/backoffice-v2/src/pages/Entity/components/Case/components/CaseOptions/hooks/useCaseOptionsLogic/mutations/useCopyCollectionFlowLinkMutation/useCopyCollectionFlowLinkMutation.tsx
```diff
@@ -9,7 +9,7 @@ export const useCopyCollectionFlowLinkMutation = ({ workflow }: { workflow: TWor
         throw new Error('Collection flow URL or token not available');
       }
 
-      const url = `${workflow.context.metadata.collectionFlowUrl}?token=${workflow.context.metadata.token}`;
+      const url = `${workflow.context.metadata.collectionFlowUrl}/?workflowId=${workflow.id}&token=${workflow.context.metadata.token}`;
```
jonathan-shemer: common?

apps/kyb-app/src/common/utils/create-query-params-string.ts
```diff
@@ -0,0 +1,15 @@
+export const createQueryParamsString = (
```
jonathan-shemer: isn't there something built in that can handle url serialization?

apps/kyb-app/src/common/utils/request.ts
jonathan-shemer: or to where we're already adding `token` (or in case we can't for some reason, [line 89](https://github.com/ballerine-io/ballerine/pull/3199/files#diff-7b24fd04c76c8a115b5d65db5056ebdb54622ff503f48215392b07ad242899a0R89) )?

PR #3219, opened 2025-04-09
feat: mm dates & table columns (BAL-3738, BAL-3942, BAL-3801, BAL-3453, BAL-3954, BAL-3413)

r4zendev:
https://github.com/user-attachments/assets/cd9d6f2b-d394-42a0-8a60-105e767bf7f7

This PR also moves Statistics page to the Home page for the following reasons:
- First, it supports updated sidebar logic that prevents the user from losing his query params on navigating to the same page. It also removes flicker when a user is on the home page and presses the home nav item. Better UX and more user-friendly.
- Second, the redirect logic there looked like a hack done via useEffect. When we reach a point where we want a separate page for Workflows dashboard, we might add this decomposition back or opt for using it as a component instead of being a separate page.

## Only cases access
Screenshot 2025-04-10 at 17 07 54

## Merchant monitoring access
Screenshot 2025-04-10 at 16 54 32

## Ongoing monitoring access
Screenshot 2025-04-10 at 16 59 11

## Access to both (no ongoing)
Screenshot 2025-04-10 at 17 08 51

## Access to all
Screenshot 2025-04-10 at 17 09 11

## Summary by CodeRabbit

- **New Features**
  - Introduced a customizable date range picker and several new analytics components on the Home page—including dynamic metrics, case graphs, pie chart cards, stats cards, and onboarding risk analytics—to provide richer, interactive insights.
- **Refactor & Improvements**
  - Enhanced navigation by preserving URL search parameters and updating route redirections.
  - Updated the calendar to allow custom action placements and reintroduced essential monitoring details.
- **Cleanup**
  - Removed outdated pages (e.g., legacy statistics and workflows) to streamline the user experience.


apps/backoffice-v2/src/common/components/organisms/DateRangePicker/DateRangePicker.tsx
```diff
@@ -0,0 +1,124 @@
...
+        <PopoverContent className="w-auto p-0" align="start">
+          <Calendar
+            {...props}
+            initialFocus
+            mode="range"
+            selected={value}
+            onSelect={onChange}
+            numberOfMonths={2}
+            bottomActions={
+              <div className={`flex`}>
```
Omri-Levy: Should probably add a space-x-2 or space-x-1 so when hovered or focused the buttons don't touch.

apps/backoffice-v2/src/common/components/organisms/DateRangePicker/DateRangePicker.tsx
```diff
@@ -0,0 +1,124 @@
...
+                      e,
+                    )
+                  }
+                >
+                  Last 30 days
+                </Button>
+
+                <Button
+                  variant={`ghost`}
+                  className={`!mt-0 h-8 select-none font-normal hover:bg-transparent`}
```
Omri-Levy: If there's both no hover state and no cursor change it will be unclear this is a button for the user.

Omri-Levy (approved):

Omri-Levy (approved):

PR #3243, opened 2025-04-15
fix: dashboard logic fixes, navigation fixes (BAL-3979, BAL-3988)

r4zendev:
## Summary by CodeRabbit

- **New Features**
  - Added support for "allow all dates" filtering, enabling users to view data without date restrictions in search and reporting features.
  - Pie chart segments are now clickable, allowing navigation to filtered views based on chart data.
  - Pie chart card titles are now customizable for better context.

- **Improvements**
  - Date range queries now include the entire end date for more accurate reporting.
  - Query string construction is more robust and structured across multiple components.
  - Enhanced navigation logic for a more intuitive back button experience.

- **Bug Fixes**
  - Corrected minor comment typos (no impact on user experience).

- **Refactor**
  - Improved code structure for date and query handling without changing user-facing behavior.

### DIFF
--- a/apps/backoffice-v2/src/common/hooks/useHomeLogic/useHomeLogic.ts
+++ b/apps/backoffice-v2/src/common/hooks/useHomeLogic/useHomeLogic.ts
@@ -16,6 +16,10 @@ export const HomeSearchSchema = z.object({
   mmTo: z.string().date().optional(),
   casesFrom: z.string().date().optional(),
   casesTo: z.string().date().optional(),
+  allowAllDates: z
+    .string()
+    .transform(value => (value === 'true' ? true : false))
+    .optional(),
 });
 
 const RISK_LEVEL_COLORS = {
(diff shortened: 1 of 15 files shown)


apps/backoffice-v2/src/pages/Home/components/StaticMetricsSection/StaticMetricsSection.tsx
```diff
@@ -44,20 +44,23 @@ export const StaticMetricsSection = ({
       {isCasesOnboardingEnabled && (
         <>
           <PieChartCard
+            title="Cases by Status"
```
jonathan-shemer: shouldn't this be `Active (ongoing) Cases By Status`?

apps/backoffice-v2/src/common/hooks/useHomeLogic/useHomeLogic.ts
```diff
@@ -16,6 +16,10 @@ export const HomeSearchSchema = z.object({
   mmTo: z.string().date().optional(),
   casesFrom: z.string().date().optional(),
   casesTo: z.string().date().optional(),
+  allowAllDates: z
+    .string()
+    .transform(value => (value === 'true' ? true : false))
```
jonathan-shemer:
```suggestion
    .transform(value => value.toLowerCase() === 'true')
```

also, might we have already implemented `string -> boolean` somewhere common?

apps/backoffice-v2/src/pages/MerchantMonitoring/schemas.ts
```diff
@@ -125,6 +125,10 @@ export const MerchantMonitoringSearchSchema = BaseSearchSchema.extend({
     .catch('All'),
   from: z.string().date().optional(),
   to: z.string().date().optional(),
+  allowAllDates: z
+    .string()
+    .transform(value => (value === 'true' ? true : false))
```
jonathan-shemer: note lower-case comment

jonathan-shemer (changes requested):

jonathan-shemer:
Otherwise, LGTM!

apps/backoffice-v2/src/pages/MerchantMonitoring/schemas.ts
rzssh: just blindly copied this from other place, the condition could be simplified to just `=== 'true'`, but i don't think adding lowercase is necessary. we will route to `'true'` pretty much from anywhere in the app, explicitly casting to lower case is not needed

jonathan-shemer (approved):

PR #3248, opened 2025-04-18
BAL-4004: Wildcard element ids

chesterkmr:
- Added `*` to element ids so they can match in cases when values are dynamic during revision or edit.

## Summary by CodeRabbit

- **New Features**
  - Added support for wildcard and exact ID matching in form field prioritization, enabling more flexible field selection.
  - Introduced new utility functions for wildcard ID formatting and matching.
  - Added helper to check if a collection flow step is in "edit" state.

- **Bug Fixes**
  - Improved revision field generation to correctly handle steps in both "edit" and "revision" states.
  - Updated how revision field IDs are formatted, now using wildcard suffixes for consistency.
  - Simplified step revision status check logic without changing behavior.

- **Tests**
  - Added and updated unit tests for new and existing helper functions, including wildcard and exact ID matching.
  - Refactored revision field generation tests for clearer and more realistic scenarios.

- **Chores**
  - Updated dependencies and package versions across multiple packages.
  - Enhanced changelogs to reflect recent updates.


apps/kyb-app/src/pages/CollectionFlow/versions/v2/components/organisms/CollectionFlowUI/helpers/check-if-step-in-edit.ts
```diff
@@ -0,0 +1,14 @@
...
+import { CollectionFlowStepStatesEnum, getCollectionFlowState } from '@ballerine/common';
+
+export const checkIfStepInEdit = (stepName: string, context: CollectionFlowContext) => {
+  const collectionFlow = getCollectionFlowState(context);
+
+  const step = collectionFlow?.steps?.find(step => step.stepName === stepName);
+
+  if (!step) {
+    return false;
+  }
```
Omri-Levy: Redundant. If there is no step, state can't be equal to edit.

packages/ui/src/components/organisms/Form/DynamicForm/helpers/is-exact-id-match/is-exact-id-match.unit.test.ts
```diff
@@ -0,0 +1,52 @@
+import { isExactIdMatch } from './is-exact-id-match';
+import { describe, expect, it } from 'vitest';
+
+describe('isExactIdMatch', () => {
+  it('should return true when ids match exactly', () => {
```
Omri-Levy: Consider separating the when clause to another describe so a single when could have multiple it clauses.

packages/ui/src/components/organisms/Form/DynamicForm/helpers/is-match-as-wildcard/is-match-as-wildcard.ts
```diff
@@ -0,0 +1,11 @@
+import { formatWildcardId } from '@/components/organisms/Form/DynamicForm/helpers/format-wildcard-id/format-wildcard-id';
+
+export const isMatchAsWildcard = (id: string, priorityFieldId: string) => {
+  if (!priorityFieldId.endsWith('-*')) {
+    return false;
+  }
```
Omri-Levy: If you add a wildcard to both and they don't match isn't the result the same as this check?

packages/ui/src/components/organisms/Form/DynamicForm/hooks/internal/usePriorityFields/usePriorityFields.ts
```diff
@@ -3,30 +3,42 @@ import { useDynamicForm } from '../../../context';
...
 import { useElementId } from '../../external';
+import { isExactIdMatch } from '../../../helpers/is-exact-id-match/is-exact-id-match';
+import { isMatchAsWildcard } from '../../../helpers/is-match-as-wildcard/is-match-as-wildcard';
 
 export const usePriorityFields = (element: IFormElement<string, any>) => {
   const { priorityFields, priorityFieldsParams = { behavior: 'disableOthers' } } = useDynamicForm();
 
   const { stack } = useStack();
   const elementId = useElementId(element, stack);
+  const originElementId = element.id;
```
Omri-Levy, line 14: Why is this needed?

packages/ui/src/components/organisms/Form/DynamicForm/helpers/is-match-as-wildcard/is-match-as-wildcard.ts
chesterkmr: Adding wildcard to "both" will break `priorityFieldId` since it may already include wildcard.

packages/ui/src/components/organisms/Form/DynamicForm/hooks/internal/usePriorityFields/usePriorityFields.ts
chesterkmr: `useElementid` - returns modified `id` for cases when this element is within array. `originElementId` is something unmutable and always same so it can be used for comparison in `wildcard` cases.

apps/kyb-app/src/pages/CollectionFlow/versions/v2/components/organisms/CollectionFlowUI/helpers/check-if-step-in-edit.ts
chesterkmr: Refactored.

packages/ui/src/components/organisms/Form/DynamicForm/helpers/is-exact-id-match/is-exact-id-match.unit.test.ts
chesterkmr: Done

packages/ui/src/components/organisms/Form/DynamicForm/helpers/is-match-as-wildcard/is-match-as-wildcard.ts
Omri-Levy: What I mean is if this if clause would be removed the logic would behave the same.

packages/ui/src/components/organisms/Form/DynamicForm/hooks/internal/usePriorityFields/usePriorityFields.ts
Omri-Levy: I didn't understand why not use element.id directly. Just a shorthand?

chesterkmr: For readablity and avoiding of confusions. If i remove`originElementId` then it will be confused why `element.id` is used instead of `elementId`.

packages/ui/src/components/organisms/Form/DynamicForm/helpers/is-match-as-wildcard/is-match-as-wildcard.ts
chesterkmr: Got it now, yeah make sense.

chesterkmr: Updated.

Omri-Levy (approved):

PR #3250, opened 2025-04-20
feat(cookies): stricter cookies policy

jonathan-shemer:
## Summary by CodeRabbit

- **New Features**
  - Added new environment configuration options for session cookie security, allowing customization of HTTP-only, SameSite, secure cookie, and proxy trust settings.
- **Enhancements**
  - Session cookies now use configurable security options from environment variables, improving flexibility and alignment with deployment requirements.
- **Other**
  - Updated environment example file with new session security variables and explanatory comments.
  - User session data now excludes personal details, retaining only the user ID for improved privacy.


services/workflows-service/src/env.ts
```diff
@@ -23,6 +23,25 @@ export const serverEnvSchema = {
...
+    .default(true),
+  SESSION_SECURE_COOKIE: z
+    .union([z.literal('true'), z.literal('false'), z.boolean()])
+    .transform((val: unknown) => val === 'true')
+    .default(true),
+  SESSION_SECURE_PROXY: z
+    .union([z.literal('true'), z.literal('false'), z.boolean()])
+    .transform(val => {
+      return val === 'true';
+    })
```
tomer-shvadron:
```suggestion
    .transform((val: unknown) => val === 'true')
```

services/workflows-service/src/env.ts
```diff
@@ -23,6 +23,25 @@ export const serverEnvSchema = {
...
   DB_URL: z.string().url(),
   SESSION_SECRET: z.string(),
+  SESSION_SAME_SITE: z
+    .union([z.literal('strict'), z.literal('lax'), z.literal('none')])
+    .default('strict'),
+  SESSION_HTTP_ONLY: z
+    .union([z.literal('true'), z.literal('false'), z.boolean()])
+    .transform(val => {
+      return val === 'true';
+    })
```
tomer-shvadron:
```suggestion
    .transform((val: unknown) => val === 'true')
```

tomer-shvadron (approved):

services/workflows-service/.env.example
```diff
@@ -38,3 +38,11 @@ HASHING_KEY_SECRET_BASE64=JDJiJDEwJFRYNjhmQi5JMlRCWHN0aGowakFHSi4=
 SECRETS_MANAGER_PROVIDER=in-memory
 SYNC_UNIFIED_API=false
 MAGIC_LINK_AUTH_JWT_SECRET=a-string-secret-at-least-256-bits-long
+# Security settings - Set to true in production to prevent XSS/CSRF attacks
+SESSION_HTTP_ONLY=true
```
alonp99: I am not sure this configuration will work locally, maybe we should have something like `env.local.example`

jonathan-shemer: i'll update the example to work locally, i'm not sure we need a separate example for local.

PR #3286, opened 2025-05-07
fix(bal-4087): magic links not working when url leads to 404 page

ArtemLavrentii:
## Summary by CodeRabbit

- **New Features**
  - Improved handling of magic link authentication via token in the root route, including automatic sign-in and redirection when a token is present in the URL.
  - Enhanced URL normalization to ensure consistent language prefixing and removal of token parameters after authentication.
  - Applied root route loader logic to catch-all routes for consistent authentication and redirection behavior.

- **Bug Fixes**
  - Resolved issues where magic link authentication and token-based sign-in were not consistently handled on all routes, including catch-all paths.


apps/backoffice-v2/src/pages/Root/Root.loader.ts
```diff
@@ -1,9 +1,34 @@
...
-  return redirect(`/en${url.pathname === '/' ? '' : url.pathname}${url.search}`);
+  if (token) {
+    try {
+      await queryClient.fetchQuery(['magic-link-auth', token], ({ queryKey }) =>
+        magicLinkSignIn({ token: queryKey[1]! }),
+      );
+    } catch (e) {
+      console.error('Error using magic link', e);
+      return redirect(`/en/auth/sign-in`);
+    }
```
Omri-Levy, line 18: From what I remember fetchQuery itself won't throw an error.

ArtemLavrentii: I just checked it, and it does throw. 
image

Omri-Levy (approved):

PR #3289, opened 2025-05-08
feat(rule-engine): introduce AML_CHECK_V2 operator for enhanced AML checks

tomer-shvadron:
- Add AML_CHECK_V2 operator to validate adverse media, PEP, sanctions, and fitness probity
- Implement rule evaluation to return results based on the presence of AML hits
- Update TypeScript types and schemas to support new operator logic and helpers

## Summary by CodeRabbit

- **New Features**
  - Introduced a new AML_CHECK_V2 operator for advanced Anti-Money Laundering checks, supporting evaluation across multiple AML categories and entity roles.
  - Added support for helper functions (e.g., retrieving end user data) during rule evaluation, enabling more flexible and dynamic rule processing.

- **Bug Fixes**
  - Improved error handling and validation in AML-related operators.

- **Tests**
  - Added comprehensive tests for AML_CHECK_V2 and updated existing tests to utilize the new helpers context.

- **Chores**
  - Updated interfaces and service methods to support and propagate the helpers context throughout rule engine workflows.


packages/common/src/rule-engine/operators/aml-v2-check.ts
```diff
@@ -0,0 +1,92 @@
...
+    const mainRepresentativeEndUserId =
+      context?.entity?.data?.additionalInfo?.mainRepresentative?.ballerineEntityId;
+
+    const endUsers = await Promise.all(
+      [...ubosIds, ...directorsIds, mainRepresentativeEndUserId]
+        .filter(Boolean)
+        .map(endUserId => helpers.getEndUserById(endUserId)),
+    );
+
+    if (isEmpty(endUsers)) {
```
alonp99, line 51: Consider checking before the db call if there are and ids to check, and just return.

if there are some ID's but nothing returned from the database, something is wrong and you should trigger sentry or at least a log

packages/common/src/rule-engine/operators/aml-v2-check.ts
```diff
@@ -0,0 +1,92 @@
...
+      return false;
+    }
+
+    const hits: Array<z.infer<typeof EndUserAmlHitsSchema>> = endUsers
```
alonp99, line 56: Since we know all won't be on the user in the future, it's worth mentioning it in a comment here to help future dev thiking why this rule is not working or where the refactor should happen in this rule

packages/common/src/rule-engine/operators/aml-v2-check.ts
```diff
@@ -0,0 +1,92 @@
...
+      const result = amlOperator.dataValueSchema?.safeParse(data);
+
+      if (result && !result.success) {
+        return false;
+      }
+
+      const conditionResult = amlOperator.conditionValueSchema?.safeParse(conditionValue.value);
+
+      if ((conditionResult && !conditionResult.success) || !conditionResult?.data) {
+        return false; // TODO: throw explicit error
```
alonp99: ^

packages/common/src/rule-engine/operators/enums.ts
```diff
@@ -12,6 +12,7 @@ export const OPERATION = {
   IN_CASE_INSENSITIVE: 'IN_CASE_INSENSITIVE',
   NOT_IN: 'NOT_IN',
   AML_CHECK: 'AML_CHECK',
+  AML_CHECK_V2: 'AML_CHECK_V2',
```
alonp99, line 15: Do we need to migrate current v1 usage to v2?

alonp99 (approved):

packages/common/src/rule-engine/operators/enums.ts
tomer-shvadron: Will be done in BAL-4113

PR #3290, opened 2025-05-08
fix: reworked document update & fixed not persisted comment & reason

chesterkmr:
- Wrapped document updates in transaction
- Fixed bug where document `comment` & `decisionReason` were not persisted during decision update
- Fixed bug where next version of document contained `comment` of previous version

## Summary by CodeRabbit

- **Bug Fixes**
  - Ensured that comments are no longer carried over when creating a new document during a revision, resulting in new documents without inherited comments.

- **Refactor**
  - Improved transaction handling for document decision updates, ensuring all related operations are performed atomically and consistently.
  - Enhanced internal consistency by updating method signatures for better transaction management and batch processing of document decisions.

### DIFF
--- a/services/workflows-service/src/document/document.service.ts
+++ b/services/workflows-service/src/document/document.service.ts
@@ -53,6 +53,9 @@ import {
   EntitySchema,
   TParsedDocuments,
 } from './types';
+import { defaultPrismaTransactionOptions } from '@/prisma/prisma.util';
+import { beginTransactionIfNotExistCurry } from '@/prisma/prisma.util';
+import { PrismaService } from '@/prisma/prisma.service';
 
 @Injectable()
 export class DocumentService {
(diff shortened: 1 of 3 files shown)


services/workflows-service/src/document/document.service.ts
```diff
@@ -476,172 +492,231 @@ export class DocumentService {
...
     });
   }
 
-  private async persistDocumentDecisionInToContext(document: Document, projectId: TProjectId) {
-    if (!document.workflowRuntimeDataId) {
-      throw new BadRequestException(
-        `Document with id ${document.id} has no workflow runtime data id`,
-      );
-    }
+  private async persistDocumentDecisionInToContext(
```
tomer-shvadron:
```suggestion
  private async persistDocumentDecisionInContext(
```

tomer-shvadron (approved):

PR #3341, opened 2025-06-01
feat(rule-engine): add bank account verification operator

tomer-shvadron:
- Implement BankAccountVerification class for rule evaluation
- Update relevant enums and constants for integration
- Add unit tests for verifying rules based on bank account data

## Summary by CodeRabbit

- **New Features**
  - Introduced support for bank account verification rules, enabling evaluation of specific rule scores for account verification scenarios.

- **Tests**
  - Added comprehensive tests to verify correct behavior of bank account verification rule evaluation.

- **Documentation**
  - Updated schemas to include the new bank account verification operator for improved clarity in rule definitions.


tomer-shvadron:
Reminder: Update packages before merging

packages/common/src/rule-engine/operators/bank-account-verification.ts
```diff
@@ -0,0 +1,98 @@
...
+        }),
+      );
+
+    const result = bankAccountVerificationSchema.safeParse(data);
+
+    if (!result.success) {
+      throw new ValidationFailedError('Extract value', 'parsing failed', result.error);
+    }
+
+    if (result.data.requestType !== 'BAVCommercial-Standard') {
```
alonp99: It looks like its being reused so i would extract it to a const

packages/common/src/rule-engine/operators/bank-account-verification.ts
```diff
@@ -0,0 +1,98 @@
...
+        }),
+      )
+      .safeParse(result.data.rules);
+
+    if (!rulesResult.success) {
+      throw new ValidationFailedError('Extract value', 'parsing failed', rulesResult.error);
+    }
+
+    const rules = rulesResult.data.filter(rule =>
+      ['CMM1069', 'CMM1048', 'CMM1052'].includes(rule.ruleId),
```
alonp99: maybe also extract it to consts and leave a comment about what they say?

packages/common/src/rule-engine/operators/bank-account-verification.ts
tomer-shvadron: I didn't understand, what is reused? `requestType` serves no purpose before or after this line, same goes for the string

alonp99: the string, in tests

alonp99 (approved):

PR #3446, opened 2025-06-24
Bal 4393 document above 10mb failed to upload with no indicative

chesterkmr:


packages/ui/src/components/organisms/Form/DynamicForm/providers/ValidatorWrapper/ValidatorWrapper.tsx
```diff
@@ -11,28 +11,33 @@ export const ValidatorWrapper = <TValue extends object>({
   const { documents, files } = useDocumentsService();
 
   const valueWithFilesAndDocuments = useMemo(() => {
+    const filesFromDocuments = documents.reduce((acc, document) => {
+      const fileId = files.composeFileId({
+        type: document.type,
+        category: document.category,
+        entityType: document.businessId ? 'business' : 'ubo',
+        entityId: document.businessId ? document.businessId : document.endUserId!,
```
Omri-Levy, line 19: Please don't use ternary for similar checks, just because there's no business it does not indicate it's an end-user.

packages/ui/src/components/organisms/Form/DynamicForm/providers/ValidatorWrapper/ValidatorWrapper.tsx
```diff
@@ -11,28 +11,33 @@ export const ValidatorWrapper = <TValue extends object>({
...
+      });
+
+      if (!document.files?.[0]) {
+        return acc;
+      }
+
+      return {
+        ...acc,
+        [fileId]: new File([], document.files?.[0]?.name!, {
```
Omri-Levy, line 28: In general we should throw more errors or runtime validate over using null assertions.

packages/ui/src/components/organisms/Form/DynamicForm/validators/document-size/document-size-validator.ts
```diff
@@ -0,0 +1,61 @@
...
+const biteToMbInteger = (bite: number) => {
+  return bite / 1024 / 1024;
+};
+
+const isExceedsFileSize = (file: File, fileSizeLimit: number) => {
+  if (fileSizeLimit && file.size > fileSizeLimit) {
+    return true;
+  }
+
+  return false;
```
Omri-Levy, line 19:
```suggestion
  return fileSizeLimit && file.size > fileSizeLimit;
```

packages/ui/src/components/organisms/Form/DynamicForm/validators/document-size/document-size-validator.ts
```diff
@@ -0,0 +1,61 @@
...
+};
+
+export const documentSizeValidator: TValidator<
+  TDocument[],
+  IDocumentSizeValidatorParams,
+  TBaseValidators | 'documentSize',
+  IValidatorWrapperContext
+> = (value, params, schema, context) => {
```
Omri-Levy, line 27: Consider using params over positional arguments.

packages/ui/src/components/organisms/Form/DynamicForm/validators/document-size/document-size-validator.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  if (isValueAnEntity(value)) {
+    const documentId = formatDocumentId({
+      type: schema.metadata?.element?.params?.template?.type,
+      category: schema.metadata?.element?.params?.template?.category,
+      entityType: 'ubo',
+      entityId: value.ballerineEntityId || value.__id!,
+    });
+
+    file = context?._files[documentId];
```
Omri-Levy, line 53: Didn't we remove properties like __id and __files?

packages/ui/src/components/organisms/Form/DynamicForm/fields/DocumentField/hooks/useDocumentUpload/useDocumentUpload.ts
```diff
@@ -51,6 +53,10 @@ export const useDocumentUpload = (
 
       setFile(file);
 
+      if (document) {
+        void deleteDocumentFiles(document.id);
```
Omri-Levy, line 57: Does this use react-query? How come it does not use the sync mutate?

Omri-Levy (approved):

packages/ui/src/components/organisms/Form/DynamicForm/providers/ValidatorWrapper/ValidatorWrapper.tsx
chesterkmr: This logic is taken from document-validator as reference for now. Currently i dont have any better approach. Reworking this means reworking other validator and its something that out of scope in current context.

packages/ui/src/components/organisms/Form/DynamicForm/validators/document-size/document-size-validator.ts
chesterkmr: Its something that added during runtime and used only for validation. Never gets to actual context.

packages/ui/src/components/organisms/Form/DynamicForm/fields/DocumentField/hooks/useDocumentUpload/useDocumentUpload.ts
chesterkmr: It uses react-query mutation under the hood yes, but in current case its more like side effect and doesnt need to be "sync" awaited

PR #3448, opened 2025-06-24
fix: replaced fuzzy search with exact search in dropdown

chesterkmr:


packages/ui/src/components/atoms/SearchableDropdown/SearchableDropdown.tsx
```diff
@@ -27,6 +27,10 @@ interface ISearchableDropdownProps {
   disablePortal?: boolean;
 }
 
+const strictFilter = (value: string, search: string) => {
+  return +value.toLowerCase().includes(search.toLowerCase());
```
Omri-Levy: If for some reason coercion to number is intentional please use the Number or parseInt namespaces.
```suggestion
  return Number(value.toLowerCase().includes(search.toLowerCase());
```

Omri-Levy (approved):

packages/ui/src/components/atoms/SearchableDropdown/SearchableDropdown.tsx
chesterkmr: Refactored, added comment.

PR #3460, opened 2025-06-29
fix: fixed rendering of csv documents

chesterkmr:
## Summary by CodeRabbit

* **New Features**
  * Added support for handling and displaying base64-encoded document files, including CSVs, throughout the application.
  * CSV files are now automatically fetched and converted to base64 format for improved compatibility and viewing.

* **Improvements**
  * Enhanced document filtering to include files with either image URLs or base64 data, optimizing performance and flexibility.
  * Updated document interfaces to support the new base64 property for better data handling.

* **Chores**
  * Updated internal dependencies and optimized unused imports for cleaner codebase.

### DIFF
--- a/apps/backoffice-v2/src/pages/Entity/components/Case/hooks/useDocuments/helpers.ts
+++ b/apps/backoffice-v2/src/pages/Entity/components/Case/hooks/useDocuments/helpers.ts
@@ -1,10 +1,11 @@
 import { isCsv } from '@/common/utils/is-csv/is-csv';
 import { convertCsvToPdfBase64String } from '../../../../../../common/utils/convert-csv-to-pdf-base64-string/convert-csv-to-pdf-base64-string';
 import { IDocumentsProps } from '../../interfaces';
+import { isBase64 } from '@/common/utils/is-base64/is-base64';
 
 export const convertCsvDocumentsToPdf = (documents: IDocumentsProps['documents']) => {
   return documents?.map(document => {
-    if (isCsv(document) && document.base64) {
+    if (isCsv(document) && isBase64(document.base64 || '')) {
       return { ...document, imageUrl: convertCsvToPdfBase64String(document.base64) };
     }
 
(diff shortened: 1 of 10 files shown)


services/workflows-service/src/document/document.service.ts
```diff
@@ -1176,17 +1180,44 @@ export class DocumentService {
...
         propertiesSchema: documentWithPropertiesSchema.propertiesSchema,
       };
     });
+
+    return Promise.all(formatPromises);
+  }
+
+  private isCsv(file: { mimeType: string | null }) {
+    return file.mimeType === 'text/csv' || file.mimeType === 'application/csv';
+  }
```
Omri-Levy: Move to a utility. As a general rule if its not truly related to the service it should sit on a relevant service or just a utility.

services/workflows-service/src/document/document.service.ts
```diff
@@ -1176,17 +1180,44 @@ export class DocumentService {
...
+
+  private async fetchCsvFromUrlAndCovertToBase64(csvUrl: string) {
```
Omri-Levy: Same comment as above.

apps/backoffice-v2/src/pages/Entity/components/Case/hooks/useDocuments/helpers.ts
```diff
@@ -4,8 +4,8 @@ import { IDocumentsProps } from '../../interfaces';
 
 export const convertCsvDocumentsToPdf = (documents: IDocumentsProps['documents']) => {
   return documents?.map(document => {
-    if (isCsv(document)) {
-      return { ...document, imageUrl: convertCsvToPdfBase64String(document.imageUrl) };
+    if (isCsv(document) && document.base64) {
```
Omri-Levy:
```suggestion
    if (isCsv(document) && isBase64(document.base64)) {
```

Omri-Levy (approved):

PR #3465, opened 2025-07-01
feat: added safe filter with warnings for missing entity ids

chesterkmr:
## Summary by CodeRabbit

* **New Features**
  * Improved filtering of entity IDs to ensure only valid IDs are included, with warnings logged when invalid or missing IDs are encountered.
* **Chores**
  * Updated internal dependencies for data migrations.

### DIFF
--- a/apps/backoffice-v2/src/common/utils/safe-entity-id-filter/safe-entity-id-filter.ts
+++ b/apps/backoffice-v2/src/common/utils/safe-entity-id-filter/safe-entity-id-filter.ts
@@ -1,6 +1,6 @@
-export const safeEntityIdFilter = <T>(list: Array<T>, filter: (item: T) => boolean) => {
+export const safeEntityIdFilter = <T>(list: Array<T>, predicate: (item: T) => boolean) => {
   return list.filter(item => {
-    const result = filter(item);
+    const result = predicate(item);
     const isMissing = !result;
 
     if (isMissing) {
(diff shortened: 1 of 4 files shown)


apps/backoffice-v2/src/common/utils/safe-entity-id-filter/safe-entity-id-filter.ts
```diff
@@ -0,0 +1,14 @@
+export const safeEntityIdFilter = <T>(list: Array<T>, filter: (item: T) => boolean) => {
+  return list.filter(item => {
+    const result = filter(item);
```
Omri-Levy:
```suggestion
export const safeEntityIdFilter = <T>(list: Array<T>, predicate: (item: T) => boolean) => {
  return list.filter(item => {
    const result = predicate(item);
```

apps/backoffice-v2/src/lib/blocks/hooks/useBusinessDocuments/useBusinessDocuments.ts
```diff
@@ -7,7 +8,9 @@ export const useBusinessDocuments = (workflow: TWorkflowById) => {
   const entityIds = useMemo(
     () =>
       workflow?.context?.entity?.ballerineEntityId
-        ? [workflow?.context?.entity?.ballerineEntityId]
+        ? safeEntityIdFilter([workflow?.context?.entity], entity =>
+            Boolean(entity.ballerineEntityId),
+          ).map(entity => entity.ballerineEntityId)
         : [],
     [workflow],
   );
```
Omri-Levy: Please switch to an if condition.

Omri-Levy (approved):

